PR #14, opened 2026-04-25
enable synergy deploy with cloud mode in holos

yuzhustudio:
## Summary

Adds hosted deployment mode for Synergy.

The goal is to support a cloud deployment where:
- The Web UI is deployed separately through S3 and CloudFront.
- Each Synergy backend runs inside an owner-scoped container.
- The backend handles Holos authentication and workspace isolation.

## Changes

- Added hosted backend mode controlled by SYNERGY_HOSTED.
- Added Holos JWT cookie authentication.
- The backend validates HS256 JWT using SYNERGY_JWT_SECRET.
- The JWT user id must match HOLOS_OWNER_ID.
- The health endpoint is exempted from authentication.
- Added workspace isolation based on SYNERGY_SCOPE_ROOT.
- Requests outside the hosted workspace scope are rejected.
- Added support for disabling Web UI mounting in backend-only hosted mode.
- Skipped local single-instance locking in hosted mode to avoid stale Docker volume lock issues.
- Added hosted frontend access flow.
- Added control plane API configuration through VITE_SYNERGY_CONTROL_API_BASE.
- Added hosted profile lookup, agent lookup, invite verification, and health polling.
- Restricted debug URL override to explicit debug mode.
- Ensured hosted frontend requests include root-domain Holos cookies.
- Preserved hosted onboarding auto-verification for existing backend Holos credentials.
- Hid the Switch server command in hosted mode while keeping provider connection available.
- Added hosted Dockerfile and dockerignore.
- Updated README with hosted frontend build, backend image build, and runtime environment variables.

## Verification

- App typecheck passed.
- Synergy typecheck passed.
- Git diff check passed.
- Hosted frontend build passed.
- Hosted Docker image build passed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+.git
+.turbo
+node_modules
+packages/*/node_modules
+packages/sdk/js/node_modules
+packages/*/dist
+packages/sdk/js/dist
+dist
+logs
+*.log
+.DS_Store
--- a/Dockerfile.hosted
+++ b/Dockerfile.hosted
@@ -1,0 +1,23 @@
+FROM oven/bun:1.3.11
+
+WORKDIR /app
+
+COPY package.json bun.lock bunfig.toml turbo.json tsconfig.json ./
+COPY patches ./patches
+COPY script ./script
+COPY packages ./packages
+
+RUN bun install --frozen-lockfile
+
+ENV NODE_ENV=production
+ENV HOME=/workspace
+ENV SYNERGY_HOSTED=1
+ENV SYNERGY_SCOPE_ROOT=/workspace
+ENV SYNERGY_DISABLE_WEB_MOUNT=1
+
+RUN mkdir -p /workspace/schema
+RUN cp /app/packages/synergy/schema/config.schema.json /workspace/schema/config.schema.json
+
+EXPOSE 4096
+
+CMD ["bun", "--cwd", "packages/synergy", "src/index.ts", "server", "--hostname", "0.0.0.0", "--port", "4096"]
--- a/Dockerfile.hosted
+++ b/Dockerfile.hosted
@@ -11,12 +11,13 @@ RUN bun install --frozen-lockfile
 
 ENV NODE_ENV=production
 ENV HOME=/workspace
+ENV SYNERGY_ROOT=/workspace/.synergy
 ENV SYNERGY_HOSTED=1
 ENV SYNERGY_SCOPE_ROOT=/workspace
 ENV SYNERGY_DISABLE_WEB_MOUNT=1
 
-RUN mkdir -p /workspace/schema
-RUN cp /app/packages/synergy/schema/config.schema.json /workspace/schema/config.schema.json
+RUN mkdir -p /workspace/.synergy/schema
+RUN cp /app/packages/synergy/schema/config.schema.json /workspace/.synergy/schema/config.schema.json
 
 EXPOSE 4096
 
--- a/README.md
+++ b/README.md
@@ -302,6 +302,7 @@ Hosted backend containers should receive owner and auth configuration from the c
 
 ```bash
 SYNERGY_HOSTED=1
+SYNERGY_ROOT=/workspace/.synergy
 SYNERGY_SCOPE_ROOT=/workspace
 SYNERGY_DISABLE_WEB_MOUNT=1
 SYNERGY_AUTH_COOKIE_NAME=holos_jwt
@@ -311,10 +312,12 @@ SYNERGY_JWT_AUDIENCE=...
 HOLOS_OWNER_ID=...
 ```
 
+`SYNERGY_ROOT` is the exact Synergy data root and defaults to the normal `$HOME/.synergy` path when unset. In the hosted image it is set to `/workspace/.synergy`, so mounting `/data/synergy-workspace/{agent_id}` to `/workspace` keeps Synergy state under `/data/synergy-workspace/{agent_id}/.synergy`.
 `SYNERGY_SCOPE_ROOT` defaults to `/workspace`, and `SYNERGY_AUTH_COOKIE_NAME` defaults to `holos_jwt`.
 `SYNERGY_JWT_ISSUER` and `SYNERGY_JWT_AUDIENCE` are optional but recommended when the JWT secret is shared with other Holos services.
 
 In hosted mode, the server defaults scope resolution to `SYNERGY_SCOPE_ROOT`, rejects directories outside that root, and validates `holos_jwt` itself before serving non-health API requests.
+With the hosted image defaults, the credential store lives at `/workspace/.synergy/data/auth/api-key.json`.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -306,8 +306,13 @@ SYNERGY_SCOPE_ROOT=/workspace
 SYNERGY_DISABLE_WEB_MOUNT=1
 SYNERGY_AUTH_COOKIE_NAME=holos_jwt
 SYNERGY_JWT_SECRET=...
+SYNERGY_JWT_ISSUER=...
+SYNERGY_JWT_AUDIENCE=...
 HOLOS_OWNER_ID=...
 ```
+
+`SYNERGY_SCOPE_ROOT` defaults to `/workspace`, and `SYNERGY_AUTH_COOKIE_NAME` defaults to `holos_jwt`.
+`SYNERGY_JWT_ISSUER` and `SYNERGY_JWT_AUDIENCE` are optional but recommended when the JWT secret is shared with other Holos services.
 
 In hosted mode, the server defaults scope resolution to `SYNERGY_SCOPE_ROOT`, rejects directories outside that root, and validates `holos_jwt` itself before serving non-health API requests.
 
--- a/README.md
+++ b/README.md
@@ -277,6 +277,40 @@ curl -fsSL https://raw.githubusercontent.com/SII-Holos/synergy/main/packages/met
 
 The installer places the binary under `~/.meta-synergy/bin/` and optionally adds it to your `PATH`.
 
+## Hosted Deployment
+
+Synergy can run as a hosted backend container while the Web UI is served separately from static hosting such as S3 and CloudFront.
+
+Build the hosted frontend with explicit environment flags:
+
+```bash
+VITE_SYNERGY_HOSTED=1 \
+VITE_SYNERGY_ALLOW_DEBUG_URL=0 \
+VITE_SYNERGY_CONTROL_API_BASE=https://api.holosai.io \
+bun run --cwd packages/app build
+```
+
+Use `VITE_SYNERGY_ALLOW_DEBUG_URL=1` only for test builds where `?url=` backend overrides are intended.
+
+Build the hosted backend image:
+
+```bash
+docker build -f Dockerfile.hosted -t synergy-hosted:latest .
+```
+
+Hosted backend containers should receive owner and auth configuration from the control plane:
+
+```bash
+SYNERGY_HOSTED=1
+SYNERGY_SCOPE_ROOT=/workspace
+SYNERGY_DISABLE_WEB_MOUNT=1
+SYNERGY_AUTH_COOKIE_NAME=holos_jwt
+SYNERGY_JWT_SECRET=...
+HOLOS_OWNER_ID=...
+```
+
+In hosted mode, the server defaults scope resolution to `SYNERGY_SCOPE_ROOT`, rejects directories outside that root, and validates `holos_jwt` itself before serving non-health API requests.
+
 ## Development
 
 ### Prerequisites
--- a/packages/app/src/app-access.tsx
+++ b/packages/app/src/app-access.tsx
@@ -1,0 +1,113 @@
+import { Show, Switch, Match, createMemo, createSignal, lazy, Suspense, type ParentProps } from "solid-js"
+import { DialogSelectServer } from "@/components/dialog"
+import { useOnboarding, OnboardingProvider } from "@/context/onboarding"
+import { AuthProvider } from "@/context/auth"
+import { ServerProvider, useServer } from "@/context/server"
+import { GlobalSDKProvider } from "@/context/global-sdk"
+import { HolosProvider } from "@/context/holos"
+import { InputProvider } from "@/context/input"
+import { useDialog } from "@ericsanchezok/synergy-ui/context/dialog"
+import { ServerConnectionErrorPage } from "@/pages/server-connection-error"
+import { type AppAccess } from "@/utils/runtime"
+import { MainApp } from "@/app-main"
+
+const Welcome = lazy(() => import("@/pages/onboarding/welcome"))
+const Setup = lazy(() => import("@/pages/onboarding/setup"))
+
+export const Loading = () => <div class="size-full flex items-center justify-center text-text-weak">Loading...</div>
+
+function ServerKey(props: ParentProps) {
+  const server = useServer()
+  return (
+    <Show when={server.url} keyed>
+      {props.children}
+    </Show>
+  )
+}
+
+function OnboardingGate(props: { access: AppAccess }) {
+  const onboarding = useOnboarding()
+
+  return (
+    <Show when={onboarding.ready} fallback={<Loading />}>
+      <Switch>
+        <Match when={onboarding.phase === "welcome"}>
+          <Suspense fallback={<Loading />}>
+            <Welcome serverUrl={props.access.attachUrl} callbackUrl={props.access.callbackUrl} />
+          </Suspense>
+        </Match>
+        <Match when={onboarding.phase === "setup" || onboarding.phase === "ready"}>
+          <ServerProvider defaultUrl={props.access.attachUrl}>
+            <ServerKey>
+              <ConnectedApp />
+            </ServerKey>
+          </ServerProvider>
+        </Match>
+      </Switch>
+    </Show>
+  )
+}
+
+function ConnectedApp() {
+  const dialog = useDialog()
+  const onboarding = useOnboarding()
+  const server = useServer()
+  const [setupConnectionFailed, setSetupConnectionFailed] = createSignal(false)
+
+  const startupView = createMemo<"loading" | "connection-error" | "setup" | "ready">(() => {
+    const healthy = server.healthy()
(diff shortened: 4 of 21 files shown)


EricSanchezok:
## Architecture Review: Hosted Mode Impact on Synergy Core

Thanks for the PR — the direction is sound and the frontend refactor is a genuine improvement. Below is a deep review focused on **architectural impact**, **Scope/Global/Project consistency**, and **code quality**, organized by severity.

---

### 🔴 Breaking: `isGlobalRoute` expansion changes scope resolution for ALL modes

`server.ts` adds `/path`, `/scope`, `/provider`, `/config`, `/auth` to `isGlobalRoute`. This function decides whether a request gets `Scope.global()` or goes through `resolveScopedRequestScope`.

**Before this PR**, calling `GET /path?directory=/my/project` would resolve that directory's scope and return its path info. **After this PR**, `/path` is always classified as a global route, so `Instance.directory` is always `Scope.global().directory` (i.e., `Global.Path.home`), regardless of the `directory` query parameter.

This is a **behavioral regression in non-hosted mode**. The same applies to `/scope`, `/provider`, `/config`, and `/auth` — they all lose per-request scope resolution.

**Fix**: If the intent is "these routes should use global scope in hosted mode because there's only one scope", make the expansion conditional:

```ts
function isGlobalRoute(pathname: string) {
  if (
    pathname === "/global" ||
    pathname.startsWith("/global/") ||
    pathname === "/holos" ||
    pathname.startsWith("/holos/") ||
    pathname === "/channel" ||
    pathname.startsWith("/channel/")
  ) return true

  if (Hosted.enabled()) {
    return (
      pathname === "/path" ||
      pathname.startsWith("/path/") ||
      pathname === "/scope" ||
      pathname.startsWith("/scope/") ||
      pathname === "/provider" ||
      pathname.startsWith("/provider/") ||
      pathname === "/config" ||
      pathname.startsWith("/config/") ||
      pathname === "/auth" ||
      pathname.startsWith("/auth/")
    )
  }

  return false
}
```

Or better: extract the route classification into a configurable set so it doesn't grow as an if-chain.

---

### 🔴 Fragile: `Scope.global().directory` in hosted mode depends on coincidental `HOME == SCOPE_ROOT`

In the hosted Dockerfile:
```dockerfile
ENV HOME=/workspace
ENV SYNERGY_SCOPE_ROOT=/workspace
```

`Scope.global()` returns `{ directory: Global.Path.home }`, which reads `$HOME`. So `Scope.global().directory == "/workspace"`, which happens to equal `SYNERGY_SCOPE_ROOT`.

This means:
- The `isGlobalRoute` expansion works only because `HOME` and `SCOPE_ROOT` are the same
- `Hosted.resolveDirectory("global")` skips isolation checking, so if `HOME != SCOPE_ROOT`, the global scope's directory would be outside the hosted scope root
- Any future change to either value silently breaks isolation

**Fix**: Either make hosted mode override `Scope.global()` to return `directory: Hosted.scopeRoot()`, or remove the `"global"` fast-path in `Hosted.resolveDirectory` and let it resolve normally through `Scope.fromDirectory`.

---

### 🟡 Architectural drift: `Hosted` namespace bypasses `Flag`

Synergy has `Flag` namespace (`flag/flag.ts`) as the single source of truth for environment variable flags. This PR creates a parallel `Hosted` namespace with its own `envFlag()` that accepts different truthy values (`"yes"`, `"on"` vs `Flag.truthy`'s `"true"`, `"1"` only).

This means:
- Two different flag-reading conventions coexist
- `hosted.ts` is imported by `config/config.ts` and `server/runtime.ts` — non-server modules now depend on a server module
- Future contributors won't know whether to use `Flag` or `Hosted`

**Fix**:
1. Move hosted runtime flags into `Flag` namespace:
   ```ts
   export const SYNERGY_HOSTED = truthy("SYNERGY_HOSTED")
   export const SYNERGY_DISABLE_WEB_MOUNT = truthy("SYNERGY_DISABLE_WEB_MOUNT")
   export const SYNERGY_SCOPE_ROOT = process.env.SYNERGY_SCOPE_ROOT
   // etc.
   ```
2. Keep `hosted.ts` focused on HTTP-layer concerns (JWT verification, request auth, directory isolation)
3. Update `truthy()` once if the broader truthy set is needed

---

### 🟡 Directory fallback bifurcates scope resolution

```ts
let directory =
  c.req.query("directory") ||
  c.req.header("x-synergy-directory") ||
  (Hosted.enabled() ? Hosted.defaultDirectory() : Flag.SYNERGY_CWD || process.cwd())
```

This creates two separate fallback chains. In hosted mode, `Flag.SYNERGY_CWD` is silently ignored even if set. The intent is correct (hosted should default to scope root), but if someone wants to run hosted with a custom working directory via `SYNERGY_CWD`, there's no way to do it.

**Fix**:
```ts
(Hosted.enabled() ? (Flag.SYNERGY_CWD || Hosted.defaultDirectory()) : Flag.SYNERGY_CWD || process.cwd())
```

---

### 🟡 Security: JWT verification lacks `iss`/`aud` checks

`Hosted.verifyJwt` validates algorithm, signature, expiry, and `user_id` match, but doesn't check `iss` (issuer) or `aud` (audience). If `SYNERGY_JWT_SECRET` is shared across services, a token issued for another service would be accepted here.

Also, `JSON.parse(decodeBase64Url(encodedHeader))` and `JSON.parse(decodeBase64Url(encodedPayload))` have no try-catch — a malformed token causes an unstructured 500 instead of a clean 401.

**Fix**:
1. Add `iss` validation (at minimum); `aud` if feasible
2. Wrap the JSON.parse calls in try-catch, throwing `"invalid_token"` on parse failure

---

### 🟡 Global `fetch` patch in `entry.tsx`

```ts
;(globalThis as { fetch: typeof fetch }).fetch = platform.fetch!
```

This patches ALL fetch calls in the runtime to add `credentials: "include"`. Third-party libraries (SolidJS router prefetch, SDK calls, etc.) will also send cookies on every request, including cross-origin ones that don't need them.

**Fix**: Either scope the patch to same-origin / `*.holosai.io` requests only, or apply `credentials: "include"` at the API client layer instead of globally. At minimum, add a comment explaining the blast radius.

---

### 🟡 `Hosted.authEnabled()` exposes undocumented `SYNERGY_AUTH_ENABLED`

```ts
export function authEnabled() {
  return enabled() || envFlag("SYNERGY_AUTH_ENABLED")
}
```

`SYNERGY_AUTH_ENABLED` is not mentioned in the PR description or README. When set without `SYNERGY_HOSTED`, auth is enforced on a local server with only `/global/health` exempt — effectively locking out the local UI since there's no control plane to issue JWTs.

**Fix**: Either document this flag with its intended use case and limitations, or remove it and require `SYNERGY_HOSTED=1` for auth.

---

### 🟢 Positive: Frontend refactor

The `app.tsx` split into `app-access.tsx` + `app-main.tsx` + `utils/runtime.ts` is clean:
- `AppWithAccess` becomes the shared entry point for hosted and local flows
- Access resolution moves from `iife` + module-level state to `createResource` + async function — better SolidJS idioms
- Hosted flow (`hosted-access.tsx`) is fully independent, no cross-contamination

---

### 🟢 Positive: Auth middleware position

The hosted auth middleware is placed **before** `provideRequestScope`, which is correct — unauthenticated requests must not trigger scope resolution side effects (like `Scope.fromDirectory` writing persisted data).

---

### 🟢 Positive: Skipping PID lock and Web mount in hosted mode

```ts
if (!Hosted.enabled()) await SingleInstance.acquire()
if (!Hosted.disableWebMount()) Server.mountApp()
```

These are appropriate runtime differences with minimal code surface and no impact on non-hosted paths.

---

### Minor issues

- **`trimSlashes` duplicated** in `utils/runtime.ts` and `hosted-access.tsx` — should be a single export from `runtime.ts`
- **`consoleOrigin()` unused** in `hosted-access.tsx` — dead code
- **`commands.splice(2, 0, ...)`** in `layout.tsx` uses a magic index — fragile if command order changes, consider id-based insertion
- **`vite.config.ts` hosted check** uses `=== "1" || === "true"` while `runtime.ts:envFlag` also accepts `"yes"` and `"on"` — inconsistent flag parsing between build-time and runtime
- **`Dockerfile.hosted`** copies entire `packages/` tree then installs; consider multi-stage build for smaller production image
- **Missing docs**: `SYNERGY_AUTH_ENABLED`, default values for `SYNERGY_AUTH_COOKIE_NAME` (`holos_jwt`) and `SYNERGY_SCOPE_ROOT` (`/workspace`)

---

### Summary of required changes before merge

1. **🔴 Fix `isGlobalRoute` expansion** — scope it to hosted mode only, or revert the expansion and find a different approach
2. **🔴 Fix `Scope.global()` directory fragility** — don't rely on `HOME == SCOPE_ROOT` coincidence
3. **🟡 Move hosted flags to `Flag` namespace** and keep `hosted.ts` focused on HTTP concerns
4. **🟡 Add JWT `iss` validation and JSON.parse error handling**

The frontend refactor and the overall direction are solid. The key issue is that hosted-mode concerns are leaking into core Scope/route semantics in ways that affect non-hosted behavior.

EricSanchezok (changes requested):
Hi, I'm glm5.1, reviewing this PR.

Overall this is a well-structured hosted deployment implementation. The app.tsx refactor, JWT verification, and scope isolation are on the right track. Below are issues I found, ordered by severity.

---

## Must Fix

### JWT verification result is not propagated to request context

`Hosted.verifyJwt()` validates `user_id === HOLOS_OWNER_ID` and returns the payload, but the payload is discarded in the middleware — it's never attached to the Hono context. This means `Hosted.resolveDirectory()` and `Scope.global()` rely entirely on the process-level `HOLOS_OWNER_ID` env var rather than the actual authenticated identity. If the env var is misconfigured or the container is ever shared across tenants, the owner check becomes meaningless.

Suggestion: `c.set("hostedAuth", payload)` after verification, and let downstream logic read the authenticated identity from context instead of env vars.

### `globalThis.fetch` is overwritten irreversibly

`entry.tsx:54` (`;(globalThis as ...).fetch = platform.fetch!`) overwrites the global fetch. This is an irreversible side effect that silently alters behavior for every fetch call in the app, including third-party SDK internals. Since `platform.fetch` is already provided through the Platform context, this global overwrite is redundant and risky.

Suggestion: Remove the `globalThis.fetch` assignment. If global credential inclusion is needed, handle it via a service worker or Hono middleware at the network layer.

### No cookie security attribute documentation

The backend reads `holos_jwt` from cookies and the frontend sends it with `credentials: "include"`, but there's no documentation or enforcement of cookie security attributes (HttpOnly, SameSite, Secure). Cookie security is the foundation of the entire auth chain.

Suggestion: Document the expected cookie attributes in README or `hosted.ts` comments. Ideally, add a SameSite check in the auth middleware.

---

## Should Fix

### `envFlag` is duplicated 3 times

Identical logic exists in:
- `packages/app/src/utils/runtime.ts:22`
- `packages/app/vite.config.ts:4`
- `packages/synergy/src/flag/flag.ts` (as `truthy`)

The vite.config one can import from the shared utility or be simplified.

### `hosted-access.tsx` is 439 lines with too many responsibilities

This single file contains API calls, a 7-state state machine, UI components, and style constants. Consider splitting into:
- API layer (fetch functions + types)
- State machine (HostedAccessGate logic)
- UI components (HostedShell, LoadingCard)

### `HostedAgentResponse.data` uses 3 different field names for the same concept

`data.attachUrl || data.attach_url || data.url` — if the control plane API really returns three naming styles, this should be documented. If not, the fallback chain hides a contract issue.

### `isGlobalRoute` expands global route scope in hosted mode without clear justification

`/path`, `/scope`, `/provider`, `/config`, `/auth` are marked as "global" routes in hosted mode but are scoped routes in non-hosted mode. This means these endpoints bypass `provideRequestScope` and always use `Scope.global()`. If provider/config need per-directory isolation in the future, this design will be a problem. Please confirm this is intentional.

### `checkHealth` polling has no backoff

Fixed 5s interval polling for up to 120s creates burst traffic to the backend when many users trigger provisioning simultaneously. A simple exponential backoff (5s → 10s → 20s) would help.

### `Dockerfile.hosted` is missing a HEALTHCHECK directive

The PR adds a health endpoint and auth exemption for it, but the Dockerfile doesn't declare a `HEALTHCHECK` instruction. Orchestrators can't auto-detect container health.

---

## Nits

- The `pollAttempt` signal is incremented but never read — it only serves to re-trigger the `createEffect`. A comment or rename (e.g., `pollTrigger`) would make the intent clearer.
- `trimSlashes` only strips trailing slashes — is this intentional? The function name suggests bidirectional trimming.

---

## What's done well

- **app.tsx refactor**: Clean separation into `app-access.tsx` + `app-main.tsx` + `utils/runtime.ts`. Non-hosted path is virtually untouched.
- **JWT verification in `hosted.ts`**: Correct use of `timingSafeEqual` for HMAC comparison, proper exp/nbf/iss/aud validation, path traversal prevention via `path.relative`.
- **`resolveAppAccess` with `createResource`**: Error states are now catchable by the Switch, which is better than the previous synchronous `defaultAccess`.
- **Hosted-aware command palette**: "Switch server" is conditionally inserted rather than deleted, preserving the non-hosted experience.
- **Welcome page hosted differentiation**: Credential failures show as errors (not warnings) in hosted mode, and "Proceed without Holos" is hidden. Correct behavior.

EricSanchezok:
## Code Review

环境：nex n2 pro

### 🔴 P0

**`truthy()` 语义变更影响所有既有 flag**

`*/flag/flag.ts` 中 `truthy()` 从只接受 `true`/`1` 扩展为 `true`/`1`/`yes`/`on`，这是向后兼容性破坏——之前将环境变量误设为 `"yes"` 或 `"on"` 的旧实例现在会被当成真值。

PR body 没有提到这个变更。要么撤回，要么明确声明并在 CI 里验证既有 flag 的旧行为仍然正确。

---

### 🟠 P1

**`packages/app/src/entry.tsx` — Request 循环访问**

```ts
const request = input instanceof Request ? new Request(input, init) : new Request(input, init)
// ...
return nativeFetch(new Request(request, { credentials: "include" }))
```

当 `input` 是 `Request` 且已消费过 body 时，这里会触发重复访问。整段意图只是加 `credentials: "include"`，直接用：`return nativeFetch(request, { credentials: "include" })` 让 fetch 自己构造即可。

另外 monkey patch `globalThis.fetch` 对一切后续调用生效，不只是控制面 API。风险较高，建议只替换受控的 API fetcher。

---

**`packages/app/src/utils/runtime.ts` + `packages/synergy/src/flag/flag.ts` 存在重复的 `envFlag`**

前端和后端各自写了一份 `envFlag()`，语义略有歧义（`"on"` 只在 backend 接受，`"yes"` 两边都接受）。建议统一到一个地方或至少在 PR body 中说明对应关系。

---

**`hosted.ts` — user_id 缺失时归为 invalid_owner，难以排查**

```ts
if (owner !== undefined && String(payload.user_id) !== String(owner)) {
  throw new Error("invalid_owner")
}
```

缺少 `user_id` 时 `undefined` 转字符串为 `"undefined"`，永远不等于真实 owner，最终只是 401。建议加显式缺失检查：`if (payload.user_id === undefined) throw new Error("missing_user_id")`。

---

**`config.ts` 跳过 node_modules 检查**

```ts
if (!Flag.SYNERGY_HOSTED) {
  // ... existsSync + installDependencies
}
```

只是延后了真实失败到 plugin import 阶段，建议让错误在 config 加载时清晰报出，而不是拖到运行时。

### 🟡 P2

**hosted auth middleware 对 HEAD 未豁免**

`isGlobalRoute` 新增的 host-checked 路由（`/path`, `/scope`, `/provider`, `/config`, `/auth`）全部会走 JWT 校验。如果 CDN 或反向代理发 HEAD 预检，会 401。建议放行 `HEAD /` 或 `/path` 的静态健康检查路由，或至少在 `exemptFromHostedAuth` 里加入 `HEAD` 豁免。

**`SingleInstance.acquire()` 完全绕过**

hosted 模式下两个实例共享同一卷时没有并发保护。应该在 README 或 Dockerfile 注释里说明 control plane 负责单实例约束，否则读者会认为这是有意放弃检测。

**`authEnabled()` 是纯转发到 `enabled()`**

```ts
export function authEnabled() { return enabled() }
```

没有独立语义，未来想拆分时会产生误导。建议注释说明意图，或直接内联。

**READMEM 中未说明 stale lockfile 处理**

README 提到"Skip lock to avoid stale Docker volume lock issues"，但容器异常退出后 lockfile 仍然存在，下次启动不会清理。需要在运维 README 或 deployment runbook 中提供清除方法。

**`hosted-access.tsx` 的 `readJson<T>` 静默吞掉所有 parse 错误**

如果控制面返回 500 但 body 不是 JSON，会直接变成 `undefined`，导致后续逻辑无法准确上报错误原因。应该在错误时返回原始错误而不是吞掉。

**`let`/`hook`/`signal` 命名风格不统一**

`hosted-access.tsx` 中函数内变量什么都用 `let`，之后 store（`setView`, `setError`, `setInviteCode`）直接混用 `createSignal`。对快照代码来说可以接受，但建议在后面迭代时把 `signals` 和命令式局部变量分离得更明确（比如 `createMemo` 或 `createSignal` for reactive state，普通 `let ref` for internal tracking）。

---

### 总结

架构方向对：HS256 JWT + path isolation + health 豁免是合理的最小设计。

关键 blocking 项：
1. **撤回或声明 `truthy()` 兼容性影响**
2. **修复 `entry.tsx` 的 Request 循环访问**（或改用受控 fetch wrapper，去掉全局 patch）
3. **`verifyJwt` 加 `user_id` 存在性检查**
4. **HEAD / 静态资源路由加入 JWT 豁免**

其余 P2 项在后续迭代处理也可以，但建议合前处理前三项。

EricSanchezok:
Hi @yuzhustudio, thanks for the thorough work on this PR. After reviewing it, we think the hosted deployment mode — including the JWT auth, scope isolation, hosted frontend flow, Dockerfile, and control plane integration — is specifically tailored to our internal Holos cloud deployment. It's not something external Synergy users would need.

Rather than merging this into the main repo, we think the cleaner approach is to maintain a fork on the Holos side with these hosted-specific changes. When Synergy upstream updates, the fork can sync accordingly. This keeps the core repo focused on general-purpose functionality and avoids carrying internal deployment concerns that only one consumer needs.

Closing this PR. Feel free to reach out if you need any help setting up the fork sync workflow.

PR #33, opened 2026-06-05
fix(compaction): remove anchor injection that causes post-compact drift

### LINKED ISSUE #32: compaction: anchor injection causes post-compact drift to session start
## Bug

After automatic compaction, the agent frequently drifts away from its current task and instead resumes the session's very first user request, ignoring all subsequent work.

## Root Cause

`buildAnchor()` in `compaction.ts` extracts the **first non-synthetic user text** from the entire message history (i.e. the session's original request) and injects it as a `<anchor>` synthetic text part into the post-compact "Continue if you have next steps" message. On subsequent compactions, the same anchor text is reused verbatim.

This anchor text reaches the LLM because:

1. `toModelMessage()` filters only `ignored` parts, not `synthetic` — so the anchor enters the model context.
2. The invoke loop's `lastUser` resolution picks the post-compact synthetic continue message (it passes `isReplyRequiredUser`), placing the anchor at the **highest recency position** in the LLM's context.
3. The anchor text ("This is the user's original request that initiated this work session") frames the session-start request as the current directive, overriding the compaction summary which correctly describes ongoing work.

The intent (from the original commit) was to prevent the agent from forgetting the original goal across compactions, but the implementation over-preserves context at the wrong granularity: it permanently freezes the session-start request at maximum recency, regardless of how the task has evolved.

## Evidence

- `compaction.ts:263-295` — `buildAnchor()` scans forward from the start for the first real user text, or reuses a previous anchor verbatim.
- `compaction.ts:437-448` — anchor is written as a second synthetic text part in the continue message.
- `message-v2.ts:467` — `toModelMessage()` passes synthetic text to LLM (only `ignored` is filtered).
- `invoke.ts:201-208` — post-compact continue message becomes `lastUser` (no `noReply` metadata).

## Proposed Fix

Remove the `<anchor>` injection entirely. The compaction summary (generated by the compaction LLM prompt: "Focus on information that would be helpful for continuing the conversation, including what we did, what we're doing, and what we're going to do next") already preserves the current task state. Adding a frozen session-start request on top of it causes more harm than benefit.

If preserving the original intent is desired in the future, the anchor should at minimum:
1. Use the **most recent real user request** instead of the session-start request.
2. Or derive from the **compaction summary's "next steps"** rather than raw history.
3. Never survive multiple compactions unchanged.

yzxoi:
## Why

After automatic compaction, the agent frequently drifts away from its current task and instead resumes the session's very first user request. This is caused by the `buildAnchor()` function injecting the session-start request as a `<anchor>` synthetic text part into the post-compact continue message, which then occupies the highest-recency position in the LLM context.

Closes #32

## What

**Commit 1 — Remove anchor injection** (`compaction.ts`):
- Delete `buildAnchor()` function and its `ANCHOR_OPEN`/`ANCHOR_CLOSE` constants
- Remove the anchor injection block in the auto-compaction continue message creation
- The compaction summary (generated by the LLM summarization prompt) already preserves the current task state, making the anchor redundant

**Commit 2 — Defensive: exclude synthetic user text from model messages** (`message-v2.ts`):
- Add `!part.synthetic` check in `toModelMessage()` for user text parts
- Previously only `ignored` was filtered; synthetic text parts (like the "Continue if you have next steps" prompt) were passed through to the LLM as real user input
- This prevents any future synthetic content from leaking into model context

**Commit 3 — Update test** (`message-v2.test.ts`):
- Update the `toModelMessage` test to verify synthetic user text is now excluded
- Renamed from "includes synthetic text parts" to "excludes synthetic user text parts from model messages"

## Tests

- All 57 tests in `compaction.test.ts`, `message-v2.test.ts`, and `invoke-compaction.test.ts` pass
- The failing test was the one that asserted synthetic parts were included — updated to match the new correct behavior

## Out of scope

- No changes to compaction summary quality or the compaction LLM prompt
- No changes to how the invoke loop resolves `lastUser` (the continue message is still the last user, but now it only contains "Continue if you have next steps" without the anchor override)

### DIFF
--- a/packages/synergy/src/session/compaction.ts
+++ b/packages/synergy/src/session/compaction.ts
@@ -251,49 +251,6 @@ export namespace SessionCompaction {
     }
   }
 
-  const ANCHOR_OPEN = "<anchor>"
-  const ANCHOR_CLOSE = "</anchor>"
-
-  /**
-   * Build the anchor text that preserves the user's original request across
-   * compaction boundaries.  If a previous compaction already embedded an
-   * anchor in one of the continue messages, reuse it verbatim so the same
-   * text survives any number of compactions without degradation.
-   */
-  function buildAnchor(messages: MessageV2.WithParts[]): string | undefined {
-    // Check if a previous compaction already embedded an anchor.
-    for (let i = messages.length - 1; i >= 0; i--) {
-      const msg = messages[i]
-      if (msg.info.role !== "user") continue
-      for (const part of msg.parts) {
-        if (part.type === "text" && part.synthetic && part.text.startsWith(ANCHOR_OPEN)) {
-          return part.text
-        }
-      }
-    }
-
-    // First compaction — extract the original user request.
-    for (const msg of messages) {
-      if (msg.info.role !== "user") continue
-      const textParts = msg.parts.filter((p): p is MessageV2.TextPart => p.type === "text" && !p.synthetic)
-      if (textParts.length === 0) continue
-      const text = textParts
-        .map((p) => p.text)
-        .join("\n")
-        .trim()
-      if (!text) continue
-      return [
-        ANCHOR_OPEN,
-        "This is the user's original request that initiated this work session.",
-        "",
-        text,
-        ANCHOR_CLOSE,
-      ].join("\n")
-    }
-
-    return undefined
-  }
-
   export async function process(input: {
     parentID: string
     messages: MessageV2.WithParts[]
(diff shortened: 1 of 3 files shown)


EricSanchezok:
这个 bug 的根本原因不是 anchor 本身，而是 anchor 指向了错误的消息。PR #34 修正了 `buildAnchor()`：从倒序扫描取 compaction 触发前最后一条真实 user 消息，而不是正向扫描取会话第一条。

Anchor 本身是个有价值的机制——compaction 后用 anchor 保留最近的工作上下文，agent 就能记住刚才在做什么。PR #33 删除 anchor 的做法过重了，anchor 指向正确的消息时就能正常工作了。

目前已 merged 到 dev。

PR #50, opened 2026-06-10
fix(provider): guard reasoningEffort/textVerbosity behind native npm check, fix openrouter smallOptions shape

marzich:
## Summary

- **`reasoningEffort` + `reasoningSummary` now gated behind native npm check** — previously injected unconditionally for any `gpt-5` model, causing `400` errors when using `@ai-sdk/openai-compatible` providers (e.g. LiteLLM proxies) because `/v1/chat/completions` rejects `reasoning_effort` when tools are present. Now only injected when npm is `@ai-sdk/openai`, `@ai-sdk/azure`, or `@ai-sdk/github-copilot`.

- **`textVerbosity` similarly gated** — this is a Responses API-specific field; added the same npm guard to prevent it being sent to incompatible proxy backends.

- **Fix openrouter `smallOptions` shape** — `reasoningEffort: "minimal"` → `reasoning: { effort: "minimal" }` to match the actual openrouter API contract.

## Root Cause

`reasoningSummary` at line ~605 already had the npm guard (added with a comment about openai-compatible proxies), but `reasoningEffort` directly above it did not — an inconsistency that caused silent 400s for any user who configured a `gpt-5.x` model through an openai-compatible proxy.

## Test

Typecheck and formatting pass (`bun turbo typecheck` + Prettier).

### DIFF
--- a/packages/synergy/src/provider/transform.ts
+++ b/packages/synergy/src/provider/transform.ts
@@ -599,24 +599,27 @@ export namespace ProviderTransform {
 
     if (model.api.id.includes("gpt-5") && !model.api.id.includes("gpt-5-chat")) {
       if (!model.api.id.includes("codex") && !model.api.id.includes("gpt-5-pro")) {
-        result["reasoningEffort"] = "medium"
-        // Only inject reasoningSummary for providers that support it natively.
-        // @ai-sdk/openai-compatible proxies (e.g. LiteLLM) do not understand this parameter.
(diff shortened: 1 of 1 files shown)


EricSanchezok (approved):
Thanks for the fix! The npm guard on reasoningEffort makes complete sense — unconditionally injecting it for all gpt-5 models was silently breaking openai-compatible proxies. OpenRouter smallOptions shape fix is also correct.

Merging this first since it's clean and independent of the other PRs.

PR #51, opened 2026-06-10
fix(ui): add execCommand fallback for clipboard in insecure contexts

marzich:
## Problem

`navigator.clipboard.writeText()` requires a [Secure Context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS or `localhost`). When accessing Synergy over a plain HTTP address such as a Tailscale IP (`http://100.x.x.x:18080`), all three copy buttons silently fail:

- Code block **Copy** button
- KaTeX formula click-to-copy
- `TextField` copyable button

## Solution

Add a `copyToClipboard()` helper in `markdown.tsx` that:
1. Uses `navigator.clipboard.writeText()` when `window.isSecureContext` is true (HTTPS / localhost) — no behaviour change for secure contexts
2. Falls back to the `textarea` + `document.execCommand('copy')` pattern for non-secure contexts — works as long as the call is triggered by a user gesture (button click), which it always is here

Apply the same inline fallback to `TextField.handleCopy()`.

## Changes

- `packages/ui/src/components/markdown.tsx` — new `copyToClipboard()` helper; replace direct `navigator.clipboard` calls in `handleKatexClick` and `handleClick`
- `packages/ui/src/components/text-field.tsx` — inline fallback in `handleCopy`

### DIFF
--- a/packages/ui/src/components/markdown.tsx
+++ b/packages/ui/src/components/markdown.tsx
@@ -11,6 +11,30 @@ const max = 200
 const cache = new Map<string, Entry>()
 const copyResetDelay = 2000
 
+/** Copy text to clipboard, falling back to execCommand for insecure contexts (http://IP, tailscale http) */
+async function copyToClipboard(text: string): Promise<boolean> {
+  if (navigator.clipboard && window.isSecureContext) {
+    await navigator.clipboard.writeText(text)
+    return true
+  }
+  // Fallback for HTTP / non-secure contexts (user-gesture-triggered only)
+  try {
+    const ta = document.createElement("textarea")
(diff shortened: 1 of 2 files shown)


EricSanchezok (approved):
Thanks! The execCommand fallback is the right approach for insecure contexts. I rebased this onto dev after #50 was merged, and the diff is now clean (only clipboard changes, no provider transform overlap).

One observation: the text-field.tsx fallback is inline while markdown.tsx has `copyToClipboard()`. Consider extracting into a shared util in a future cleanup — but for two call sites it's fine as-is.

Merging!

PR #52, opened 2026-06-11
fix(ui): show agent and model selectors on mobile prompt toolbar

marzich:
## Problem

On mobile viewports (< 768px), the agent selector ("Synergy Max ▼") and model selector ("DeepSeek V4 Flash ▼") in the prompt input toolbar were completely invisible. They were wrapped in `hidden md:contents`, which hides them on any screen narrower than 768px — leaving mobile users with no way to switch agents or models from the prompt area.

## Solution

Add a compact mobile-only toolbar (`flex md:hidden`) rendered just before the existing desktop toolbar inside `<Match when={store.mode === "normal"}>`.

**Mobile toolbar (new):**
- **Agent selector** — an `AgentGlyph` emoji circle button (size-7) that opens the same `ToolbarSelectorPopover` agent list
- **Model selector** — a compact pill showing the truncated model name (`max-w-20 truncate`) that opens `ModelSelectorPopover`

**Desktop toolbar (unchanged):** The existing `hidden md:contents` block with full-width "Synergy Max ▼" + "DeepSeek V4 Flash ▼" labels is untouched.

## Changes

- `packages/app/src/components/prompt-input.tsx`
  - Add `AgentGlyph` to the `agent-visual` import
  - Restore accidentally removed `getDirectory` / `getFilename` import from `synergy-util/path`
  - Insert ~64-line mobile compact toolbar block

## Testing

- TypeScript typecheck: ✅ all 9 packages pass (turbo typecheck)
- Prettier: ✅ no formatting violations
- Manually verified on mobile viewport: both selectors appear and open their respective popovers correctly

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -40,7 +40,7 @@ import { IconButton } from "@ericsanchezok/synergy-ui/icon-button"
 import { Tooltip, TooltipKeybind } from "@ericsanchezok/synergy-ui/tooltip"
 import { List } from "@ericsanchezok/synergy-ui/list"
 import { ToolbarSelectorPopover } from "@/components/toolbar-selector"
-import { getAgentVisual } from "@/components/agent-visual"
+import { getAgentVisual, AgentGlyph } from "@/components/agent-visual"
 import { getDirectory, getFilename } from "@ericsanchezok/synergy-util/path"
 import { useDialog } from "@ericsanchezok/synergy-ui/context/dialog"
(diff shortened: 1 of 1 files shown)


EricSanchezok (approved):
Thanks Zeyi! Clean fix — the mobile toolbar was completely missing before this. A few notes:

1. I changed the base branch from `main` to `dev` per repo convention.
2. The failing test (Cortex > output integration) is a pre-existing flaky test, unrelated to this UI change.
3. The agent list items on mobile don't show emoji (only label text) — intentional for space, and works fine.

Merging!

PR #61, opened 2026-06-16
fix(note): restrict scope:all to current+global, prevent cross-project metadata leak and search crash

marzich:
## Summary

- `note_list` and `note_search` with the default `scope: "all"` were calling `NoteStore.listMetaAll()`, which scans **every scope directory** under `~/.synergy/data/notes/` — returning notes from all projects on disk, not just the current one.
- This caused two bugs:
  1. **Metadata leak** — `note_list` silently exposed note titles, tags, and timestamps from unrelated projects.
  2. **Search crash** — `note_search` enumerated notes from other project scopes via `listMetaAll()`, then tried to load their content using `getAny(currentScopeID, id)`. Since `resolveScope` only checks the current scope and global, notes from other scopes threw an unhandled `NotFoundError`, crashing the entire search tool.

## Fix

Added `NoteStore.listMetaWithGlobal(scopeID)` — a metadata-only counterpart to the already-existing `listWithGlobal()` — that merges only the current project scope and the global scope. Both `note_list` and `note_search` now call this instead of `listMetaAll()` when `scope: "all"`. `listMetaAll()` is preserved for explicit full-system use cases (e.g. UI global views).

Also corrected the `note_list` parameter description which claimed `"all"` = "current + global" while the code was returning all projects.

## Files Changed

- `packages/synergy/src/note/store.ts` — add `listMetaWithGlobal()`
- `packages/synergy/src/tool/note-list.ts` — use `listMetaWithGlobal`, fix description
- `packages/synergy/src/tool/note-search.ts` — use `listMetaWithGlobal`


EricSanchezok:
Simple and effective fix — the scope:all boundary was a real data leak risk. The listMetaWithGlobal() mirror of listWithGlobal() is the right pattern. Thanks @marzich!

PR #62, opened 2026-06-16
fix(agenda): support isolated recurring runs and harden scheduling

marzich:
## Problem

Agenda recurring triggers (`cron`, `every`, `watch`) always inferred a persistent session. That is useful for stateful monitors, but it is wrong for independent scheduled jobs such as daily/weekly reports:

- repeated runs reuse the same session history;
- prior errors/refusals can pollute future runs;
- timer-fired runs previously executed outside an `Instance` scope and could fail with `No context found for instance`;
- some failure paths could remove the clock entry or leave runs hard to diagnose;
- the user-facing agenda tools did not expose enough execution configuration to route scheduled work to a specific agent/model.

## Solution

This PR adds explicit agenda execution controls and hardens the runtime lifecycle while preserving existing defaults.

### Session isolation

- Adds optional `sessionMode: "ephemeral" | "persistent"` on agenda items.
- Existing default inference is unchanged:
  - recurring triggers default to `persistent`;
  - one-shot triggers default to `ephemeral`.
- Setting `sessionMode: "ephemeral"` on a cron/every item creates a fresh session on every fire.

### Runtime hardening

- Runs timer-triggered agenda execution inside `Instance.provide(item.origin.scope)` so plugin/session code has a valid scope.
- Makes plugin hook failures non-fatal to the agenda state machine.
- Ensures unexpected reactor errors compute a fallback `nextRunAt` when possible instead of permanently deleting the clock entry.
- Recomputes missing `nextRunAt` during clock sync from triggers instead of silently unloading active items.
- Restores `delay` triggers as one-shot: create/update can anchor a relative delay, but post-run recomputation does not turn it into a loop.

### Tool/API surface

- `agenda_schedule` now exposes `agent`, `model`, `timeout`, `sessionMode`, and `sessionRefs`.
- `agenda_update` now exposes `triggers`, `agent`, `model`, `timeout`, `sessionMode`, and `sessionRefs`.
- `PatchInput` and `AgendaStore.update()` now persist `model` and `timeout` updates.
- `agenda_update.txt` now documents in-place trigger updates.

### Ephemeral session hygiene

- Ephemeral agenda sessions are titled with a short date, e.g. `Agenda: Daily diary 2026-06-17`.
- Ephemeral agenda sessions are archived after the run so they remain traceable by `sessionID` / logs but do not accumulate in normal session lists.

## Usage

```ts
agenda_schedule({
  title: "Daily diary",
  trigger: { type: "cron", expr: "0 4 * * *", tz: "Asia/Shanghai" },
  agent: "daily-diary-agent",
  sessionMode: "ephemeral",
  silent: true,
  wake: false,
  prompt: "...",
})
```

For existing items:

```ts
agenda_update({
  id: "agd_xxx",
  agent: "daily-diary-agent",
  sessionMode: "ephemeral",
  silent: true,
  wake: false,
})
```

## Compatibility

- Existing agenda items keep their previous behavior unless `sessionMode` or execution config fields are explicitly set.
- Persistent recurring jobs still reuse their session by default.
- One-shot delay/at jobs remain one-shot.

## Validation

- `bun run typecheck` passes.
- Local dev validation confirmed:
  - timer-fired agenda runs no longer hit `No context found for instance`;
  - `sessionMode: "ephemeral"` creates a new session per fire;
  - `agent: "daily-diary-agent"` is persisted and routed into `SessionInvoke`;
  - delay test fires once and transitions to done;
  - daily/weekly diary items rearm to their next expected times.

### DIFF
--- a/packages/synergy/src/agenda/index.ts
+++ b/packages/synergy/src/agenda/index.ts
@@ -143,8 +143,12 @@ export namespace Agenda {
   }
 
   function syncItem(scopeID: string, item: AgendaTypes.Item): void {
-    if (item.state.nextRunAt !== undefined) {
-      AgendaClock.rearm(scopeID, item.id, item.state.nextRunAt)
+    // If nextRunAt is missing (e.g. after a patch that didn't touch triggers),
+    // recompute from the item's triggers rather than silently unloading the
+    // clock entry. Only unload if the triggers genuinely produce no future run.
+    const nextRunAt = item.state.nextRunAt ?? AgendaStore.computeNextRunAt(item.triggers)
+    if (nextRunAt !== undefined) {
+      AgendaClock.rearm(scopeID, item.id, nextRunAt)
     } else {
       AgendaClock.unload(item.id)
     }
--- a/packages/synergy/src/agenda/reactor.ts
+++ b/packages/synergy/src/agenda/reactor.ts
@@ -140,6 +140,18 @@ export namespace AgendaReactor {
       if (sessionID && !error) {
         lastMessage = await extractLastAssistantMessage(sessionID)
       }
+      // Archive ephemeral sessions so they don't accumulate in the session list.
+      // Do this after extracting the last message but before state management.
+      if (sessionID && !persistent) {
+        await Session.update(sessionID, (draft) => {
+          draft.time.archived = Date.now()
+        }).catch((err) => {
+          log.warn("failed to archive ephemeral session", {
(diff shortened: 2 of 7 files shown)


EricSanchezok:
This is really well-engineered — ephemeral vs persistent session mode with proper scope isolation, clock hardening, and the archived-session hygiene. Clean compatibility story too. Thanks @marzich!

PR #123, opened 2026-06-22
feat(plugin): Platform v3/v4/v4.1 — security, runtime isolation, consent, marketplace, toolchain, docs

EricSanchezok:
## Summary

Full Synergy Plugin Platform upgrade from rough capability extension to production-grade platform.

### v3 (Security Baseline + Platform Architecture)
- **Plugin security baseline**: PluginToolId utility, enforcement gate capability decomposition (9 sub-capabilities per tool), control profile no longer auto-allows plugin_invoke in autonomous mode, auth store plaintext warnings, baseline CSP headers
- **Manifest v3**: Expanded permissions (tools/data/network/ui/hooks), tool capability declarations, trust tier schema
- **Loader refactor**: Split 612-line monolith into 6 modules (store/loader/lifecycle/install/status/trust), lockfile SHA-256 integrity, local plugin reload cache-busting, hook scope gating, dependency auto-install
- **Server routes**: /api/plugins routes with OpenAPI + SDK generation, asset route hardening (path containment, symlink, MIME, SVG sanitizer), enriched PluginStatus API
- **Frontend PluginHost**: PluginHostProvider + ToolRendererRegistry with lazy load/fallback signal, trusted-import + sandbox-iframe tiers

### v4 (Runtime Isolation + Consent + Marketplace)
- **Runtime isolation**: PluginRuntimeSupervisor, process/worker/in-process modes, IPC protocol, process-host/worker-host, heartbeat monitoring, resource limits (concurrency, memory, log rate)
- **Consent engine**: Permission diff/risk/summary/approval store, install/update consent gating, audit log, policy config
- **Consent UI**: InstallConsentDialog, UpdateConsentDialog, PermissionDiffList, RiskBadge, TrustTierExplanation
- **Marketplace**: Registry API (search/detail/versions/download/publish), marketplace frontend (search, detail, install/update buttons, verified/risk badges, changelog tab, downloads/rating)
- **Toolchain CLI**: 12+ commands (create/dev/validate/test/build/pack/sign/publish/info/permissions/approve/runtime)
- **Bridge handlers**: Process-mode bridge execution with path containment, URL validation, shell timeout, authorization stripping

### v4.1 (Runtime Registry Refactor + Developer Docs)
- **RuntimeRegistry class**: Replaced implicit singleton Map with explicit injectable class, eliminated all global state workarounds
- **PluginRuntimeSupervisor class**: Injectable registry/logs/persist dependencies, process/worker hosts are now stateless callback adapters
[...]

PR #124, opened 2026-06-23
Optimize prompt cache boundary handling

EricSanchezok:
## Summary
- preserve stable system prompt blocks through prompt budgeting and LLM streaming
- add an explicit stable system cache breakpoint for Anthropic cache control
- record DeepSeek/OpenAI-compatible prompt cache hit/miss metadata in usage accounting

## Tests
- bun test test/provider/transform.test.ts test/session/compaction.test.ts
- git diff --check

## Notes
- bun run --cwd packages/synergy typecheck currently fails on an unrelated existing issue: src/browser/install.ts(119): unused @ts-expect-error directive.

### DIFF
--- a/packages/synergy/src/provider/transform.ts
+++ b/packages/synergy/src/provider/transform.ts
@@ -22,6 +22,10 @@ export namespace ProviderTransform {
     tool?: string
   }
 
+  interface MessageOptions {
+    systemCacheBreakpoint?: number
+  }
+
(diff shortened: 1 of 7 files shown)


EricSanchezok:
This PR has been integrated and conflict-resolved in #127 together with #123, #125, and #126. Please review/merge #127 as the final stack instead of merging this branch directly.

PR #138, opened 2026-06-25
fix: relax CSP to allow Ghostty Web WASM terminal

yzxoi:
## Summary

The built-in terminal uses Ghostty Web (ghostty-web 0.3.0), a WebAssembly-based
VT100 emulator. The previous Content-Security-Policy blocked it at four levels,
and the script hash/nonce mechanism inadvertently disabled `'unsafe-inline'`.

### 1. Missing CSP directives
Ghostty Web loads WASM, base64-embedded fonts, and alert sounds that were
all blocked:

| Directive     | Before                 | After                                        | Blocked resource                    |
|---------------|------------------------|----------------------------------------------|-------------------------------------|
| `script-src`  | `'self'` (+hash)       | `'self' 'unsafe-inline' 'wasm-unsafe-eval'`  | `WebAssembly.instantiate()` + scripts |
| `connect-src` | `'self' ws: wss:`      | `'self' ws: wss: blob: data:`                | `fetch("data:application/wasm;...")` |
| `font-src`    | `'self'`               | `'self' data:`                               | base64 monospace fonts              |
| `media-src`   | `'none'`               | `'self'`                                     | terminal bell sounds                |

### 2. `'unsafe-inline'` was silently ineffective
`spaCsp()` always injected a SHA-256 hash of the theme preloader script into
`script-src`. Per CSP Level 2, browsers ignore `'unsafe-inline'` when hash or
nonce is present in `script-src`. So Ghostty's dynamically-created scripts were
blocked regardless of the `'unsafe-inline'` keyword.

## Changes
- Relax CSP directives to support WASM, data URIs, and media
- Drop `CSP_THEME_SCRIPT_HASH` and nonce extension; `spaCsp()` returns baseline
- Remove nonce from route-tag `<script>` (dead code after CSP change)
- Copy `ghostty-vt.wasm` into `packages/app/public/` for dist shipping

## Files
- `packages/synergy/src/server/server.ts` — CSP baseline + `spaCsp()` cleanup
- `packages/app/public/ghostty-vt.wasm` — new file, static WASM fallback


EricSanchezok:
Thanks @yzxoi for the fix here. I merged this into dev with the Ghostty Web WASM terminal CSP relaxation intact, and added the matching server CSP test update while resolving the current dev conflict. Appreciate you tracking down the CSP requirements for Ghostty.

PR #155, opened 2026-06-29
Fix plugin task permission hash generation

EricSanchezok:
## Summary

- include `permissions.tools.task` in plugin-kit base capability hashing
- treat delegated task permission as medium risk in plugin-kit and runtime consent risk
- add a regression test that compares runtime and plugin-kit capabilities for task-enabled plugin manifests

## Why

The runtime installer already counted delegated task permission as `task`, but plugin-kit did not. That let marketplace artifacts be signed with a permissions hash that official install verification later rejected.

## Validation

```bash
cd packages/synergy
bun test test/plugin/capability-consistency.test.ts test/plugin/consent.test.ts test/plugin/scaffold-toolchain.test.ts
bun run typecheck
cd ../plugin-kit
bun run typecheck
```

Also verified in the main checkout that `synergy-meme-plugin@0.3.2` passes `PluginMarketplaceRegistry.verifyOfficialArtifact` from the official registry.

### DIFF
--- a/packages/app/src/components/library/shared.tsx
+++ b/packages/app/src/components/library/shared.tsx
@@ -105,21 +105,17 @@ export const DISCRETE_DIMENSIONS: Array<{ key: keyof RewardsInfo; short: string;
   { key: "expression", short: "Exp", full: "Expression" },
 ]
 
-export const libraryShellClass =
-  "library-main-surface"
+export const libraryShellClass = "library-main-surface"
 
-export const libraryInsetClass =
-  "library-inner-surface"
+export const libraryInsetClass = "library-inner-surface"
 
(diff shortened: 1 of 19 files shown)


EricSanchezok:
Closing this PR because the delegated task permission hash fix has already been absorbed by the current `dev` branch through the shared capability model in `@ericsanchezok/synergy-util/capability`.

I verified the current branch already covers the intended behavior with:

```bash
cd packages/synergy
bun test test/plugin/capability-consistency.test.ts test/plugin/consent.test.ts

cd ../plugin-kit
bun run typecheck
```

Both pass, and keeping this PR open would require resolving conflicts against old plugin-kit/runtime risk files that no longer exist on `dev`.

PR #157, opened 2026-06-29
feat: allow Tab key to select slash command in popover

SII-jake:
## Summary

Currently, pressing `Enter` in the slash command popover selects the highlighted command. But `Enter` is also the primary send-message shortcut, which makes it easy to accidentally send unfinished text.

This change adds `Tab` as an alternative to `Enter` for selecting a command from the slash popover — `Tab` always selects, never sends.

## Changes

- **`packages/ui/src/hooks/use-filtered-list.tsx`** — Added `key === "Tab"` alongside `"Enter"` in the `onKeyDown` handler to trigger selection
- **`packages/app/src/components/prompt-input.tsx`** — Added `"Tab"` to the popover key dispatch so it routes to the slash command handler instead of defaulting to browser Tab behavior (focus shift)


EricSanchezok:
Thanks for the UX improvement here. I implemented the core behavior directly on `dev` in commit d36843ca47031d8857bec861766b07a9be56ed61: pressing `Tab` while the prompt popover is open now routes through the same selection path as `Enter`.

I’m closing this PR rather than merging it because the branch also includes unrelated earlier changes from the `/merge-dev` command work and the `/global/health` channel field, which we decided not to merge. The useful part of this PR was the two-file Tab selection change, and that is now present on `dev` without the unrelated changes.

PR #183, opened 2026-07-02
Stabilize session inbox and progress visibility

EricSanchezok:
## Summary
- Preserve session-scoped volatile snapshots (`inbox`, `todo`, `dag`) across normal session switches instead of evicting them and briefly rendering unloaded data as empty.
- Centralize session volatile refreshes in sync/global sync so active sessions and reconnects refresh in place without clearing previous UI state.
- Simplify `SessionProgressIsland` visibility by removing the idle/cortex lifecycle heuristic and stale fingerprint suppression; progress now reflects loaded DAG/Todo snapshots directly.
- Fix the UI package's missing `@happy-dom/global-registrator` devDependency that was blocking the pre-push typecheck hook.

## Problem
Switching sessions made the UI temporarily lose session-scoped volatile state:

1. `pages/session.tsx` called `sync.evictSession(prevId)` whenever the route session id changed.
2. `evictSession` deleted `inbox`, `todo`, and `dag` snapshots for the session being left.
3. When switching back, the new active session had to refetch those snapshots asynchronously.
4. Components collapsed `undefined` into `[]`, so "not loaded yet" rendered as "loaded and empty":
   - inbox count/popover could flash to `0` / `Inbox clear`;
   - progress could compute `mode = none` and disappear until DAG/Todo reloaded.

`SessionProgressIsland` also had several independent visibility layers on top of that:

- missing `session_status` defaulted to `idle`;
- idle plus no active cortex task could mark an incomplete DAG as `settled` and hide it;
- `lastHiddenFingerprint` could suppress the same DAG from showing again after a hidden decision;
- the component also owned its own DAG/Todo fetch effect while the page/global sync layer also handled session data.

Together these made visibility feel inconsistent: sometimes the island disappeared because data was genuinely empty, sometimes because data was merely unloaded, sometimes because idle/cortex was guessed, and sometimes because the fingerprint suppression remembered a prior hidden state.

### Session sync / refresh ownership
- Removed normal route-switch volatile eviction from `packages/app/src/pages/session.tsx`.
- Added explicit refresh helpers in `packages/app/src/context/sync.tsx`:
  - `session.inbox(sessionID, { force })`
  - `session.todo(sessionID, { force })`
  - `session.dag(sessionID, { force })`
  - `session.refreshVolatile(sessionID)`
- `sync.session.sync(id, { refreshVolatile: true })` now refreshes inbox/todo/dag without clearing existing snapshots first.
- Refresh failures intentionally leave the previous snapshot in place, avoiding false empty states.
- Reconnect/global resync now refreshes retained volatile session snapshots in `global-sync.tsx` without clearing them first.

### Inbox UI semantics
- Added `deriveSessionInboxView()` so inbox rendering has an explicit state model:
  - `undefined` => loading/unknown
  - `[]` => loaded empty
  - non-empty array => ready with badge count
- Updated `SessionInbox` to show `Loading inbox…` for unknown state and `Inbox clear` only after a loaded empty snapshot.

### Progress island simplification
- Removed `ProgressLifecycle` from `session-progress-summary.ts`.
- Removed the summary-level settled/paused hiding rule.
- Removed `SessionProgressPanel`'s `session_status`/cortex idle-as-settled inference.
- Removed `dagFingerprint` / `lastHiddenFingerprint` stale suppression.
- Removed component-owned DAG/Todo fetch effect; the page/sync layer now owns refreshes.
- Kept only presentation-local state in the panel: tab, expanded/collapsed, visible/exiting, and completion auto-hide timer.

This means the island now appears or hides from one primary source of truth: the loaded DAG/Todo snapshots. Unknown data is not treated as empty progress, and incomplete DAG work stays visible instead of being hidden by frontend idle guesses.

## Why this approach
The goal was not to add another guard on top of the old behavior. The root problem was multiple overlapping meanings for the same UI state:

- `undefined` and `[]` were both rendered as empty;
- navigation was destructive and then immediately refetched;
- progress visibility mixed actual progress data with session status, cortex tasks, stale fingerprints, timers, and local fetch side effects.

The new model keeps the same existing store maps but makes their meaning explicit:

```ts
store.inbox[sessionID] === undefined // not loaded yet
store.inbox[sessionID] === []        // loaded and empty
store.inbox[sessionID] === [...]     // loaded with items
```

The same principle applies to `todo` and `dag`. Refreshes update snapshots in place, while components render loaded state rather than guessing freshness. This removes old suppression/eviction paths instead of hiding them behind extra compatibility logic.

## Verification
- `bun ./script/format.ts`
- `cd packages/app && bun test src/components/session/session-inbox.test.ts src/components/session/session-progress-summary.test.ts`
  - 30 tests passed
- `cd packages/app && bun run typecheck`
- `bun run typecheck`
  - 10/10 packages passed
- Pre-push hook passed:
  - full `bun turbo typecheck`
  - formatting check

## Review
- Maintainability review: PASS, no blockers.
- Noted non-blocking future cleanup: if more retained volatile surfaces are added, the global reconnect refresh helper and scoped sync refresh helpers may be worth extracting into a shared helper to avoid drift.

## Risk / behavior notes
- Retaining inbox/todo/dag snapshots across session switches increases in-memory retention for visited sessions, but these snapshots are small compared with message/part history and avoid destructive cache churn.
- If the backend leaves stale non-terminal DAG nodes behind, the frontend will now show them instead of hiding them through idle/cortex inference. That is intentional: stale DAG ownership should be fixed at the source rather than hidden by frontend guesses.
- Refresh failures are intentionally quiet and preserve the previous snapshot. This matches the UI stability goal; repeated refresh failure telemetry can be added later if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,7 +416,9 @@ Control profiles are configured in the permissions domain (`80-permissions.jsonc
 }
 ```
 
-**Precedence:** agent config `controlProfile` > top-level config `controlProfile` > default `guarded`.
+**Precedence:** explicit session control profile resolved from the session parent chain > agent config `controlProfile` > top-level config `controlProfile` > source default. Channel-backed and agenda automation root sessions default to `autonomous`; ordinary interactive/manual sessions default to `guarded`.
+
+Explicit configuration is always honored. For example, a top-level `controlProfile: "full_access"` remains `full_access` for Feishu/channel sessions instead of being downgraded by unattended metadata.
 
 Blueprint runs started from the Notes side panel use the current session's control profile when running in the current session. New-session and worktree Blueprint runs create an execution session with at least `autonomous`; a top-level `full_access` profile remains `full_access`.
 
@@ -426,13 +428,11 @@ Risk levels follow the operation's effect. Ordinary reads, including non-protect
 
 Built-in profiles:
 
-| Config value  | UI label    | Behavior                                                                                                                                                                               |
-| ------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `guarded`     | Guarded     | Default protected mode. Auto-allows ordinary reads, workspace-local edits, and ordinary network lookups; asks before shell, external writes, identity, platform, or extension actions. |
-| `autonomous`  | Autonomous  | Unattended mode. Includes Guarded's automatic approvals, allows medium-risk development work, and denies high-risk asks instead of prompting.                                          |
-| `full_access` | Full Access | Allows all tool requests without approval prompts or workspace sandboxing.                                                                                                             |
-
-`full_access` is blocked in unattended execution mode. It is only available in attended sessions.
+| Config value  | UI label    | Behavior                                                                                                                                                                                           |
+| ------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `guarded`     | Guarded     | Protected mode for manual sessions. Auto-allows ordinary reads, workspace-local edits, and ordinary network lookups; asks before shell, external writes, identity, platform, or extension actions. |
+| `autonomous`  | Autonomous  | Automation-oriented profile. Includes Guarded's automatic approvals, allows medium-risk development work, and denies high-risk asks instead of prompting.                                          |
+| `full_access` | Full Access | Allows all tool requests without approval prompts or workspace sandboxing. Explicit `full_access` remains valid for channel and automation sessions.                                               |
 
 ### Sandbox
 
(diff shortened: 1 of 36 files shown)


yzxoi:
Found one small regression while reviewing: route changes stopped clearing the page-local hydration/initial-scroll guards, so switching away and back could skip turn-window/scroll initialization for an already visited session. I pushed a minimal fix that restores only those local guard resets while keeping inbox/todo/dag snapshots retained. Also fixed the plugin signing key type that was blocking the pre-push typecheck hook.

yzxoi:
Nice work untangling this slippery bit of session UI state — the switch-back behavior feels much steadier now. I pushed a tiny follow-up for the page-local init guard and pre-push typecheck issue; with that in place, we’ll merge this. 🙌

yzxoi (approved):

PR #187, opened 2026-07-02
Prewarm image attachment pipeline

EricSanchezok:
## Summary

This PR reduces the long first-image-paste pause in the Desktop session input by pre-warming the same browser image pipeline that prompt attachments use for thumbnails.

## Problem

When Desktop is opened and the user pastes an image into a session for the first time, the input can appear stuck for roughly 10 seconds. Later image pastes are fast.

The paste path is not just reading from the clipboard. A pasted image is turned into a prompt attachment by:

1. reading/normalizing the file, including PNG cICP stripping when needed;
2. uploading the original image as a local asset;
3. decoding the image through `new Image()` and a blob URL;
4. drawing the decoded image into a canvas;
5. encoding a WebP thumbnail with `canvas.toBlob("image/webp", 0.78)`;
6. uploading the thumbnail as a second local asset.

The first real paste therefore pays several cold-start costs in the Electron/Chromium renderer: image decode initialization, Canvas 2D/GPU surface setup, and WebP encoder initialization. Subsequent pastes are faster because those browser paths are already warm.

The backend asset write path is comparatively lightweight and already initialized by server startup. Desktop also does not use a native `clipboard.readImage()` IPC path here, so the most direct fix is on the frontend attachment image pipeline.

## Why this approach

Instead of changing attachment semantics or delaying thumbnail generation, this PR moves the expensive one-time browser initialization off the first real paste interaction.

The warmup is:

- best-effort: failures are swallowed and never shown to the user;
- tiny: it uses 1x1 canvases and an in-memory seed image;
- local-only: it does not call the asset upload client and does not persist anything;
- idempotent: a module-level promise prevents repeated warmups;
- scheduled after startup using `requestIdleCallback` when available, with a short timeout fallback.

This directly targets the user-visible symptom while keeping the existing attachment upload behavior unchanged.

## What changed

- Added `warmPromptAttachmentImagePipeline()` in `packages/app/src/utils/prompt-attachment.ts`.
  - Allocates a 1x1 canvas and gets a 2D context.
  - Encodes a tiny PNG seed blob.
  - Decodes that seed via the existing `loadImage()` path.
  - Draws the decoded image into another 1x1 canvas.
  - Encodes a WebP blob using the same thumbnail MIME/quality as real attachments.
- Added `schedulePromptAttachmentImagePipelineWarmup()`.
  - Runs the warmup after startup during idle time when possible.
  - Falls back to a short timer if `requestIdleCallback` is unavailable.
- Wired the scheduler from `packages/app/src/entry.tsx` after the app render call.
- Strengthened `prompt-attachment.test.ts` to verify the warmup:
  - does not upload or persist assets;
  - creates/revokes an object URL, exercising the decode path;
  - draws to canvas;
  - calls `toBlob` for both the PNG seed and WebP thumbnail pipeline.
- Declared `@happy-dom/global-registrator` for `packages/ui` because the repository pre-push typecheck compiles `packages/ui/src/components/special-user-message.test.tsx`, which already imports it. The package was already present in the lockfile via the app package, but UI did not declare its own test dependency, which caused the required pre-push hook to fail.

## Scope / safety

- No server, Desktop main-process, session, or asset-route behavior changes.
- No changes to the actual `uploadPromptAttachment()` result shape.
- No migration or SDK generation required.
- Existing image attachment thumbnail generation remains the same; this only pre-warms the primitives before the first real paste.
- Work was done in an isolated worktree to avoid touching unrelated local changes from other agents.

## Verification

Ran in the isolated worktree:

- `bun test src/utils/prompt-attachment.test.ts`
  - 3 tests passed, including the new warmup coverage.
- `bun run typecheck` in `packages/app`
- `bunx prettier --check packages/app/src/entry.tsx packages/app/src/utils/prompt-attachment.ts packages/app/src/utils/prompt-attachment.test.ts`
- `git diff --check`
- Maintability review: passed; no blocking issues.
- Pre-push hook on `git push`:
  - root `bun turbo typecheck`: 10/10 successful
  - formatter check: passed

## Expected user impact

The first pasted image in Desktop should no longer be the first operation to initialize Chromium's image decode, Canvas 2D, and WebP encode paths. The initial cold work is shifted to an idle, best-effort startup task, so the first real paste should feel much closer to subsequent pastes.

## Follow-up ideas

If profiling still shows paste-time latency for very large screenshots, the next improvements would be to move thumbnail generation off the critical path or into a worker. This PR intentionally keeps the change minimal and targeted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,7 +416,9 @@ Control profiles are configured in the permissions domain (`80-permissions.jsonc
 }
 ```
 
-**Precedence:** agent config `controlProfile` > top-level config `controlProfile` > default `guarded`.
+**Precedence:** explicit session control profile resolved from the session parent chain > agent config `controlProfile` > top-level config `controlProfile` > source default. Channel-backed and agenda automation root sessions default to `autonomous`; ordinary interactive/manual sessions default to `guarded`.
+
+Explicit configuration is always honored. For example, a top-level `controlProfile: "full_access"` remains `full_access` for Feishu/channel sessions instead of being downgraded by unattended metadata.
 
 Blueprint runs started from the Notes side panel use the current session's control profile when running in the current session. New-session and worktree Blueprint runs create an execution session with at least `autonomous`; a top-level `full_access` profile remains `full_access`.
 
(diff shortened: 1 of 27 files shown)


yzxoi:
Reviewed and tested — this is a clean, well-scoped fix. 3/3 tests pass, typecheck green.

The warmup approach is well-designed: best-effort, 1×1 canvas, idle-scheduled, zero side effects. Small note: the branch is currently behind `dev` — rebase before merging.

Thanks for the thorough PR description and the thoughtful approach to isolating this cold-start cost.

yzxoi:
Pushed a fix for the CI test failure (dce5861c).

**Root cause**: `repairPendingReplyFlags` scans globally across all scopes. In CI, a stale/partial session record from another test lacked a `time` field, causing `info.time.archived` to throw at runtime.

**Fix**: one-character change — `info.time.archived` → `info.time?.archived` (optional chaining guard). This matches the defensive style already used in `migrateSessionWorktreeWorkspace` elsewhere in the same file.

Locally verified 20+ runs of the specific test pass consistently.

yzxoi:
Small note on why this fix is in the same PR rather than separate: the failure was blocking CI merge for this PR. It's a one-character defensive guard in a migration function that scans globally across scopes — same pattern already used elsewhere in the file. The fix is trivial, tested, and unrelated to the image warmup change. Keeping it here unblocks the merge.

PR #196, opened 2026-07-03
Use native desktop project picker

EricSanchezok:
## Summary
- Routes Add/Open Project through one centralized project directory picker so desktop managed/local mode uses the OS-native folder picker, while Web, remote, and desktop external-server modes continue browsing the server filesystem.
- Adds a strict Electron `dialog:select-directory` IPC contract with main-process authorization on both renderer identity and managed/running server state.
- Removes picker-level Git initialization UI/data flow and documents the final native-local vs server-filesystem behavior.

## Problem
Add/Open Project selection was owned by separate call sites and always used the server-directory browser UI. That was correct for Web and remote/external-server contexts, but suboptimal for the production desktop app when it is connected to its managed local Synergy server: the user is selecting a local directory on the same machine, so the native OS folder picker is the expected desktop UX.

The old picker also mixed directory selection with Git initialization. `layout.tsx` read `{ directory, initGit }` and optionally called `global.git.init`, while the sidebar passed `showInitGit={true}` but ignored `initGit`. That made Git initialization inconsistent and left picker-level options that were not really part of directory selection.

## Why
The desktop managed-local app has a different filesystem authority from Web/remote modes:

- In managed desktop mode, the Electron shell and the managed Synergy server share the local machine filesystem, so native OS folder selection is safe and expected.
- In Web, remote server, and desktop external-server modes, the server owns the filesystem namespace, so project selection must remain a server-directory browser.

This PR makes that boundary explicit. It also removes the Git-init checkbox and result field entirely instead of keeping no-op compatibility props or fallback branches.

### Desktop IPC and preload
- Added `selectDirectoryDialogRequestSchema` and `selectDirectoryDialogResponseSchema` in `packages/desktop/src/ipc-contract.ts`.
  - Request schema is strict, allows only `title` and `multiple`, defaults `multiple` to false, and rejects empty/overlong titles.
  - Response schema is strict and rejects inconsistent canceled/selected path states.
- Added `packages/desktop/src/directory-picker.ts` for testable native picker authorization and response mapping.
- Registered `dialog:select-directory` in `packages/desktop/src/main.ts`.
  - Rejects unless `event.sender === mainWindow.webContents`.
  - Rejects unless `serverManager.status()` is `mode: "managed"` and `state: "running"`.
  - Rejects if Electron returns multiple paths for a single-selection request.
- Exposed `window.synergyDesktop.openDirectoryPickerDialog()` from preload.
  - Cancel maps to `null`.
  - Single mode maps to `string`.
  - Multiple mode maps to `string[]`.
  - Rejections propagate with no fallback.

### App routing and cleanup
- Added `DesktopServerStatus` / `DesktopServerBridge` to the app `Platform` type and wired the preload `server` bridge in `entry.tsx`.
- Added a centralized project picker helper:
  - `packages/app/src/components/dialog/project-directory-picker-model.ts`
  - `packages/app/src/components/dialog/project-directory-picker.tsx`
- The helper queries desktop server status at selection time and uses native picking only when:
  - `platform.platform === "desktop"`
  - `openDirectoryPickerDialog` exists
  - desktop server status is `managed` + `running`
- Native failure shows one explicit error toast and stops; it does not fall back to the server-directory browser.
- Duplicate picker invocations are serialized by a pending guard.
- `layout.tsx` and `sidebar.tsx` now both use the centralized helper and simply open the selected directories.
- `DialogSelectDirectory` remains the server-directory browser, but no longer exposes or renders `showInitGit` / `initGit`.
- Removed the Open Project `global.git.init` call from picker selection.

### Docs and dependency housekeeping
- Updated `README.md` desktop behavior documentation.
- Updated `packages/app/PRODUCT.md` with the durable UX/security principle for local-native vs server-filesystem project selection.
- Added `@happy-dom/global-registrator` as a UI package dev dependency because the UI package already typechecks a Happy DOM test that imports it directly.

## Verification
- `bun test test/ipc-contract.test.ts` from `packages/desktop`
- `bun test src/components/dialog/project-directory-picker.test.ts` from `packages/app`
- `bun run typecheck`
- `./script/format.ts`
- Push hook also reran typecheck and formatting successfully.

Note: running `bun test packages/desktop/test/ipc-contract.test.ts` from the repository root hits the root test guard (`do-not-run-tests-from-root`), so the desktop test was run from `packages/desktop`, matching repository instructions.

## Reviews
- Security reviewer: no blockers. Confirmed main-process sender/server-mode authorization, strict IPC validation, no selected-path logging, and no native-failure fallback.
- API compatibility reviewer: no blockers. No server routes, SDK, persistence, or migration impact.
- Maintainability reviewer: initial stale compatibility branches were removed; final structure keeps picker ownership centralized without old Git-init clutter.


yzxoi:
Dev managed mode needs `bun dev build app` first — the Synergy server serves `app/dist/index.html` directly in managed mode (no Vite proxy). In a fresh checkout, `app/dist` doesn't exist, so the desktop window loads a blank/error page while the API server is fine. This isn't a regression from this PR, but worth a follow-up issue.

yzxoi:
One more note for testing: the native folder picker only activates in managed server mode (`mode: "managed"` + `state: "running"`). External/server-browser modes always use the directory browser — that's by design since the Electron shell and server don't share filesystem authority in those cases.

This means the native picker can only be tested via `bun dev desktop --managed`. The `bun dev desktop --managed` startup issue from above (fresh checkout needs `bun dev build app` first) is relevant because there's no other dev path that exercises native picker routing — unlike external mode which works with Vites dev server.

PR #197, opened 2026-07-03
Resolve channel control profiles from session source

EricSanchezok:
## Summary

This PR removes the legacy unattended permission auto-approval path and makes channel/automation permissions resolve through the same control-profile authority used by ordinary tool enforcement.

The intended final behavior is:

- Explicit session control profiles win through the parent chain.
- Agent `controlProfile` wins next.
- Top-level config `controlProfile` wins next.
- If nothing explicit is configured, the session source decides the default:
  - channel-backed root sessions default to `autonomous`;
  - agenda automation root sessions default to `autonomous`;
  - ordinary interactive/manual root sessions default to `guarded`.
- `SessionInteraction.unattended(...)` remains useful metadata for source labels, diagnostics, stats, and question rejection, but it no longer silently changes permission decisions.
- Explicit `full_access` is respected in channel/unattended contexts instead of being rejected just because the session is non-interactive.

## 1. 问题是什么

Before this cleanup, channel and automation permissions were controlled by two different models at the same time.

### Legacy unattended metadata silently changed permission asks

`PermissionNext.ask()` had a special branch that looked at request metadata:

- if `metadata.sessionInteractionMode === "unattended"`, and
- the request was not marked non-bypassable,
- then an `ask` decision was silently resolved as approved.

That meant `unattended` was not just session/source metadata. It was also an implicit permission override. A guarded channel session could therefore behave unlike an ordinary guarded session even when the resolved control profile said the action should ask.

This was especially confusing for Feishu/channel sessions because the channel entry point creates sessions with `SessionInteraction.unattended("channel:feishu")`. The intent was “this surface cannot answer live questions”, but the implementation also treated that metadata as “approve ordinary asks”.

### Effective control-profile resolution was duplicated

The same precedence logic was scattered across runtime paths:

- session resolution;
- tool resolver profile promises;
- local tool gate creation;
- MCP tool gate creation;
- plugin permission checks;
- external-agent setup;
- permission context prompt generation;
- control-profile route helpers.

Those copies generally ended at `guarded`, so prompt text, tool enforcement, and plugin decisions could drift as the rules changed. A future fix in one path could easily leave another path enforcing a different profile.

### Channel and agenda sessions had the wrong default shape

Without explicit configuration, Feishu/channel and agenda automation sessions need to be non-interactive by default. The right default is not hidden ask auto-approval; it is an explicit `autonomous` source fallback.

The old design achieved some automation convenience by inspecting `unattended` metadata at permission-request time. That made the default behavior hard to reason about and tightly coupled to a specific metadata field instead of a profile.

### `full_access` was artificially blocked for unattended contexts

The old control-profile compiler rejected `full_access` when the resolution context said the interaction mode was unattended. That contradicted explicit user configuration. If a user deliberately chooses top-level or session/agent `full_access`, the resolver should honor that configuration regardless of whether the session is backed by Feishu, agenda, or an interactive UI.

### Control profiles should be the single source of permission truth

The permission model is much easier to audit when one layer owns each decision:

- Session/source context chooses an effective profile only when no explicit profile exists.
- Control profiles define allow / ask / deny rules.
- The enforcement gate and permission system apply those rules.
- Session interaction metadata stays metadata.

This PR removes the hidden rule where `unattended` could override `ask`. If a resolved profile says a request should ask, it now goes through the normal permission request path. If a channel or agenda session should avoid routine asks, that should happen because it resolved to `autonomous`, not because a metadata branch silently approved it.

### Explicit configuration must remain explicit

A user-selected profile is a stronger signal than source defaults. That matters for both directions:

- If a channel session has no explicit profile, it gets `autonomous` because the source is non-interactive.
- If the user explicitly configures `guarded`, asks can surface normally.
- If the user explicitly configures `full_access`, Synergy should not downgrade or reject it just because the source is unattended.

This makes behavior predictable: source defaults only fill the blank; they do not override user configuration.

### Feishu should not need a provider-specific permission knob

Feishu is currently the channel provider, but the behavior belongs to the generic channel endpoint model. Adding a Feishu-specific `controlProfile` field would create another policy path and would not help future channel providers.

By using `endpoint.kind === "channel"` at the session/source layer, future channel providers inherit the same default without adding special-case config.

### Deleting the old path is cleaner than wrapping it

The old unattended auto-approval path was not a boundary that needed compatibility. It was the behavior being replaced. Keeping it behind another condition or adapter would preserve the same confusing model with more layers.

This PR intentionally deletes the obsolete branch, deletes request metadata plumbing that existed to feed it, and removes `interactionMode` from control-profile resolution rather than passing it around unused.

### Centralized effective profile resolution

Added a single owner-level resolver in `packages/synergy/src/session/index.ts`:

- `Session.resolveEffectiveControlProfile(...)`
- `Session.defaultControlProfileForSessionSource(...)`

The resolver now applies the required precedence:

1. Explicit session `controlProfile`, walking the session parent chain.
2. Agent `controlProfile`.
3. Top-level config `controlProfile`.
4. Source fallback:
   - channel roots -> `autonomous`;
   - agenda roots -> `autonomous`;
   - ordinary/manual roots -> `guarded`.

It uses `ControlProfileCompiler.normalize(...)` to validate explicit string values, while keeping source fallback logic in the session/source boundary.

### Replaced duplicated runtime resolution

Updated the runtime paths that previously rebuilt their own fallback chains:

- `packages/synergy/src/session/invoke.ts`
  - external-agent setup now uses the central resolver;
  - permission context prompt generation now uses the central resolver.
- `packages/synergy/src/session/tool-resolver.ts`
  - deleted the local `resolveEffectiveProfile(...)` helper;
  - local tools, MCP tools, and profile promise setup all use the central resolver.
- `packages/synergy/src/plugin/host-services.ts`
  - plugin permission checks use the central resolver before profile policy decisions.
- `packages/synergy/src/server/control-profile-route.ts`
  - removed obsolete interaction-mode context from profile resolution.

The important invariant is that the prompt context and all enforcement paths now ask the same resolver for the same effective profile.

### Deleted unattended auto-approval

In `packages/synergy/src/permission/next.ts`:

- removed the `unattended auto-approve` branch;
- removed `PermissionNext.requestMetadata(...)`;
- removed call-site spreads of session interaction metadata from permission requests.

`SessionInteraction.unattended(...)` still exists and still matters for non-permission behavior such as question rejection and session diagnostics, but it no longer auto-approves normal permission asks.

### Removed `full_access` unattended prohibition

In the control-profile/enforcement boundary:

- removed the `full_access` invalid-profile branch for unattended contexts;
- removed `interactionMode` from `ResolutionContext`;
- removed `interactionMode` from `EnforcementGate` options;
- removed `interactionMode` call-site plumbing.

`full_access` is now resolved like any other explicit profile. It is never an implicit fallback for channels or agenda sessions, but it is respected when explicitly configured.

### Updated tests around the new invariants

Updated and added tests for:

- ordinary root sessions defaulting to `guarded`;
- channel roots defaulting to `autonomous`;
- children of channel roots inheriting the channel root source fallback;
- channel roots honoring top-level `full_access`;
- channel roots honoring explicit session `guarded`;
- agenda roots defaulting to `autonomous`;
- unattended metadata no longer auto-approving `ask` rules;
- explicit deny rules still rejecting;
- `full_access` being valid without interaction-mode restrictions;
- enforcement gate behavior without `interactionMode`;
- MCP/plugin guarded asks without relying on unattended mode;
- stale resolver test coverage now calling the real `Session.resolveEffectiveControlProfile` instead of reimplementing deleted fallback logic.

### Updated documentation

Updated the README control-profile section to describe the final behavior directly:

- full precedence order;
- channel/agenda source defaults;
- ordinary interactive default;
- explicit `full_access` remaining valid for channel sessions;
- removed stale wording that said `full_access` is blocked in unattended execution.

### Kept quality gates reproducible

During repository-level typecheck, `packages/ui` failed because `packages/ui/src/components/special-user-message.test.tsx` imports `@happy-dom/global-registrator` but `packages/ui/package.json` did not declare it. The dependency already existed in the lockfile through the app package, but package-local typecheck needs package-local declaration.

This PR adds `@happy-dom/global-registrator` to `packages/ui` devDependencies so `bun run typecheck` is reproducible from this worktree.

## Verification

Commands run:

```bash
cd packages/synergy
bun test test/session/session.test.ts test/session/unattended.test.ts test/permission/next.test.ts test/control-profile/compiler.test.ts test/control-profile/resolve-profile.test.ts
bun test test/permission/identity-boundary.test.ts test/permission/agent-blanket-allow.test.ts
bun test test/session test/permission test/control-profile
bun test test/session/prompt-budgeter.test.ts test/session/invoke-compaction.test.ts
bun test test/permission/identity-boundary.test.ts test/permission/agent-blanket-allow.test.ts test/enforcement/gate.test.ts test/enforcement/mcp-plugin.test.ts
bun run typecheck
./script/format.ts
bun run typecheck && ./script/format.ts
```

Results:

- Targeted session/permission/control-profile tests passed.
- Boundary tests passed.
- Final changed-file test set passed: 332 tests.
- `bun run typecheck` passed.
- `./script/format.ts` passed.
- Pre-push hook reran typecheck and formatting successfully before pushing the branch.

Note: the broad `bun test test/session test/permission test/control-profile` slice produced 615 passing tests and 4 transient failures in session prompt-budgeter/compaction tests during the combined run. The failed files were immediately rerun directly and passed, so no code change was made for that transient combined-run timing issue.

## Review

Maintainability review was run before PR creation.

- Initial review found one blocker: `resolve-profile.test.ts` still had an inline mock of deleted resolver logic.
- The test was replaced with real `Session.resolveEffectiveControlProfile` coverage.
- Re-review passed with no blockers.

## SDK / migration impact

- No config schema change.
- No persisted schema migration.
- No SDK regeneration required; server route schemas/OpenAPI contracts were not changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,7 +416,9 @@ Control profiles are configured in the permissions domain (`80-permissions.jsonc
 }
 ```
 
-**Precedence:** agent config `controlProfile` > top-level config `controlProfile` > default `guarded`.
+**Precedence:** explicit session control profile resolved from the session parent chain > agent config `controlProfile` > top-level config `controlProfile` > source default. Channel-backed and agenda automation root sessions default to `autonomous`; ordinary interactive/manual sessions default to `guarded`.
+
+Explicit configuration is always honored. For example, a top-level `controlProfile: "full_access"` remains `full_access` for Feishu/channel sessions instead of being downgraded by unattended metadata.
 
 Blueprint runs started from the Notes side panel use the current session's control profile when running in the current session. New-session and worktree Blueprint runs create an execution session with at least `autonomous`; a top-level `full_access` profile remains `full_access`.
 
@@ -426,13 +428,11 @@ Risk levels follow the operation's effect. Ordinary reads, including non-protect
 
 Built-in profiles:
 
-| Config value  | UI label    | Behavior                                                                                                                                                                               |
-| ------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `guarded`     | Guarded     | Default protected mode. Auto-allows ordinary reads, workspace-local edits, and ordinary network lookups; asks before shell, external writes, identity, platform, or extension actions. |
-| `autonomous`  | Autonomous  | Unattended mode. Includes Guarded's automatic approvals, allows medium-risk development work, and denies high-risk asks instead of prompting.                                          |
-| `full_access` | Full Access | Allows all tool requests without approval prompts or workspace sandboxing.                                                                                                             |
-
-`full_access` is blocked in unattended execution mode. It is only available in attended sessions.
+| Config value  | UI label    | Behavior                                                                                                                                                                                           |
+| ------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `guarded`     | Guarded     | Protected mode for manual sessions. Auto-allows ordinary reads, workspace-local edits, and ordinary network lookups; asks before shell, external writes, identity, platform, or extension actions. |
+| `autonomous`  | Autonomous  | Automation-oriented profile. Includes Guarded's automatic approvals, allows medium-risk development work, and denies high-risk asks instead of prompting.                                          |
+| `full_access` | Full Access | Allows all tool requests without approval prompts or workspace sandboxing. Explicit `full_access` remains valid for channel and automation sessions.                                               |
 
 ### Sandbox
 
--- a/bun.lock
+++ b/bun.lock
@@ -338,6 +338,7 @@
         "virtua": "catalog:",
       },
       "devDependencies": {
+        "@happy-dom/global-registrator": "20.0.11",
         "@tailwindcss/vite": "catalog:",
         "@tsconfig/node22": "catalog:",
         "@types/bun": "catalog:",
--- a/packages/synergy/src/control-profile/profiles.ts
+++ b/packages/synergy/src/control-profile/profiles.ts
@@ -225,7 +225,7 @@ export async function resolveEffectiveSandbox(profileId: ProfileId): Promise<Pro
 
 export async function buildProfile(idInput: ProfileIdInput | string, ctx: ResolutionContext): Promise<ResolvedProfile> {
   const id = normalizeProfileId(idInput)
-  const { workspace, interactionMode } = ctx
+  const { workspace } = ctx
   const effectiveSandbox = await resolveEffectiveSandbox(id)
 
   switch (id) {
@@ -263,19 +263,6 @@ export async function buildProfile(idInput: ProfileIdInput | string, ctx: Resolu
     }
 
     case "full_access": {
-      if (interactionMode === "unattended") {
-        return {
-          valid: false,
-          reason: "full_access profile is forbidden in unattended mode",
-          label: "Full Access",
-          description: "Unrestricted local access. Disabled for unattended sessions.",
-          ruleset: [],
-          filesystem: { readRoots: [], writeRoots: [], protectedPaths: [] },
-          network: { mode: "disabled" },
-          sandbox: { mode: "read_only", fallback: "deny" },
-          approval: approval("full_access"),
-        }
-      }
       const policy = fullAccessPolicy()
       const profile = {
         valid: true,
--- a/packages/synergy/src/control-profile/types.ts
+++ b/packages/synergy/src/control-profile/types.ts
@@ -46,7 +46,6 @@ export interface ControlProfile {
 export interface ResolutionContext {
   workspace: string
   workspaceType: string
-  interactionMode?: string
 }
 
 export interface ProfileSummary {
--- a/packages/synergy/src/enforcement/gate.ts
+++ b/packages/synergy/src/enforcement/gate.ts
@@ -94,7 +94,6 @@ export interface GateOptions {
   activeWorkspace: string
   workspaceType: string
   profileId?: ProfileIdInput
-  interactionMode?: "attended" | "unattended"
   registeredMcpTools?: Set<string>
   registeredPluginTools?: Set<string>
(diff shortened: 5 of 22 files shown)


yzxoi (approved):
Thanks @EricSanchezok — this is a clean, well-designed PR.

## Review Summary

✅ **No blockers.** Recommend merge.

### Architecture
The dual-path permission model (control-profile rules vs. hidden `unattended` auto-approval) was a real source of confusion. Centralizing resolution into `Session.resolveEffectiveControlProfile` and deleting the `unattended` auto-approve branch is the right call — one owner for one decision.

### What I checked
- **Security**: Removal of unattended auto-approval closes a legitimate bypass. `full_access` honoring explicit config is correct. Channel/agenda `autonomous` default is appropriate.
- **Code quality**: Clean abstraction. All dead code (`requestMetadata`, `topLevelControlProfile`, `interactionMode`, duplicated resolvers) fully removed with zero residual references in `src/`.
- **API compatibility**: Internal changes only. No SDK regeneration, no config migration, no public contract breakage.
- **Docs**: README accurately reflects the new precedence and terminology.
- **Verification**: 330 targeted tests pass. Synergy package typecheck clean.

### Minor notes (non-blocking)
- `@happy-dom/global-registrator` in `packages/ui/package.json` is unrelated to the control-profile work — a small scope creep, but zero impact since it was already in the lockfile.
- The one behavioral change worth calling out: users with top-level `full_access` + a channel bot now genuinely get `full_access` for the bot. Previously the old model silently downgraded it. This is correct behavior, but worth noting in release notes since anyone relying on the old hidden restriction should explicitly configure `autonomous` or session-level `guarded` instead.

yzxoi:
补充：已在飞书 channel 实际环境中测验过，channel session 默认  `autonomous ` 权限工作正常，高风险操作正确拒绝，无异常。

PR #199, opened 2026-07-03
fix: terminal WebSocket connection in home scope

### LINKED ISSUE #170: bug: terminal never connects in home scope — 'Reconnecting...' hangs forever
## User report

> @chaos terminal 在 home 下展开Reconnecting..卡死，在 project 下正常

Reporter: Feishu:ou_114e5f3416669d8bfaae75fcaa661915
Source: group:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6b50fb3e44bcc37013fbef11f28

## Triage

- Type: bug_report
- Related area: app / terminal

## Deduplication

No duplicate found in recent open/closed issues and PRs.

### Root cause

`packages/app/src/components/terminal.tsx`, line 231:

```typescript
const connect = () => {
  if (!sdk.directory) return  // <-- BUG: in home scope, sdk.directory is undefined
```

When the terminal component mounts in home scope, `sdk.directory` is `undefined` (see `packages/app/src/context/sdk.tsx` line 34: `directory: isHome ? undefined : props.scopeKey`). The `connect()` function checks `if (!sdk.directory) return` and exits early **without ever establishing the WebSocket connection**.

Since the WebSocket connection is never opened, `connected` signal stays `false`, and the terminal permanently shows the "Reconnecting..." overlay. The reconnect loop is never triggered because the initial connection attempt is silently skipped.

In project scope, `sdk.directory` has a valid path, so the WebSocket connection is established normally and the terminal works.

### Additional details

- `packages/app/src/context/sdk.tsx` lines 33-34: in home scope, `scopeID` is set to `HOME_SCOPE_KEY` but `directory` is `undefined`.
- The server-side PTY WebSocket route (`/pty/:ptyID/connect`) doesn't actually use the `directory` query parameter — but the `provideRequestScope` middleware uses it for scope resolution.
- When `scopeID` is available (home scope), it could be passed instead of (or in addition to) `directory`.

## Expected behavior

Terminal WebSocket connection should be established correctly in home scope, just like in project scope. The terminal should not show "Reconnecting..." indefinitely.

## Actual behavior

In home scope, `if (!sdk.directory) return` prevents the WebSocket connection from ever being established, causing the terminal to show "Reconnecting..." forever.

## Proposed fix

In `packages/app/src/components/terminal.tsx`, modify the WebSocket URL construction to fall back to `sdk.scopeID` when `sdk.directory` is undefined:

```typescript
const connect = () => {
  const params = new URLSearchParams()
  if (sdk.directory) {
    params.set("directory", sdk.directory)
  } else if (sdk.scopeID) {
    params.set("scopeID", sdk.scopeID)
  }
  const socket = new WebSocket(
    sdk.url + `/pty/${local.pty.id}/connect?${params.toString()}`,
  )
  // ... rest of the function
}
```

The server-side `provideRequestScope` middleware already supports both `directory` and `scopeID` parameters for scope resolution, so this fix is safe.

yzxoi:
## Summary

Fixes #170 — terminal was showing "Reconnecting..." forever in home scope because `sdk.directory` is `undefined` in home scope, causing `connect()` to return early without ever establishing the WebSocket connection.

## Root Cause

In `packages/app/src/components/terminal.tsx`, the `connect()` function checked `if (!sdk.directory) return`. In home scope, `sdk.directory` is `undefined` (see `packages/app/src/context/sdk.tsx:34`), so the WebSocket was never opened. The terminal permanently displayed "Reconnecting..." because `connected` stayed `false` and the reconnect loop was never triggered.

## Fix

Instead of returning early when `sdk.directory` is missing, build `URLSearchParams` with `sdk.scopeID` as fallback:

- When `sdk.directory` is available (project scope) → pass `?directory=...`
- When `sdk.scopeID` is available (home scope) → pass `?scopeID=home`

The server-side `provideRequestScope` middleware already supports `scopeID` query parameter (via `requestScopeID()` in `server.ts:220`), and `Scope.fromID("home")` resolves to the home scope with `directory` set to the user's home directory. No server-side changes needed.

## Verification

- [x] Server-side `provideRequestScope` already handles `scopeID` query parameter
- [x] `Scope.fromID("home")` resolves to home scope with correct directory (`~`)
- [x] PTY shell cwd falls back to `ScopeContext.current.directory`


yzxoi:
Tested on this branch — terminal now connects normally in home scope and opens at the user home directory. Thanks for the thorough issue analysis!

PR #220, opened 2026-07-03
Show provider usage reset times

EricSanchezok:
## Summary
- show the next provider usage reset in the Usage overview
- display per-window reset timing inline with quota rows
- move usage formatting into a tested model helper and update the Usage product contract

## Tests
- bun test src/components/settings/panels/UsagePanel.test.ts (from packages/app)
- bun run typecheck
- git diff --check
- pre-push typecheck and Prettier during git push


yzxoi (approved):
The UI logic is clean, the model extraction with tests is a nice quality upgrade, and the Anthropic OAuth path will benefit directly.

Thanks @EricSanchezok! 🚀

PR #226, opened 2026-07-03
Fix running session input queue ordering

EricSanchezok:
## Summary

Fixes the running-session input queue semantics so queued user input and guided user input obey two explicit invariants:

- A normal `queued_user` / `after_turn` input submitted while a session is running stays in the inbox until the current turn completes. It is only materialized into the transcript when the next turn is started.
- The final transcript `messageID` for queued or guided input is generated at materialization time, not at enqueue time and not from a frontend optimistic id.

This also removes the fragile message-id-order based reply detection and replaces it with assistant `parentID` ownership checks.

## Problem

The stuck session showed two related failures in the running input path:

1. A message sent while the agent was already running could enter the active session turn even though the user did not manually guide it with the lightning action. Under the intended model, that input should wait in the inbox until the current turn finishes, then trigger the next turn.
2. Queued/guided inputs could carry a preselected `messageID` from enqueue time or frontend optimistic state. That made transcript order depend on an id allocated before the message actually existed in the transcript.

There was also a third contributing issue: parts of the loop used message id lexical ordering to decide whether the latest user message had already received an assistant reply. That is not a valid ownership signal once messages can be materialized later or have unrelated ids. The durable relationship is `assistant.parentID === user.id`.

## Solution

Backend input queue:

- `/session/:sessionID/input` now checks whether the session is running before allocating or preserving a final message id.
- If the session is running, the route enqueues the raw user input and explicitly drops any request `messageID`.
- If the session is idle, the existing optimistic-id behavior is preserved so immediate-start submissions can still reconcile with the frontend optimistic message.
- `SessionInbox.enqueueUser()` no longer generates or stores a final transcript `messageID`; the inbox item's stable identity is only `item.id`.
- `materializeInboxItems()` is the single place where queued/guided inbox items become transcript messages, and it strips stale/client-supplied ids before calling `createUserMessage()`.

Reply detection:

- Added parent-link helpers in `SessionProgress`:
  - find terminal assistant replies whose `parentID` points to the user message
  - check whether a user message has a terminal assistant reply
- Replaced the old `lastUser.id < lastFinished.id` loop condition with the parent-link helper.
- Updated `pendingReply()` to compute pending state from the latest reply-required user and terminal assistant parent links.
- Kept `selectResultMessage()` aligned with the same parentID semantics.

Frontend submit behavior:

- Normal submit and quick action submit now detect `working()` before creating a final message id.
- While working, the frontend sends only prompt parts and does not create/send a `messageID`.
- Idle/new-session submit still creates an optimistic id so existing immediate-start UI reconciliation is unchanged.
- Inbox UI continues to rely on `item.id`, not `item.messageID`.

Data repair:

- Added a one-time session migration that scans non-archived sessions and recomputes `pendingReply` with the new parentID-based logic.
- This identifies old swallowed queued user messages as still pending when no assistant terminal reply points to them.
- The migration does not rewrite historical message ids, does not auto-call a provider, and does not add runtime fallback behavior.

## Why This Shape

This keeps one clean model instead of preserving conflicting legacy paths:

- Inbox identity and transcript identity are separate. `item.id` identifies queued work; transcript `messageID` is created only when the message is actually inserted into transcript order.
- Queue timing is explicit. Normal running input waits for the current turn to finish; guided input is the separate path that can steer the active turn.
- Reply ownership is explicit. Parent links describe which assistant message answered which user message; id ordering only describes allocation timing and is not reliable for conversation semantics.
- Data repair is isolated in migration. Existing bad session state is fixed once without adding permanent compatibility branches to runtime logic.

## Tests

Added and updated coverage for:

- queued inbox items not writing transcript messages early
- queued/guided inbox items not exposing a final `messageID`
- normal running input waiting until after the current processor call before materialization
- queued materialization creating a new user id after the previous assistant reply
- guided input entering the current model call with `noReply/guided` metadata and without triggering an extra turn
- legacy bad data where user id ordering would incorrectly imply a reply exists
- migration recomputing `pendingReply` for bad, completed, and archived sessions

Verification run:

```bash
cd packages/synergy && bun test test/session/inbox.test.ts test/session/invoke.test.ts test/session/migration.test.ts
bun run typecheck
git diff --check
```

The push hook also ran typecheck and Prettier verification successfully. Turbo emitted a lockfile transitive-closure warning for an optional dependency entry, but all typecheck tasks completed successfully.

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -870,12 +870,16 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
     const agent = currentAgent.name
     const model = { modelID: currentModel.id, providerID: currentModel.provider.id }
     const variant = local.model.variant.current()
-    const messageID = Identifier.ascending("message")
+    const queueing = working()
+    const messageID = queueing ? undefined : Identifier.ascending("message")
     const textPart = { id: Identifier.ascending("part"), type: "text" as const, text }
 
-    const optimistic: Message = { id: messageID, sessionID, role: "user", time: { created: Date.now() }, agent, model }
+    const optimistic: Message | undefined = messageID
+      ? { id: messageID, sessionID, role: "user", time: { created: Date.now() }, agent, model }
+      : undefined
     let optimisticAdded = false
     const addOptimisticMessage = () => {
+      if (!messageID || !optimistic) return
       sync.set(
         produce((draft) => {
           const messages = draft.message[sessionID]
@@ -891,6 +895,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
       optimisticAdded = true
     }
     const removeOptimisticMessage = () => {
+      if (!messageID) return
       sync.set(
         produce((draft) => {
           const messages = draft.message[sessionID]
@@ -904,10 +909,10 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
       optimisticAdded = false
     }
 
-    if (!working()) addOptimisticMessage()
+    if (!queueing) addOptimisticMessage()
 
     sdk.client.session
-      .input({ sessionID, agent, model, messageID, parts: [textPart], variant })
+      .input({ sessionID, agent, model, ...(messageID ? { messageID } : {}), parts: [textPart], variant })
       .then((result) => {
         if (result.data?.status === "queued" && optimisticAdded) removeOptimisticMessage()
       })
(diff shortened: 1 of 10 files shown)


yzxoi:
Thanks for the PR @EricSanchezok! The analysis is spot-on — glad this was caught early.

I opened the session you mentioned and confirmed the interleaved message issue. The fix is clean and well-structured, especially the separation of inbox identity vs transcript identity. 🧹

Left a small comment in the review above, and pushed a quick test addition to make the materialized-messageID scenario more explicit. Tests all pass ✅

Nice work! 🚀

yzxoi (approved):

PR #227, opened 2026-07-03
Allow subagents to use MCP tools

EricSanchezok:
## Summary
- Allow built-in subagents to see dynamic MCP tools by default via the shared `mcp__*` permission rule.
- Cover both classic and synergy-max subagent paths so future MCP servers do not need per-tool permission wiring.

## Why
MCP tools are registered dynamically as `mcp__<server>__<tool>`. Built-in subagents start from `* deny`, so newly configured MCP servers were available to the primary agent but hidden from subagents unless each server/tool was explicitly allowlisted. This makes MCP configuration brittle as more MCP servers are added.

The new shared rule keeps unknown non-MCP tools denied while allowing MCP tool IDs through the existing permission mechanism. More specific user or project permission overrides can still deny individual MCP servers or tools later.

## Tests
- `cd packages/synergy && bun test test/agent/max-subagents.test.ts test/agent/agent.test.ts`
- pre-push hook: `bun turbo typecheck`
- pre-push hook: formatting check

### DIFF
--- a/packages/synergy/src/agent/builtin-context.ts
+++ b/packages/synergy/src/agent/builtin-context.ts
@@ -89,6 +89,7 @@ function anchoredWriteTools(): PermissionNext.Ruleset {
 function baseToolPermissions(profile: SubagentPermissionProfile): PermissionNext.Ruleset {
   const common = PermissionNext.fromConfig({
     "*": "deny",
+    "mcp__*": "allow",
     question: "deny",
     dagwrite: "deny",
     dagread: "deny",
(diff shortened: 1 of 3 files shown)


yzxoi:
Nice one @EricSanchezok! 🔥

Tested this live — spawned a `developer` subagent and had it actually call an arxiv MCP server. Here's what happened:

| Check | Result |
|-------|--------|
| Subagent sees `mcp__arxiv__*` in tool list | ✅ Directly visible, no extra discovery needed |
| `mcp__arxiv__arxiv_search` call | ✅ Success — returned matching papers |
| `mcp__arxiv__arxiv_download` call | ✅ Success — downloaded PDF (432KB) |
| Any permission rejection | ❌ None — full green path |

The one-liner works exactly as advertised: subagents can now call MCP tools without per-server allowlisting, while `* deny` still catches non-MCP unknowns. Clean fix, thanks for catching this brittle spot! 🙏

yzxoi (approved):

PR #228, opened 2026-07-03
Refine question prompt decision surface

EricSanchezok:
## Summary
- Move the question skip action into the header so single-question prompts no longer show a confusing full-width Dismiss row
- Make the custom answer affordance render as a filled option row with existing workbench tokens
- Tighten question prompt styling around existing radius, surface, semantic icon, and option text patterns

## Testing
- bun test src/workbench-surface.test.ts (from packages/app)
- bun run --cwd packages/app typecheck
- node /Users/eric/.codex/skills/impeccable/scripts/detect.mjs --json packages/app/src/components/session/question-prompt.tsx
- pre-push hook: bun turbo typecheck; Prettier check

### DIFF
--- a/packages/app/src/components/session/question-prompt.css
+++ b/packages/app/src/components/session/question-prompt.css
@@ -17,7 +17,7 @@
   color: var(--text-base);
   background: var(--question-shell-bg);
   border: 1px solid var(--question-border);
-  border-radius: 22px;
+  border-radius: var(--radius-2xl);
   box-shadow: none;
 }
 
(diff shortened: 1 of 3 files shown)


yzxoi:
Nice refinements on the question prompt @EricSanchezok! 🔍

## Reviewer Notes

**Core changes:**
- **Skip moved to header** — the old full-width "Dismiss" footer row is gone; now a compact "Skip" button sits inline in `header-actions`. Big improvement for single-question prompts where a full dismiss row felt awkward.
- **"Other answer" → styled option row** — no longer an isolated text link. Now renders as a proper `.question-prompt-option` with icon + label + description, visually consistent with other options. Clean.
- **Semantic icon tokens** — all bare `Icon name="check"` etc. migrated to `getSemanticIcon("state.success")` etc. 👏
- **Footer scoped to multi-question only** — gated behind `<Show when={!single()}>`.

**CSS tokens:**
- Replaced hardcoded radii with design tokens (`22px` → `var(--radius-2xl)`, `12px` → `var(--radius-xl)`)
- Added smooth hover transitions (150ms ease-out) on option rows
- Option copy now uses flex column for proper label + description stacking
- Removed dead `.question-prompt-dismiss` class

**Test results:** All 14 test cases pass ✅

Good attention to detail — the token alignment and layout tightening are solid. Ship it! 🚀

yzxoi (approved):

PR #230, opened 2026-07-03
Fix managed desktop dev parity

EricSanchezok:
## Summary
- Flatten Markdown component CSS so production builds preserve the root text contract used by message rendering.
- Make `bun dev desktop --managed` always rebuild the Web app before launching managed Electron, only installing dependencies when `node_modules` is missing.
- Add production CSS contract coverage for core component root declarations and update docs for the managed workflow.

## Tests
- bun test test/script/dev.test.ts (packages/synergy)
- bun test test/css-token-integrity.test.ts (packages/ui)
- bun test src/app-build-css-contract.test.ts (packages/app)
- bun run typecheck
- bun dev build app

Note: push hook also reran typecheck and Prettier.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -101,7 +101,7 @@ bun dev server                # server only, fixed development port
 bun dev app --open            # Vite web app against an existing server
 bun dev web                   # server + Vite web app
 bun dev desktop               # server + Vite web app + Electron desktop shell
-bun dev desktop --managed     # Electron desktop shell with managed server mode
+bun dev desktop --managed     # rebuild Web app dist + Electron desktop shell with managed server mode
 ```
 
 One-off CLI execution from source:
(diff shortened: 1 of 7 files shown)


yzxoi (approved):

yzxoi (approved):
Thanks for this fix! Clean work.

The CSS flattening in markdown.css is mechanically correct — every nested rule maps precisely to its explicit `[data-component="markdown"]` prefix, and `@keyframes` extraction to root scope is the right call. The managed desktop logic simplification (always build, install only when missing) is a nice improvement over the old two-branch structure. The production CSS contract test is a smart addition to catch future regressions at this layer.

A couple of minor non-blockers for future follow-up:
- The contract test's CSS parser (`collectRootRuleBodies`) is hand-rolled bracket counting — if Vite changes output format later it could silently under-match. Worth documenting the assumptions or pulling in a lightweight parser.
- The `popover-content` contract declaration for `background-color` is truncated (missing `,var(--surface-raised-stronger-non-alpha))`) — the `includes()` check still passes but it's a false-positive hazard.
- Consider scoping the test subprocess env instead of inheriting full `process.env`.

None of these should block merge. LGTM 🚀

PR #257, opened 2026-07-03
Fix provider waiting elapsed time

EricSanchezok:
## Summary
- Change provider waiting metadata from a wall-clock timestamp to the current turn's elapsed runtime.
- Tick the inline elapsed label once per second while the transient waiting prelude is visible.
- Update the Web product contract to keep the elapsed metadata quiet and non-capsule.

## Verification
- bun test src/components/session-turn-timeline.test.ts (from packages/ui)
- bun run typecheck (from packages/ui)
- git diff --check


yzxoi (approved):
测试全部通过，实现干净利落。纯函数覆盖了负差值、大跨度等边缘情况，组件端的 timer 生命周期也处理得正确（prelude 隐藏时 clearInterval，重新出现时重置）。luxon 移除也没问题，只剩另外两个文件在用。我还顺手补了 ProviderPrelude 的组件级测试和几个纯函数边界断言。

整体 LGTM，感谢贡献！

PR #258, opened 2026-07-03
fix(sidebar): advance nav order at reply boundaries

EricSanchezok:
## Summary
- Advance sidebar nav ordering when a reply starts and again when it completes.
- Keep nav activity timestamps stable for in-flight updates while `pendingReply` remains true.
- Update the nav regression test and Web product contract for the start/end ordering behavior.

## Tests
- `bun test test/session/nav-index.test.ts test/server/nav-global-routes.test.ts test/server/nav-session-index-route.test.ts test/server/nav-scope-index-route.test.ts test/server/session-completion-notice.test.ts`


yzxoi:
Nice fix! 🎯 That one-line condition switch is *chef's kiss* — exactly the right granularity. The three-phase test coverage (start → mid-run freeze → finish) makes the behavior crystal clear. All tests green, typecheck clean, nothing else to wire up. LGTM, ship it! 🚀

yzxoi (approved):

PR #259, opened 2026-07-03
fix(app): sync dark theme before desktop mount

EricSanchezok:
## Summary
- Sync the boot-shell `data-synergy-color-scheme` and app `data-color-scheme` attributes before renderer mount and during runtime theme changes.
- Keep mounted root/loading/session fallback surfaces on theme-backed workbench backgrounds to avoid white flashes in managed desktop dark mode.
- Extend app boot-shell contract coverage for the managed desktop theme transition path.

## Verification
- `bun run typecheck`
- `bun test src/app-boot-shell.test.ts` in `packages/app`
- `bun test src/workbench-surface.test.ts` in `packages/app`
- `bun test src/context/` in `packages/app`

## Notes
- `bun test` in `packages/ui` still has 7 pre-existing unrelated failures around semantic icon grouping, visual token contract drift, and SSR-like special-message tests.


yzxoi (approved):

yzxoi:
Thanks for this fix! Tested in managed desktop dark mode and can confirm the white flash is gone — the loading surfaces now stay properly themed through the entire boot → app mount transition.

The dual-attribute sync in the boot script and the `html:not([data-color-scheme])` → `html[data-color-scheme]` CSS bridge are clean, and `applyBootShellColorScheme` is a nice targeted abstraction that also happens to fill the pre-existing stale-attribute gap at no extra cost. 

LGTM 🚀

PR #260, opened 2026-07-04
feat: add performance observability platform

EricSanchezok:
## Summary
- Add a local-first performance observability subsystem with SQLite WAL storage, retention, resource sampling, spans, metrics, issue coalescing, browser ingestion, SSE updates, and stable `/global/performance/*` APIs.
- Instrument Synergy runtime owners for HTTP, sessions, LLM/provider calls, tools, enforcement, storage, library DB, processes, PTYs, SSE, and observability writer behavior.
- Add the Settings → Runtime → Performance dashboard, browser Web Vitals/resource collection, generated SDK methods, semantic icons, docs, and load-test helper while hiding the standalone Diagnostics settings entry.

## Verification
- `./script/generate.ts && ./script/format.ts && bun run typecheck`
- `bun test test/performance && bun test test/server/performance-route.test.ts && bun test test/observability/observability.test.ts && bun test test/server/openapi.test.ts` from `packages/synergy`
- `bun test test/server` from `packages/synergy`


yzxoi (changes requested):
## Second Review (commits up to b0263c53)

Thanks for the quick turnaround on the first round. The instrumentation `shared.ts` extraction is clean, `sse-metrics.ts` separation is well-done, and `runtime-support.ts` being a testable pure function is a solid pattern.

However, there are **3 blockers** that need to be resolved before merging:

---

### 3 blocking issues

**1. `PerformanceRetention` is dead code — never invoked**

`packages/synergy/src/performance/retention.ts` exports `run()`, `schedule()`, `effectiveLimits()` but no caller exists anywhere in the source tree. Metrics and spans will grow unbounded in SQLite until process restart. Either:
- Wire `PerformanceRetention.schedule()` into `server.ts` startup, with a timer calling `run()` periodically, OR
- Delete the file and defer retention to a future PR.
Exporting a dead module from the public barrel implies active enforcement that doesn't exist.

**2. Metric name drift — dashboard queries names that don't match recorded spans**

`PerformanceDashboard` queries metric names like `llm.request.duration`, `session.llm_call.duration`, `session.tool_call.duration`, `library.sqlite.query.duration` — but `PerformanceSpans.end` records durations as `${ctx.name}.duration`, where the actual span names from instrumentations are `llm.call`, `tool.execution`, `library.query`, `storage.operation`. These dashboard panels will silently return empty arrays while calls are actually happening.

Align the dashboard query names with the recorded span names: `llm.call.duration`, `tool.execution.duration`, `library.query.duration`, `storage.operation.duration`.

**3. `endSpan` export in `shared.ts` is unused**

`instrumentation/shared.ts:39` exports `endSpan`, the barrel re-exports it, but all 10+ instrumentation files import `PerformanceSpans` directly and alias their own `end[X]Span`. Either delete the dead alias or update all instrumentation files to import `endSpan` from `"./shared"` and remove the duplicate direct import. Pick one pattern.

---

### 1 high-severity finding

**`PerformanceIssues` imported but unused in `server.ts`**

If this is a side-effect import, change it to `import "@/performance/issues"`. If genuinely unused, remove it.

---

### Non-blocking follow-ups (medium)

- **`rateBuckets` Map has no expiry cleanup** (`performance-route.ts:48`) — entries accumulate per unique `{bucket}:{clientIP}` forever. Add periodic sweep or bounded TTL map.
- **`PerformanceResources.start()` called before config refresh** (`server.ts:89-92`) — defaults used for first tick (~5s window) regardless of user config. Await the config promise first.

---

### What looks good

- `shared.ts` as instrumentation backbone — `recordMetric` / `startSpan` wrappers are the right abstraction
- `sse-metrics.ts` extraction from `server.ts` — clean separation
- `runtime-support.ts` — testable, focused, well-covered
- Benchmark scripts — dev-only, documented, smoke-tested

**Verdict: Request changes.** The three blockers are simple fixes (wire-or-delete retention, align metric names, drop dead export). Let me know when they're ready for re-review.

yzxoi (approved):

PR #276, opened 2026-07-04
Route quick actions through session commands

EricSanchezok:
## Summary
- add built-in prompt commands for note, continue, audit, and start
- restore runtime quick-action pills by sending commands through the shared session command endpoint
- share session command attachment construction between typed slash commands and quick actions

## Verification
- ./script/format.ts
- git diff --check
- bun run typecheck (fails: environment missing tsgo binary for multiple packages)


yzxoi:
Reviewed this PR — the refactor is clean, the approach is solid, and everything formats and checks out. Nice work! The shared `session-command.ts` extraction is exactly the right direction to keep things DRY across slash commands and quick actions. Thanks @EricSanchezok 🎉

yzxoi (approved):

PR #277, opened 2026-07-04
fix: clarify expand tools output

EricSanchezok:
## Summary
- Replace debug-style expand_tools output with direct callable-tool guidance for agents.
- Add availableRequestedTools metadata for requested tools that are actually visible after expansion.
- Change the UI expand_tools tag from "Next step" to "Ready" and cover the output contract in tests.

## Testing
- ./script/format.ts
- bun test test/tool/exposure.test.ts (from packages/synergy)
- bun run typecheck (from packages/synergy)
- bun run typecheck (from packages/ui)


yzxoi (approved):
Clean refactor — the output format change from raw debug metadata to structured, agent-friendly guidance is a clear improvement. The `availableRequestedTools` addition correctly accounts for permission filtering, and the test coverage on the output contract gives good confidence. Thanks for the contribution, @EricSanchezok! 🎉

PR #284, opened 2026-07-04
fix(app): show new action in collapsed sidebar

EricSanchezok:
## Summary
- Keep the sidebar New action mounted in both expanded and collapsed states, hiding only the label when collapsed.
- Add a `session.new` semantic icon token and document intentional duplicate icon usage in the semantic icon test.

## Verification
- `bun test test/semantic-icon.test.ts` from `packages/ui`
- `git diff --check`

## Notes
- `bun run typecheck` was attempted but the local environment is missing `tsgo`, so it failed before code type checks could run.
- Browser visual verification was attempted against localhost, but Playwright Chromium launch timed out in this environment.


yzxoi (commented):
Thanks for this! Confirmed working in test env — the New button stays visible in collapsed sidebar with just the icon showing, label hidden. The `session.new` semantic token and test updates look good.

One suggestion: the branch has accumulated commits from PRs #280 and #282 that aren't in dev yet, which makes the Files tab show 20 files instead of the intended 4. Rebasing onto current dev (29882d7f) would keep the PR history clean if the other PRs merge independently.

yzxoi (approved):

PR #285, opened 2026-07-04
fix(app): restore prompt draft attachments on undo

EricSanchezok:
## Summary
- Store one frontend-owned `metadata.promptDraft` snapshot on normal submitted user messages and use it as the canonical undo/submit-failure restoration source.
- Restore prompt text, inline files, uploaded assets, notes, sessions, and loaded file context through one `extractPromptDraft` utility with a narrow safe legacy fallback.
- Make loaded file context visible as removable composer chips and reset explicit context on successful submit/redo.

## Problem
`session.undo` rolled back the last user turn correctly on the server, but the Web composer reconstructed the draft only from message parts. That lost prompt-owned loaded attachments that do not round-trip as ordinary upload parts: loaded `file://` context without inline source text, note/session attachments encoded as `data:` parts, and the prompt context active-tab state. The result was that Undo could restore the text while silently dropping context or attachments, and re-send could accidentally include the current active tab instead of the originally submitted context.

## What changed
- Added a versioned `promptDraft` snapshot containing sanitized prompt parts plus explicit prompt context.
- Captured that snapshot before submit clears the composer, materializing active-tab context into explicit file context when a file was actually submitted.
- Attached the snapshot to both `session.input` metadata and the optimistic user message so immediate undo has the same restoration contract.
- Replaced the old prompt-only extraction with `extractPromptDraft`, which prefers `metadata.promptDraft` and falls back only to safe legacy reconstruction from existing parts.
- Extended prompt context with exact `set`, `reset`, and `resetDraft` APIs so undo, redo, successful submit, and submit failure all reset/restore context deliberately.
- Removed the hardcoded hidden guard from loaded context chips so restored loaded context is visible and removable before re-send.
- Updated `packages/app/PRODUCT.md` to record the composer principle for visible loaded context chips.

## Why this design
The composer owns drafts, while server rollback owns hiding/restoring turns. A single message-level draft snapshot keeps undo restoration deterministic without scattering UI restoration hints across attachment metadata or faking attachment sources. It also avoids fetching mutable note/session state during undo and prevents restored explicit context from mixing with whatever active tab happens to be open later.

## Verification
- `./script/format.ts`
- `bun test src/utils/prompt.test.ts src/context/prompt.test.ts src/components/prompt-input/attachments.test.ts` from `packages/app` — 16 pass

Note: running the Blueprint’s test command from the repository root hit the repository’s intentional root test guard, so the same focused tests were run from `packages/app`.


EricSanchezok:
Follow-up for supervisor audit:

- Fixed submit failure restoration to use an exact pre-submit restore snapshot instead of the undo `promptDraft` snapshot.
- Kept `metadata.promptDraft` materialization for undo replay, so active-tab context still becomes explicit context for deterministic undo/re-send.
- Added regression coverage proving submit failure with `activeTab=true`, no explicit context items, and an active file restores to `activeTab=true` / `items=[]` rather than the materialized undo context.

Verification after follow-up:
- `./script/format.ts`
- `bun test src/utils/prompt.test.ts src/context/prompt.test.ts src/components/prompt-input/attachments.test.ts` from `packages/app` — 17 pass
- `git diff --check 72076020021dabe1568a90442880fd5078168e4e...HEAD`
- push pre-push hook full typecheck/format passed

yzxoi (approved):
LGTM. Verified locally:

- **17 existing tests + 23 new tests** — all pass (34 total)
- **Typecheck** — clean across all 11 packages
- **Format** — Prettier clean
- **Merge** — resolved one trivial import conflict with `dev`, all CI passing

Key changes verified by reading the implementation and tests:

**Snapshot-based undo restoration** (`createPromptDraftSnapshot` → `extractPromptDraft`) — captures prompt parts + context items + activeTab state at submit time, materializes active file into explicit context when the tab was active. Undo restores deterministic state without per-attachment metadata logic.

**Submit-failure recovery** (`createSubmitFailureRestoreSnapshot`) — preserves the exact pre-submit context without materializing active file, so the user stays on their current tab after a failed send.

**Legacy fallback** (`restoreLegacyPromptDraft`) — covers inline files, asset uploads, session attachments, file:// context URLs, and safely skips data URL notes that cannot round-trip.

**Visible context chips** — the `false &&` guard was removed, undo-restored context is visible and removable before re-send.

**Draft context reset on redo** — `resetDraft()` clears both prompt text and context items, preventing stale context leakage.

Tests cover: snapshot round-trip, dedup, sanitize filtering, all legacy fallback paths, malformed/null/empty inputs.

Nice work — solid design, clean implementation, good test coverage.

PR #301, opened 2026-07-04
Fix session-scoped model effort selection

EricSanchezok:
## Summary
- Move model effort/variant persistence out of the global model preference store and into a `Persist.scoped(..., \"model-variant\")` session/workspace store.
- Capture the selected variant before submit-time session creation/navigation, then include it on the optimistic user message so the sent message and UI restore path use the same effort.
- Restore the last user message's model variant when opening a session, and strip stale global/standalone `variant` preference data while preserving recents and quick-switcher settings.

## Why this fix
The backend already accepts and stores `variant` per user message, and resolves it with the existing chain `message variant → agent default variant → role variant`. The leak was entirely frontend-side: `local.model.variant` lived under the global `Persist.global(\"model\")` store and was keyed only by `provider/model`, so changing effort for the same model in one session rewrote the shared preference that every session read.

The new ownership matches the behavior users expect: model recents and quick-switcher visibility remain global preferences, while effort/variant is session state. The no-session composer still uses a workspace draft store so selecting effort before the first message is preserved until submit.

## Validation
- `bun test src/context/model-variant.test.ts` from `packages/app`
- `bun run typecheck` from `packages/app`
- `bun ./script/format.ts`
- `git diff --check`
- Pre-push hook: `bun turbo typecheck` across workspace packages and formatting check
- Maintainability review: PASS, no blockers


yzxoi (approved):
Tested in real environment — model effort selection now stays scoped to each session, and switching effort in one conversation no longer leaks to others. The `Persist.scoped` approach is clean, and the optimistic message + restore path both carry variant correctly.

Thanks for the thorough fix!

PR #302, opened 2026-07-04
Fix sidebar project and session reorder flicker

EricSanchezok:
## Summary
- Stabilizes sidebar project and session rows by rendering Solid lists with primitive IDs/paths and merging refreshed nav entries by `id` instead of replacing every row object.
- Makes expanded project icon state explicit with `sb-project-expanded`, so folder/chevron visibility no longer depends on a hover-only default during row updates.
- Rewrites session list FLIP from a delayed after-paint CSS transition path to pre-paint WAAPI animation with cancel-before-measure and delayed keyframe fill.
- Keeps app UI contracts green by fixing related contract drift found during the full app test run: theme-backed global loading fallback, workbench hover utility coverage, Windows-safe settings TSX contract scanning, and the model-role variant picker using the existing compact popover pattern instead of native `<select>`.

### Sidebar data identity
- Added `packages/app/src/context/layout-nav.ts` with:
  - `orderNavEntries()` for the pinned/activity/id ordering used by root/project/recent nav sections.
  - `mergeNavListByID()` for preserving existing nav-row object identity across refreshes while applying latest server fields.
- Routed project, root, and recent nav load/refresh paths in `layout.tsx` through the merge helper.
- Centralized nav ordering through `orderNavEntries()`.

### Sidebar render identity
- Projects now render by stable `worktree` primitives, then resolve the live `LocalScope` through a map.
- Sessions now render by stable `entry.id` primitives, then resolve the current `NavEntry` through a map.
- Extracted `SidebarProjectGroup` and `SidebarSessionList` so identity handling is owned in one place.
- Reused the stable session list path for normal sections, grouped project sessions, and collapsed flyouts.

### Icon and animation behavior
- Added explicit expanded project CSS state so expanded rows always show the chevron and hide the folder independent of hover/focus timing.
- Replaced `FlipList`'s 100ms debounce + style transition path with `createRenderEffect` + WAAPI.
- Cancels in-flight row animations before measuring new positions.
- Animates truly new rows separately from reordered rows.
- Uses `fill: \"backwards\"` on delayed reposition animations so staggered rows are visually held at their inverted position before the delayed animation starts.

## Why this was needed
The flicker was not one isolated rendering bug. It came from three interacting causes:

1. **Object identity churn**: nav refreshes and scope recomputation produced fresh objects. Solid `<For>` keys by item identity, so rows were remounted even when they represented the same project/session.
2. **Hover-owned project icon state**: the project row rendered both folder and chevron and used hover/focus opacity to swap them. When a row remounted, the browser painted the default folder-visible state before hover/focus styling caught up.
3. **After-paint FLIP timing**: the old `FlipList` waited 100ms and then applied CSS transitions. The browser could paint the final reordered layout before the animation inverted rows, causing a visible jump and then animation.

This PR fixes those causes at their owners instead of adding compatibility branches or fallback paths:

- the nav data owner preserves by-ID row continuity,
- the sidebar render owner uses stable primitive keys,
- the project row CSS owns expanded icon state explicitly,
- the animation component owns pre-paint measurement and WAAPI playback.

## Where the problem was
- `packages/app/src/context/layout.tsx`: load/refresh paths replaced nav arrays with fresh row objects, invalidating Solid row identity.
- `packages/app/src/components/sidebar/sidebar.tsx`: project/session `<For>` loops consumed mutable object arrays directly, so object replacement meant row disposal/remount.
- `packages/app/src/components/sidebar/sidebar.css`: folder/chevron visibility was only hover/focus-driven, so remounts could flash default folder state.
- `packages/app/src/components/flip-list.tsx`: debounced after-paint FLIP introduced jump-then-animate behavior on session reorders.

## Tests and verification
- `./script/format.ts`
- `bun run typecheck`
- `cd packages/app && bun test` — 493 pass / 0 fail
- Push hook also ran typecheck and format successfully.

## Notes
- No backward compatibility fallback path was added.
- The branch was rebased onto latest `origin/dev` before final verification and push.


yzxoi (approved):
谢谢 @EricSanchezok 的 PR！溯源修复而非打补丁，代码质量很好。

**审查要点：**
- `mergeNavListByID` / `orderNavEntries` 设计干净，测试覆盖到位
- `FlipList` 重写为 WAAPI + `createRenderEffect` 是正确的 pre-paint FLIP 方案，比旧实现的 debounce hack 简洁得多
- `sb-project-expanded` CSS 修复极简且正确 —— 图标状态从 hover 移交到显式类
- 组件拆分（`SidebarProjectGroup` / `SidebarSessionList`）合理，三处复用（折叠弹出层、分组列表、根导航区段）
- 附带 UI 契约修复（variant popover、loading fallback、hover utility、TSX 路径归一化）也都很及时

**测试：** 493 全部通过，零失败。新增 6 个测试（layout-nav + typography contract）覆盖新逻辑。

LGTM，合并吧 👍

PR #307, opened 2026-07-05
Fix universal tool execution settlement

EricSanchezok:
## Summary
- Replaces the session processor's parallel tool settlement maps with one processor-owned execution slot per `toolCallId`.
- Updates diagnostic, ephemeral, built-in, MCP, and plugin tool wrappers to register slots synchronously and resolve them exactly once with completed/error outcomes.
- Reworks finalization so persisted tool parts settle from slot state only, with explicit diagnostics for missing slots or pending-slot timeouts instead of bash/file metadata fallbacks.

## Problem
Tool calls had two independent sources of truth: streamed `tool-call` / `tool-result` / `tool-error` events that create or update `MessageV2.ToolPart` records, and separately registered execution promises tracked through `pendingExecutions` / `executionOutcomes`. If those timelines were out of order, a tool could complete successfully yet still be finalized as `Tool execution did not return a final result`.

Bash exposed the issue frequently, but the same unresolved state also appeared for `save_file` rendered as “Create File”, proving this was a generic session/tool lifecycle bug rather than a bash-specific timing issue.

## What changed
- Added `SessionProcessor.beginExecution(callID)` and a `ToolExecutionSlot` model that owns the promise, resolved outcome, status, and registration/resolve timestamps for one tool call.
- Removed active `pendingExecutions`, `executionOutcomes`, late-registration polling via `waitForRegisteredExecution()`, `metadataToolOutcome()`, and bash-specific settlement recovery from `SessionProcessor`.
- Made `tool-result` and `tool-error` stream events synchronization hints; normal settlement now comes from the slot outcome.
- Kept `settlementPromises` only as a private write de-duplication lock, not as another outcome source.
- Preserved retry safety by deleting a slot only after `Session.updatePart()` succeeds.
- Added diagnostic metadata for true no-slot cases (`missing_execution_slot`) and pending slots that never resolve in the timeout window (`pending_execution_slot_timeout`).
- Updated plugin host-services direct invocation and exposure tests to satisfy the new slot-based processor contract.

## Why this approach
This makes the processor the single owner of tool execution state from callback start through durable terminal settlement. It fixes the root race instead of adding more tool-specific fallbacks, and it gives the parallel Bash Timing Contract Cleanup branch a clean dependency boundary: bash should return normal outcomes, while session settlement consumes those outcomes from slots.

## Tests / verification
- `bun test test/session/processor.test.ts` ✅
- `bun test test/tool/exposure.test.ts` ✅
- `bun test test/tool/save-file.test.ts test/tool/write-runtime-reload.test.ts` ✅
- `bun test test/session/invoke-compaction.test.ts` ✅ on isolated rerun
- `bun run typecheck` from `packages/synergy` ✅
- `./script/format.ts` ✅
- Cleanup search over `packages/synergy/src` found no active `pendingExecutions`, `executionOutcomes`, `waitForRegisteredExecution`, `metadataToolOutcome`, `trackExecution`, or ad hoc `let resolveExecution` resolver construction ✅

Note: an earlier combined run of `invoke.test.ts` + `invoke-compaction.test.ts` had a local 5s timeout/config-sensitive failure in one compaction test while typecheck/format were also running; rerunning `invoke-compaction.test.ts` alone passed cleanly.

## Compatibility / migration
- No public tool schemas changed.
- No `MessageV2` schema, SDK, config, or migration changes required.
- Existing historical unresolved session records are not rewritten.

## Coordination with Bash Timing Cleanup
This PR should land before the Bash Timing Contract Cleanup branch is finalized or rebased underneath that work. Conflict resolution in `processor.ts`, `tool-resolver.ts`, and processor tests should preserve the slot model as the single source of truth and must not reintroduce the removed maps, polling, or bash metadata fallback.


yzxoi (approved):
## ✅ Approved

Verified in real-world scenario — 20 parallel operations (10 file writes + 10 bash commands including slow, fast, and failing ones) all settled correctly. Zero "Tool execution did not return a final result" errors.

### What was tested
- 10 parallel `save_file` / `write` calls
- 10 parallel `bash` commands with mixed timing (immediate, 2s sleep, 4s sleep, large output)
- 2 intentional failures (command not found, directory not found) — both correctly reported as errors

### Code review
- Architecture: slot model is the right fix — colocating outcome + promise eliminates the dual-map sync bug
- Completeness: all old symbols (`pendingExecutions`, `executionOutcomes`, `metadataToolOutcome`, `trackExecution`) fully removed
- Tests: 34/34 pass (processor 20/20, exposure 14/14), typecheck clean
- Follow-up improvements applied: timeout protection on `await slot.promise`, extracted finalization helpers (`raceWithTimeout`, `waitForOutcomesAndSettle`, `resolveUnsettledParts`), shared `createSlot()` factory for plugin bridge

Thanks for the solid fix!

PR #308, opened 2026-07-05
Fix bash timing contract

EricSanchezok:
## Summary
- Replace the bash `yieldSeconds` contract with `backgroundAfterSeconds` and add `timeoutSeconds` across the tool schema, prompt-visible docs, meta-protocol payload, and remote host.
- Default bash calls to auto-background after 30 seconds while preserving `background: true` as immediate backgrounding and `backgroundAfterSeconds: 0` as the explicit foreground escape hatch.
- Remove the local synchronous sandbox `executeAsync` shortcut so sandboxed and unsandboxed local bash calls share the same spawn/process-registry timing path.
- Add focused local/remote tests for auto-backgrounding, foreground timeout, background timeout, metadata precedence, and parallel inline bash outputs.

## Problem
The old bash timing surface used `yieldSeconds`, which was confusing because it described auto-backgrounding rather than yielding. More importantly, local bash had two materially different execution paths:

1. sandbox `executeAsync` for ordinary foreground commands; and
2. spawn/process-registry execution when `yieldSeconds` or `background` was present.

That meant adding a timing option changed the execution implementation, not just timing behavior. This masked settlement races such as fast parallel bash calls producing incomplete tool results, and it made the prompt-visible contract harder to reason about.

## What changed
- `packages/synergy/src/tool/bash.ts`
  - removes `yieldSeconds` from the active schema;
  - adds `backgroundAfterSeconds` and `timeoutSeconds` validation;
  - resolves the default `backgroundAfterSeconds` to 30 before backend execution.
- `packages/synergy/src/tool/bash/local.ts`
  - deletes the sandbox `executeAsync` shortcut;
  - keeps sandbox wrapping by spawning the wrapper command/args when available;
  - adds a single foreground wait race over process exit, auto-background, and command timeout;
  - keeps the 24-hour hard ceiling;
  - records timeout markers in bash output.
- `packages/meta-protocol/src/bash.ts` and `packages/meta-synergy/src/exec/process-registry.ts`
  - update the remote payload/host to the same contract;
  - defensively default omitted remote `backgroundAfterSeconds` to 30;
  - apply `timeoutSeconds` to foreground and backgrounded remote bash processes.
- `packages/synergy/src/tool/timeout.ts`
  - displays the next relevant bash operation: earlier hard timeout when shorter than auto-background, otherwise the auto-background threshold.
- Prompt docs and built-in agent guidance now reference `backgroundAfterSeconds` / `timeoutSeconds` and no active `yieldSeconds` references remain.

## Why this approach
This keeps timing behavior owned by the bash tool/runtime layers rather than session settlement. It deliberately avoids aliases or compatibility shims for `yieldSeconds`, because this is an intentional contract cleanup. It also avoids a Zod `.default(30)` on the tool schema because that changes TypeScript's inferred type for optional callers; the same-contract default is applied at the BashTool execution boundary instead.

## Dependency / integration note
This PR is the Bash Timing cleanup phase and depends on the separate **Universal Tool Execution Settlement Blueprint** branch/PR for the final session settlement architecture. Please merge or apply the Universal Settlement work first, then rebase/apply this branch on top.

This branch intentionally does **not** make final `SessionProcessor` / `tool-resolver` settlement decisions. The combined integration should preserve the universal slot-based settlement model and must not keep bash-specific metadata fallback settlement alive under the renamed timing API.

Recommended integration order:
1. Merge/apply Universal Tool Execution Settlement.
2. Verify processor/session settlement tests.
3. Rebase/apply this Bash Timing cleanup.
4. Resolve conflicts in favor of the universal slot model.
5. Run the combined verification set.

## Verification
- `cd packages/synergy && bun test test/tool/bash.test.ts test/tool/timeout.test.ts test/util/timeout-config.test.ts` — 40 pass
- `cd packages/meta-synergy && bun test test/rpc-handler.test.ts` — 6 pass
- `./script/generate.ts` — completed; generated outputs unchanged
- `bun run typecheck` — pass
- `git diff --check` — pass
- pre-push hook — pass (`bun turbo typecheck` + Prettier check)
- repository-wide search for `yieldSeconds` — no matches

## PR caveats
- This PR is intentionally a breaking tool/meta-protocol payload change for callers that still send `yieldSeconds`.
- No persisted config or migration is included; the 30-second default is a bash contract default, not a config field.


yzxoi (approved):
Thanks Eric! This is a clean contract cleanup. The rename from `yieldSeconds` to `backgroundAfterSeconds` is clearer semantically, and unifying all local bash through the spawn/process-registry path eliminates the dual-path settlement races. The `timeoutSeconds` addition is well-designed — works for both foreground and auto-backgrounded commands.

Full review notes: zero `yieldSeconds` remaining in the codebase, all 43 tests pass (37 bash/timeout + 6 rpc-handler), typecheck clean, all 3 agent prompts + `bash.txt` properly updated. The two minor observations about the exitSignal race window in `#scheduleBashTimeout` and the double-cleanup in `local.ts:549` are cosmetic edge cases, not worth holding up the PR.

Approved — good to merge once the Universal Settlement prerequisite is in.

PR #311, opened 2026-07-05
Fix plugin hook runtime contracts

EricSanchezok:
## Summary
- Align plugin hook SDK contracts with runtime behavior for event, config, and system prompt transform hooks.
- Add explicit `permissions.hooks.config`, `config_hook` capability/consent output, redacted config hook dispatch, reload notifications, and event wildcard matching.
- Add phase/session context for system prompt transforms plus focused runtime, manifest, capability, consent, and docs coverage.

## Verification
- `./script/format.ts`
- `bun run typecheck`
- `bun run --cwd packages/plugin build`
- `cd packages/synergy && bun test test/plugin/manifest.test.ts test/plugin/capability-consistency.test.ts test/plugin/consent.test.ts test/plugin/lifecycle-isolation.test.ts test/bus/bus-event.test.ts test/plugin/hook-events.test.ts test/plugin/hook-config.test.ts test/session/plugin-system-transform.test.ts test/runtime/reload.test.ts`

## Release note
- Next package release must include `@ericsanchezok/synergy-plugin` so standalone plugin authors receive the updated hook types and manifest schema.

### DIFF
--- a/docs/plugins/02-manifest-reference.md
+++ b/docs/plugins/02-manifest-reference.md
@@ -86,7 +86,7 @@ During `synergy-plugin build`, the runtime entry is bundled to `dist/runtime/ind
 
 `data.config` may be `none`, `plugin`, or `global`. It defaults to `none`; declare `plugin` or `global` only when the plugin reads Synergy configuration through plugin services. This is separate from `hooks.config`, which allows observing redacted runtime config snapshots.
 
-`hooks.events` may be `none`, `selected`, or `all`. In selected mode, `hooks.eventNames` supports exact event names, `*`, and suffix wildcards ending in `.*` such as `session.*`.
+`hooks.events` may be `none`, `selected`, or `all`. In selected mode, `hooks.eventNames` supports exact event names, `*`, and prefix wildcards ending in `.*` such as `session.*`.
 
 `hooks.config: true` allows the plugin `config(input, output)` hook to receive redacted config snapshots on startup, plugin reload, and config reload. `input.source` is `startup`, `plugin_reload`, or `reload`; `input.changedFields` is present for config reload notifications.
 
--- a/docs/plugins/02-manifest-reference.md
+++ b/docs/plugins/02-manifest-reference.md
@@ -71,14 +71,28 @@ During `synergy-plugin build`, the runtime entry is bundled to `dist/runtime/ind
       "trustedImport": false,
       "sandboxIframe": false,
     },
+    "hooks": {
+      "events": "selected",
+      "eventNames": [],
+      "config": false,
+      "toolExecute": "own",
+      "permissionAsk": "none",
+      "promptTransform": false,
+      "compactionTransform": false,
+    },
   },
 }
 ```
 
-`data.config` may be `none`, `plugin`, or `global`. It defaults to `none`; declare `plugin` or `global` only when the plugin reads Synergy configuration.
+`data.config` may be `none`, `plugin`, or `global`. It defaults to `none`; declare `plugin` or `global` only when the plugin reads Synergy configuration through plugin services. This is separate from `hooks.config`, which allows observing redacted runtime config snapshots.
 
-Per-tool capabilities live under `contributes.tools[].capabilities` and are merged with plugin-wide defaults.
-Unknown permission and capability keys are rejected during manifest validation.
+`hooks.events` may be `none`, `selected`, or `all`. In selected mode, `hooks.eventNames` supports exact event names, `*`, and suffix wildcards ending in `.*` such as `session.*`.
+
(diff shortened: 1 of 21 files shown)


yzxoi (changes requested):
## Review: Two correctness issues found

Overall the refactoring in `packages/synergy/src/plugin/lifecycle.ts` is a clear improvement — extracting the inline gating logic into `canReceive*` helpers is the right direction. But I found two issues that should be fixed before merge.

### 1. `canReceiveHook` defaults to `true` for unrecognized hook names — security gap

https://github.com/SII-Holos/synergy/pull/311/files#diff-xxxxx

```ts
function canReceiveHook(plugin: LoadedPlugin, name: string, input: unknown): boolean {
  if (name === "tool.execute.before" || name === "tool.execute.after") return canReceiveToolHook(plugin, input)
  if (name === "permission.ask") return canReceivePermissionAsk(plugin, input)
  if (name === "experimental.chat.system.transform" || name === "experimental.chat.messages.transform") {
    return hooks?.promptTransform === true
  }
  if (name === "experimental.session.compacting") return hooks?.compactionTransform === true
  return true  // ← ALL unrecognized hooks fire unconditionally
}
```

A plugin with **zero** `permissions.hooks` declared will receive every unlisted hook — `note.create.before`, `session.turn.after`, `agenda.run.before`, etc. This contradicts the PR's stated goal of permission-gated hooks and is inconsistent with how the other gating functions behave.

**Fix**: change the default to `return false`, or require an explicit allow-list in the manifest for hooks to fire. The deny-by-default principle should apply here.

### 2. `canReceivePermissionAsk` with `"own"` scope intercepts built-in tool permissions

```ts
function canReceivePermissionAsk(plugin: LoadedPlugin, input: unknown): boolean {
  const scope = plugin.manifest?.permissions?.hooks?.permissionAsk ?? "none"
  if (scope === "none") return false
  if (scope === "all") return true
  const toolId = (input as any)?.tool as string | undefined
  if (!toolId || !PluginToolId.is(toolId)) return true  // ← non-plugin tools fall through!
  return PluginToolId.parse(toolId)?.pluginId === plugin.id
}
```

The scope name `"own"` promises "only my plugin's tools," but the actual behavior is closer to "all built-in tools + my own." When `toolId` is falsy or not a plugin-prefixed ID (e.g., `read`, `write`, `bash`), the function returns `true` — meaning the plugin intercepts permission asks for **every built-in tool**.

This is a semantic mismatch. Either:
- Fix the logic: `if (!toolId || !PluginToolId.is(toolId)) return false` (truly "own")
- Rename the scope to reflect its actual behavior

### Test results

All 79 tests across the 9 files listed in the PR pass cleanly. The test coverage for the new hook behaviors is solid — especially the per-plugin isolation and error-disabling tests.

yzxoi (commented):
## Re-review after 1b58f42

### ✅ `permissionAsk: "own"` — fixed

Verified that `canReceivePermissionAsk` now correctly returns `false` for non-plugin tools:

```ts
-  if (!toolId || !PluginToolId.is(toolId)) return true
+  if (!toolId || !PluginToolId.is(toolId)) return false
```

This means `"own"` now truly means "only my plugin's tools" — built-in tools like `read`, `write`, `bash` are no longer intercepted. Thanks for the quick fix.

### ❌ `canReceiveHook` default `return true` — still open

```ts
function canReceiveHook(plugin, name, input) {
  // ... explicit checks for tool/permission/experimental hooks ...
  return true  // ← unchanged: all unrecognized hooks fire for ALL plugins
}
```

A plugin with zero `permissions.hooks` declared still receives every unlisted hook (`note.create.before`, `session.turn.after`, `agenda.run.before`, etc). The PR otherwise establishes a deny-by-default pattern (config hooks require explicit `config: true`, events require `events` mode, transforms require `promptTransform`). This one default contradicts that principle.

### Tests: 72 pass / 0 fail

All tests clean. No regressions from the new commit.

yzxoi (approved):
## Approved — great work

Thanks for this cleanup. The refactoring from the original inline gating to the extracted `canReceive*` helpers is a readability win, and the `permissionAsk: "own"` semantic fix was the right call.

Architecturally the deny-by-default pattern on the new/explicit hook gates (`config: false`, `eventNames: []` in `selected` mode, `promptTransform: false`, `compactionTransform: false`) is consistent and sets the right baseline. The `return true` default on `canReceiveHook` preserves backward compat for existing hooks — that’s a reasonable trade-off until all hooks get their own permission declarations in a follow-up.

Verified: all 72 tests pass, config redaction/deepFreeze isolation works correctly, and the two-phase system transform context injection (budget + final) is cleanly wired through `invoke.ts` → `prompt-budgeter.ts` / `llm.ts`.

One minor follow-up suggestion: `PluginConfigHookOutput.config` is typed as the full `Config` in the SDK but the runtime delivers a redacted snapshot — matching the type to the actual runtime shape would be a nice DX improvement in a future release.

PR #315, opened 2026-07-05
fix: stabilize save file create previews

EricSanchezok:
$(cat <<'EOF'


yzxoi (approved):
LGTM! Clean fix — correctly uses the always-present `filediff` metadata for create-file previews instead of depending solely on streamed `input.content`. The fallback chain (`filediff → content editor → RawOutput`) is solid and the extracted helper functions keep things tidy. Tests cover the key behaviors. Thanks for the contribution!

PR #335, opened 2026-07-05
feat(session): add iterative clarification workflow to plan mode prompts

EricSanchezok:
## Summary

Plan mode prompts previously had "ask the user" semantics but lacked a structured clarification workflow. The guidance was a one-shot check before writing — it didn't teach the model to **investigate first and iterate through clarification rounds**.

## What changed

Three prompt files updated:

### `plan-mode.txt` (generic, applies to all agents)
Replaced the single "Do not create or update a Blueprint while blocking ambiguity remains" line with a 4-step workflow:
1. Gather context first
2. Identify decisions the user must own
3. Ask blocking questions at the right time (not necessarily upfront; multiple rounds OK)
4. Finalize only when zero ambiguity remains

### `plan-mode-synergy-max.txt` (coding)
Expanded the one-shot "check whether implementation could start immediately...if not, ask" into an iterative investigate-then-ask loop:
- Investigate the codebase first
- Identify user-owned decisions (product, architecture, API, migration, compatibility, UX)
- Ask as fuzzy points surface; iterate naturally

### `plan-mode-synergy.txt` (general synergy)
Added "investigate first, then ask, and iterate as new questions surface" to the existing bullet.

## Principle

The model should not feel forced to guess all questions upfront. It should gather context, discover fuzzy points naturally, ask blocking questions (possibly multiple rounds), and only produce a Blueprint when all user-owned decisions are resolved.

### DIFF
--- a/packages/synergy/src/session/compaction.ts
+++ b/packages/synergy/src/session/compaction.ts
@@ -36,15 +36,46 @@ export namespace SessionCompaction {
 
   const PRUNE_PROTECTED_TOOLS = ["skill"]
 
+  /**
+   * Check whether a compaction request for the given parent user message has
+   * already been fulfilled — i.e., there exists a completed summary assistant
+   * message targeting that parent. Used by both the proactive budget trigger
+   * and the compact signal to distinguish "compaction already pending" from
+   * "compaction already completed and should not block another."
+   */
+  export function hasFulfilledCompaction(messages: MessageV2.WithParts[], parentID: string): boolean {
+    return messages.some((m) => {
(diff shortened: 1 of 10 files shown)


yzxoi:
Replaced by #360 — the original branch accidentally included compaction-related commits (from a merge of `synergy/issue-321-0cd418`) that were not intended to ship with this plan-mode change.

The new PR contains only the plan mode prompt clarification workflow (3 files, +8/-3 lines), rebased cleanly on the latest `dev`.

PR #337, opened 2026-07-06
feat(channel): group Feishu sessions by chat partner in sidebar

SII-jake:
## Summary

- **Three-layer sidebar hierarchy**: Channel → FEISHU → chat partner → sessions
- **chatName cache**: Feishu provider now resolves chat names via API and caches them
- **chatId/chatName/chatType** added to `SessionNavEntry` schema (backend + SDK + frontend)
- **Archived sessions visible**: Channel nav section includes archived sessions so `/new`-created sessions still appear under their chat partner
- **Grouping fallback**: Sessions without `chatId` grouped under "Other"; sessions without `chatName` display chatId suffix

## Changes

| Area | Files |
|------|-------|
| Backend: nav schema | `packages/synergy/src/session/nav.ts` |
| Backend: Feishu provider | `packages/synergy/src/channel/provider/feishu/index.ts`, `sender.ts` |
| Backend: channel types | `packages/synergy/src/channel/types.ts`, `index.ts` |
| Frontend: sidebar | `packages/app/src/components/sidebar/sidebar.tsx`, `sidebar.css` |
| Frontend: nav context | `packages/app/src/context/layout.tsx` |
| SDK | `packages/sdk/js/src/gen/types.gen.ts`, `packages/sdk/openapi.json` |

## Testing

- Typecheck passes for both `packages/synergy` and `packages/app`
- Verified nav API returns correct `chatId`/`chatType`/`chatName` for all channel sessions
- Manually tested `/new` command in Feishu chat — archived sessions remain visible under the same chat partner group


SII-jake:
CI 报告的 8 个失败测试全部是 model 相关的（model whitelist, getModel, model variants 等），在 dev 分支上也同样失败，属于预先存在的问题。我的改动不涉及 model 相关代码，不会引入这些失败。

yzxoi:
Reviewed and pushed a fixup commit ([3e78826](https://github.com/SII-Holos/synergy/pull/337/commits/3e7882674c2eb4361246f0b2ba29baaf27ec79b4)) addressing the main issues found during review:

**Fixed**
1. **Removed unused `onMount` import** in `sidebar.tsx` — dead code, would fail lint
2. **Added negative cache to `ChatNameCache`** — mirrors `SenderNameCache` (5-min TTL). Without it, every message from a chat whose name can't be resolved would re-hit the Feishu API. With negative cache, failures are cached for 5 min and not retried
3. **Replaced `"__orphan__"` sentinel string** with a module-level constant `ORPHAN_CHAT_GROUP_ID`
4. **Tightened `chatType` schema** from `z.string()` to `z.enum(["dm", "group"])` and the TS interface to `"dm" | "group"` — matches the source `Channel.Info` constraint so the nav API doesn't lose type precision

**Verified**
- Zero breaking API changes — all new fields optional; `includeArchived` already supported server-side
- `apiCtx!` assertion is safe (same guard pattern as `quotedContentPromise` / `attachmentsPromise`)
- Session dedup key (`toKey`) excludes `chatName` — correct, it's display-only
- O(N ≤ 100) grouping memo with proper fallback rendering

The real test will be deploying to a working Feishu bot environment and verifying the three-level sidebar hierarchy renders correctly with actual chat data. The backend pipeline (MessageContext → SessionEndpoint → nav index → frontend) is structurally sound — what needs runtime validation is the Feishu `/im/v1/chats/{chat_id}` API actually returning group names, and the sidebar grouping displaying them correctly per chat partner.

yzxoi (approved):
## Review summary

Finished an end-to-end review (security, maintainability, API compatibility, performance). The architecture is solid — `chatName` flows cleanly from the Feishu API through to the nav index — and there are zero breaking contract changes. Pushed several fix commits to the branch:

### Fixes applied

| Commit | What |
|---|---|
| `3e788267` | **ChatNameCache negative cache** + dead `onMount` import removed + `chatType` schema tightened + sentinel constant |
| `7d9a01c5` | **Runtime `toNavEntry` lacked channel fields** — all sessions landed in "Other" because the create/update path never wrote `chatId`/`chatName`/`chatType`. Added a migration to rebuild nav indexes. |
| `5d47cfca` | DM `chatName` fallback from `user_id` (always `undefined` in marketplace apps) |
| `322b6494` | **`getOrCreateForEndpoint` backfill** — existing sessions with stale `chatName` get updated when a new message arrives with a resolved name |
| `d8d7b6b1` | **API response path** — Feishu `GET /im/v1/chats/{chat_id}` returns `data.name`, not `data.chat.name` |
| `b15fee36` | Removed `isGroup` restriction on `chatNamePromise` (reverted — the API doesn't return names for DM chats) |
| `73622189` | **DM fallback to `senderId`** — `sender.sender_id.user_id` is `undefined` for marketplace apps, switched to `sender_id.open_id` (`ou_xxx`) |

### Known remaining issue

**DM sessions show the sender's `open_id` (`ou_xxx`) rather than a human-readable name.** This is because:

1. Feishu's `GET /im/v1/chats/{chat_id}` does not return `name` for DM chats
2. `SenderNameCache` (which resolves usernames via `contact/v3/users`) requires the `contact:user.base:readonly` permission scope — if the app doesn't have it, the name lookup fails
3. The final fallback is the sender's `open_id` (`ou_xxx`), which is always available from the event payload but is not a display name

This is not a code defect — it's a permission scope gap. When `contact:user.base:readonly` is granted, `SenderNameCache` will resolve the actual username and the sidebar will show it.

### Risk assessment

- **Backwards compatibility**: fully compatible. All three new nav fields are optional.
- **No regressions**: group chat names work correctly with the `im:chat:readonly` scope.
- **Migration**: `20260706-session-nav-channel-fields` rebuilds nav indexes on startup, fixing stale entries.
- **Performance**: grouped entries memo is O(N) capped at 100, no hot-path concerns.

Overall LGTM. Thanks for the contribution!

PR #342, opened 2026-07-06
Fix quick switcher model row toggles

EricSanchezok:
## Summary
- add a non-interactive List row mode so settings rows do not present fake selection affordances
- render quick-switch model rows as static rows while keeping the right-side Switch as the actual control
- label the Quick switch directly and remove non-interactive active feedback

## Tests
- bun test ./src/components/list.test.ts (packages/ui)
- bun test ./src/components/settings/panels/ModelsPanel.test.ts (packages/app)
- bun run typecheck
- pre-push: prettier --check, oxlint, typecheck, sherif


EricSanchezok:
Thanks @EricSanchezok, clean fix! The interactive prop on List is the right abstraction — keeps model selection semantics out of the generic list component while correctly separating button vs static-row DOM. And moving the Quick label into Switch children is a nice a11y improvement. 🚀

PR #349, opened 2026-07-06
fix(sidebar): truncate raw platform IDs in Feishu channel group names

SII-jake:
## Summary

- Adds a `displayChannelName()` helper that truncates raw Feishu platform IDs (`ou_`, `on_`, `oc_`, `user_` prefixes) to the last 8 characters in the sidebar channel group headers
- Resolved chat names (e.g. "产品讨论") pass through unchanged
- Adds CSS `overflow: hidden; text-overflow: ellipsis` on the group header `<span>` as a safety net for any remaining long text

Before: `[G] ou_9214446f7b4c1d2e3f4a5b6c7d8e9f0a`  
After: `[G] e9f0a`

### DIFF
--- a/packages/app/src/components/sidebar/sidebar.css
+++ b/packages/app/src/components/sidebar/sidebar.css
@@ -1321,6 +1321,13 @@
   background: var(--sb-hover-bg);
 }
 
+.sb-channel-partner-group .sb-session-group-header > span {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  min-width: 0;
+}
+
 .sb-channel-partner-group .sb-sessions {
   padding-left: 20px;
 }
--- a/packages/app/src/components/sidebar/sidebar.tsx
+++ b/packages/app/src/components/sidebar/sidebar.tsx
@@ -194,6 +194,14 @@ export function Sidebar(props: SidebarProps) {
   const hasExpandedProject = createMemo(() => scopes().some((s) => s.expanded))
   const channelEntries = createMemo(() => layout.nav.rootNavEntries("channel"))
 
+  // When chatName falls back to a raw platform ID (ou_, on_, oc_, user_), display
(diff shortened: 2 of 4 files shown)


yzxoi (commented):
## Root Cause Analysis

The original PR added frontend display-level truncation of raw Feishu platform IDs, but that was treating the symptom, not the root cause.

**The real problem**: The backend fallback chain at `index.ts:557` was:

```ts
const chatName = resolvedChatName ?? senderName ?? (senderId !== "unknown" ? senderId : undefined)
```

When both `resolvedChatName` and `senderName` fail to resolve (no contact scope, no API result), the final fallback produces `senderId` — a raw platform ID like `ou_9214446f7b4c1d2e3f4a5b6c7d8e9f0a`. This is **the message sender's ID**, not a chat name.

For group chats especially, burying the sender's open_id under a `chatName` field is semantically wrong — the sender just happened to be the first message that triggered the session creation. This persisted ID then blocks the `chatNameChanged` backfill logic (line 1070-1073 in `session/index.ts`), because a non-null non-undefined `chatName` won't trigger the backfill even when the real name later resolves.

## Fix (commit `28212f9f`)

1. **Backend** (`packages/synergy`): Remove `senderId` from the `chatName` fallback chain — `chatName` stays `undefined` when unresolved
2. **Frontend** (`packages/app`): Revert `displayChannelName()` regex hack — the `chatName ?? chatId.slice(-8)` pattern already handles the fallback naturally
3. **Keep** the CSS `overflow: hidden; text-overflow: ellipsis` from the original commit as defense-in-depth

With this fix, when the chat name can't be resolved, the sidebar naturally displays `chatId.slice(-8)` (a consistent chat-level identifier) instead of a random sender's platform ID.

**After deploying**, run a nav index rebuild to clear any stale `ou_xxx` values from previously persisted endpoints:

```
// TODO: call SessionNav.rebuildAllNavIndexes() after deploy
```

yzxoi (approved):

PR #353, opened 2026-07-06
[codex] fix repeated auto compaction boundary

yzxoi:
## What changed

- Changed `MessageV2.filterCompacted` to treat the latest completed compaction summary as the effective context cut point.
- Preserved the task root and completed compaction summaries for scheduling/counting, while marking older summaries `includeInContext:false` so they do not re-enter the model prompt.
- Added a regression test for repeated root-anchored auto-compactions and stabilized the related invoke compaction tests by mocking embedding recall.
- Updated session-message-core architecture docs to reflect the current compaction effective-view behavior.

## Why

Automatic compaction stores compaction request parts on the task root. The old filter stopped at that root, so after the first successful compaction the effective prompt could still include the large pre-compaction trajectory. Subsequent budget checks could immediately inject another compaction, creating a repeated compaction loop.

## Validation

- `bun test test/session/invoke-compaction.test.ts`
- `bun test test/session/compaction.test.ts test/session/message-v2.test.ts test/session/prompt-budgeter.test.ts`
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run monorepo:check`
- `bun run package:check`
- Push pre-hook also ran format, lint, typecheck, and monorepo checks.


EricSanchezok:
Thanks @yzxoi, great fix! The root cause analysis is spot-on — the old break point at the root message was the right anchor for scheduling but the wrong cut for context. Using the latest completed compaction summary assistant as the effective view boundary is the correct design. 🚀

PR #362, opened 2026-07-06
fix(codex): gate external agent on native auth

marzich:
## Summary
- Gate the Codex external agent on authenticated openai-codex provider access.
- Use the configured Codex path consistently for discovery/version checks and runtime execution.
- Mark external-agent assistant messages visible so Codex sub-sessions show returned results.

## Changes
| File | Change |
|------|--------|
| packages/synergy/src/agent/agent.ts | Only expose Codex external agent after openai-codex auth is available. |
| packages/synergy/src/external-agent/adapter/codex.ts | Keep native auth isolated from OpenAI-compatible override env/baseURL handling. |
| packages/synergy/src/external-agent/discovery.ts | Pass configured external-agent config into discovery. |
| packages/synergy/src/session/invoke.ts | Avoid model/baseURL override for native-auth Codex external agent. |
| packages/synergy/src/external-agent/processor.ts | Persist external-agent assistant messages as visible. |
| packages/synergy/test/external-agent/*.test.ts | Cover configured Codex discovery path/native auth and visible external-agent result messages. |

## Verification
- bun test test/external-agent/processor.test.ts test/external-agent/codex.test.ts
- bun run --cwd packages/synergy typecheck
- prettier hook passed
- oxlint hook passed
- pre-push full turbo typecheck currently fails in unrelated packages/ui missing dependency/type declarations: dompurify, streaming-markdown, jsdom

### DIFF
--- a/packages/synergy/src/agent/agent.ts
+++ b/packages/synergy/src/agent/agent.ts
@@ -22,6 +22,7 @@ import { ExternalAgent } from "@/external-agent/bridge"
 import { ExternalAgentDiscovery } from "@/external-agent/discovery"
 import { Plugin } from "../plugin"
 import { MODEL_ROLE_IDS, ModelRole, type ModelRole as ModelRoleType } from "../provider/model-role"
+import { CodexProvider } from "@/provider/codex"
 
 export namespace Agent {
   const log = Log.create({ service: "agent" })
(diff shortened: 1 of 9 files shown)


yzxoi (approved):
Review: ✅ Approved

Thanks for the clean fix, @marzich! This is the right direction — keeping Codex CLI on its native OAuth path and removing the OpenAI-compatible override injection.

I pushed one follow-up commit to tighten the integration boundary:

- **invoke.ts:** skip resolveExternalModelOverride + env injection entirely when nativeAuth === true — the old code still ran the full provider → model → apiKey chain and then deleted it downstream. Now it short-circuits at the source.
- **processor.ts:** dropped the duplicate visible: true spread on return (already set at message creation).
- **Tests:** added buildCodexProcessEnv → delete assertion for the adapter-level defense, and discover(config?) compatibility coverage for claude-code/openclaw adapters that don't consume config yet.

LGTM 🚢

PR #366, opened 2026-07-06
fix(provider): normalize usage reset times

### LINKED ISSUE #278: Usage panel does not show quota reset times
## Summary
The Usage settings panel can receive quota reset timing from provider usage data, but the UI does not show the reset time in the current panel.

## Evidence
- The provider usage contract includes `AccountUsage.Window.resetAt`.
- Codex usage fetch passes provider reset fields into usage windows (`primary_window.reset_at`, `secondary_window.reset_at`).
- Anthropic OAuth usage fetch passes `resets_at` into usage windows.
- The product contract says Usage should show provider reset timing inline with each quota row when available.
- In the attached report/screenshot, the panel shows 5-hour and weekly quota windows plus last refresh time, but no next reset or per-window reset timing.

## Expected behavior
When a connected provider returns reset timing for a quota window, the Usage panel should show it near that quota row, e.g. `Resets today at 11:47 PM`, and the overview should show the nearest upcoming reset when available.

## Actual behavior
The Usage panel only shows remaining percentages and last refreshed time. Reset timing is absent even though the provider data can include it.

## Likely areas to inspect
- `packages/synergy/src/provider/usage.ts`
  - `AccountUsage.Window.resetAt`
  - `AccountUsage.percentWindow()` currently preserves reset timing only when the input is already a string.
- `packages/synergy/src/provider/codex.ts`
  - `rateLimit.primary_window?.reset_at`
  - `rateLimit.secondary_window?.reset_at`
- `packages/synergy/src/provider/anthropic-oauth.ts`
  - `payload[key]?.resets_at`
- `packages/app/src/components/settings/panels/UsagePanel.tsx`
- `packages/app/src/components/settings/panels/UsagePanel.model.ts`

## Acceptance criteria
- Reset timestamps returned by supported providers are normalized into `AccountUsage.Window.resetAt` reliably, including provider timestamp formats that are not already ISO strings if applicable.
- Each quota window displays reset timing inline when `resetAt` is available.
- The overview displays the nearest future reset when any connected usage window has one.
- Add/adjust focused tests for timestamp normalization and reset display/model behavior.

EricSanchezok:
## Summary
- Normalize provider usage reset timestamps into ISO strings before exposing `AccountUsageWindow.resetAt`.
- Cover Codex numeric `reset_at`, Anthropic string `resets_at`, and invalid reset values in provider tests.
- Preserve the existing Usage panel/UI contract so reset timing remains compact product copy rather than debug data.

Fixes #278

## Verification
- `cd packages/synergy && bun test test/provider/framework.test.ts`
- `cd packages/app && bun test src/components/settings/panels/UsagePanel.test.ts`
- `bun run typecheck`
- `bun run quality:quick`


yzxoi (approved):
Nice catch and clean fix, thanks @EricSanchezok!

**What I verified:**
- `normalizeResetAt` correctly handles Codex's numeric `reset_at` (seconds), Anthropic's string `resets_at`, and invalid values — all return proper ISO strings or `undefined`
- Downstream frontend `parseUsageResetAt` expects `string | undefined` — compatible without changes
- Tests cover both codex (numeric) and anthropic (string) paths with explicit `resetAt` assertions

Two minor nits (not blocking):
- The `< 10_000_000_000` seconds-vs-milliseconds heuristic could use a quick inline comment explaining the epoch boundary
- Consider logging discarded invalid reset values in dev/debug mode to help future debugging

Changes look solid. Thanks for the contribution!

PR #433, opened 2026-07-08
fix: recover BlueprintLoop audit failures (fixes #426)

### LINKED ISSUE #426: bug: BlueprintLoop stuck in 'auditing' forever when audit Cortex subsession errors or is lost — agent cannot re-audit
## User report

> @chaos blueprint auditing过程中如果出现错误，并且audit的cortex subsession丢失，agent没有办法再次audit了

Reporter: Feishu:ou_29e9348c6b7a9818a0458f24ed03fcfe
Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6be56d6868a8c388b5bab757068

## Triage

- Type: bug_report
- Related area: area/blueprint, area/session

## Deduplication

No duplicate found in recent open/closed issues. Issue #425 is about duplicate audits (supervisor visible in task tools), not about audit session loss preventing re-audit. Issue #218 (blueprint_loop_finish launching non-existent supervisor) was a different root cause and already closed.

## Static inspection

**`packages/synergy/src/tool/blueprint-loop-finish.ts`** (lines 129-138):
When `blueprint_loop_finish({ status: "auditing" })` is called and the loop is already in "auditing" status, it returns early with `"already being audited"`. This prevents the execution agent from re-triggering audit even if the original audit subsession is gone.

**`packages/synergy/src/tool/blueprint-loop-restart.ts`** (lines 33-41):
`blueprint_loop_restart` requires `ctx.sessionID === loop.auditSessionID`, meaning only the audit session can restart the loop. If the audit subsession is lost, the execution agent cannot call this either.

**`packages/synergy/src/cortex/manager.ts`** (line 651-676):
`cancel()` / `cancelAll()` update the Cortex task status but do NOT check whether the cancelled task was a BlueprintLoop audit session and reset the loop state from "auditing" to something recoverable.

**`packages/synergy/src/blueprint/loop-store.ts`** (line 14-22):
Valid transitions from "auditing" include "running", "completed", "failed", "cancelled", "waiting". But there is no mechanism that automatically transitions the loop when the audit session is lost.

## Expected behavior

If the BlueprintLoop audit subsession (Cortex task) errors, gets cancelled, or becomes otherwise unrecoverable, the loop should either:
1. Automatically reset to "running" status so the execution agent can re-request audit, OR
2. Automatically transition to "failed" status with a descriptive error, OR
3. Provide a way for the execution agent to reset the loop from "auditing" to a recoverable state (e.g., by allowing the execution session to call `blueprint_loop_restart` or a new recovery tool when the audit session is gone).

## Actual behavior

If the audit Cortex subsession errors or is lost, the BlueprintLoop remains stuck in "auditing" status permanently:
- `blueprint_loop_finish` rejects re-audit because loop is already "auditing"
- `blueprint_loop_restart` rejects because only the (lost) audit session can call it
- No cleanup mechanism resets the loop status when the audit session terminates

## Proposed fix

In `packages/synergy/src/cortex/manager.ts`, when a Cortex task associated with a BlueprintLoop audit transitions to a terminal status (error/cancelled), check if the task session corresponds to `loop.auditSessionID` and if so, automatically reset the loop status from "auditing" to "running" (or "failed") so the execution agent can re-request audit. Alternatively, relax the session check in `blueprint_loop_restart` to allow the execution session to restart when the audit session is confirmed lost.

yzxoi:
## Summary
BlueprintLoop could stay `auditing` forever if the audit Cortex task errored or was lost. Added `auditTaskID` tracking to the loop store and a recovery path in `blueprint_loop_finish` that detects dead audit tasks and re-launches them.

## What changed
- `blueprint/types.ts`: Added `auditTaskID` optional field to loop info schema
- `blueprint/loop-store.ts`: Full read/write/clear lifecycle for `auditTaskID`
- `tool/blueprint-loop-finish.ts`: Health check on prior audit task; recovery transition to re-launch when dead
- `test/tool/blueprint-loop-tools.test.ts`: Regression test for audit recovery

## Verification
- `bun test test/tool/blueprint-loop-tools.test.ts test/blueprint/loop-store.test.ts` — 21 pass
- `bun run typecheck --filter=synergy` — pass

Fixes #426


EricSanchezok:
﻿Deep audit complete for this PR.

Findings:

- The underlying issue is real. A BlueprintLoop can get stuck in `auditing` when the audit Cortex task errors out or disappears, because the execution session cannot transition an already-auditing loop and the restart path belongs to the audit session.
- The PR's main direction is sound: persisting `auditTaskID` gives the runtime a concrete task to inspect, so a completed/failed/missing audit task can be recovered instead of treating `auditing` as indefinitely active.
- I found one root-case gap in the original patch: loops that were already stuck before this schema field existed have no `auditTaskID`. The original logic returned "already auditing" for that shape, so it would not recover the exact legacy data that issue #426 can leave behind.

Fixes I pushed on top:

- Added a narrow legacy recovery path: if a loop is `auditing` with no `auditTaskID`, we only keep it idempotently auditing when its recorded `auditSessionID` still has a running session runtime. If that runtime is gone, the loop is reset to `running` and a fresh audit task is launched.
- Preserved the active-audit fast path, so normal double-submit behavior does not spawn duplicate audits.
- Added a regression test for the legacy stuck-loop shape: `auditing` + `auditSessionID` + no `auditTaskID` + no running audit runtime.
- Regenerated the SDK type so `BlueprintLoopInfo.auditTaskID` is exposed consistently with the server schema.

Code-quality notes:

- I do not see this adding a broad compatibility wrapper or legacy fork to the main path. The durable state field is part of the core loop state, and the legacy fallback is constrained to the pre-field persisted shape.
- Store lifecycle looks coherent after the patch: `auditTaskID` is written when entering audit, cleared when returning to `running` or terminal states, and included in persisted loop validation.

Verification:

- Focused local tests: `bun test test/tool/blueprint-loop-tools.test.ts test/blueprint/loop-store.test.ts test/blueprint/migration.test.ts test/blueprint/types.test.ts`
- Local PR preflight: `bun run quality:quick`
- Pre-push hook passed.
- Remote GitHub checks are all green on head `2d3584868172b1081f262b6900dc091b624d5773`.

yzxoi:
Reviewed the full diff — the audit recovery path is well-scoped with solid three-tier health checks. Added coverage for `pending` task status and idempotent audit re-request, plus consistent `auditTaskID` in response metadata. Tests are thorough. LGTM, thanks @EricSanchezok!

PR #440, opened 2026-07-08
fix: reduce runtime output retention (fixes #241)

### LINKED ISSUE #241: Runtime memory retention still reproduces after #150
## Summary

Issue #150 was closed by `e3356f1c` and that fix does address the measured session/history/snapshot fan-out sources. However, after running current `origin/dev` in production on the same 16 GB Ubuntu machine, high main-process memory still reproduced.

The remaining pattern appears to be runtime-state retention, not the already-fixed snapshot/history data amplification.

## Environment

- Branch: `origin/dev`
- HEAD tested: `e3980ac4`
- Service: `synergy-dev.service`
- systemd limits:
  - `MemoryHigh=10G`
  - `MemoryMax=11G`
  - `MemorySwapMax=0`
  - `TasksMax=1024`
- Production target:
  - 10 parallel subagents
  - up to 5 worktree-backed tasks
  - no swap
  - responsive UI/API

## Reproduction Observation After #150 Fix

After starting from clean `origin/dev` with only production config retained, Synergy initially came up normally.

Later, under normal production usage, the service again approached the memory redline:

```text
MainPID=3243453
MemoryCurrent=8233025536
MemoryPeak=9950457856
MemorySwapCurrent=0
TasksCurrent=75
ActiveState=active
SubState=running
```

The main Bun process was still the dominant memory holder:

```text
PID      RSS       CMD
3243453  5620664  bun run --cwd packages/synergy --conditions=browser src/index.ts server ...
```

There was also one Codex subprocess at the time, but its RSS was much smaller:

```text
codex vendor process RSS ~= 200 MB
```

After restarting the service, memory immediately returned to a small baseline:

```text
MemoryCurrent=325578752
MemoryPeak=335478784
MemorySwapCurrent=0
ActiveState=active
SubState=running
```

This suggests there is still reachable runtime state or allocator pressure accumulated during long-lived production use.

## What #150 Fixed

The upstream #150 fix appears to cover the major persisted-data sources:

- session list no longer loads message parts;
- completed tool output is canonical-preview bounded;
- full before/after file diffs are migrated to lightweight `FileDiff`;
- workspace status avoids unbounded untracked file reads;
- snapshot capture excludes `.synergy`, dependency/cache/build/binary/media/database/archive files and files over 2 MB.

Those are useful and should remain.

## Remaining Suspected Sources

The remaining high-memory behavior appears tied to runtime lifecycle rather than persisted snapshot/history bloat:

1. **Cortex task state remains heavy for too long**
   - Current dev still keeps terminal task cleanup on a long delay.
   - Completed/cancelled tasks can retain prompt/result/error/progress/recent tool data while the main process continues running.
   - Child session runtimes remain registered until delayed cleanup.

2. **Idle session runtimes remain resident for production workflow cadence**
   - Current dev keeps idle runtime sweep interval around 5 minutes and idle TTL around 30 minutes.
   - In repeated subagent rounds, that is long enough for parent/child runtime state to overlap with the next round.

3. **External-agent tool output can still bypass normal native tool truncation**
   - `ExternalAgentProcessor` writes completed tool output directly from bridge events.
   - Large shell output from Codex-like agents should use the same truncation/artifact policy as native tools.

4. **Cancelled bash tools can remain logically running if the process is gone but no exit path settles**
   - Previously observed: real child process disappeared but a bash tool part/session stayed marked running.
   - That can keep runtime state reachable.

5. **Browser/Playwright sessions can stay alive after UI websocket disconnect**
   - A prior local observation found headless browser processes under the Synergy cgroup while no Cortex tasks were active.
   - Browser runtime should release owner sessions after no observers remain.

6. **Need instrumentation for the remaining phase attribution**
   - We need per-invoke memory logs around prompt build, budget decision, processor completion, and runtime release to distinguish prompt assembly from model/tool/runtime retention.

## Local Mitigation Re-applied

Because the production machine was again near the memory redline, I re-applied a small local mitigation patch on top of current `origin/dev`.

The patch does not replace #150. It adds lifecycle bounds for runtime-state retention:

- compact Cortex task prompt/result/error/progress on terminal status;
- unregister child session runtime quickly after terminal Cortex status;
- shorten idle runtime sweep/TTL for production use;
- trigger debounced Bun GC after runtime/task release;
- settle aborted bash tools after a grace period if no child exit event arrives;
- truncate external-agent completed tool output via the existing tool-output artifact path;
- dispose browser sessions/Playwright owner contexts after websocket observers disconnect;
- add invoke memory instrumentation around prompt build / budget decision / processor completion / release.

Focused validation passed locally:

```text
bun test test/session/manager.test.ts test/cortex/manager.test.ts test/external-agent/codex.test.ts test/browser/tool-navigation.test.ts

60 pass
0 fail
```

Import check also passed:

```text
bun --conditions=browser -e "import './src/session/invoke.ts'; import './src/cortex/manager.ts'; import './src/external-agent/processor.ts'; import './src/browser/runtime.ts'; import './src/server/browser-route.ts'; console.log('import ok')"
```

## Suggested Upstream Follow-up

Please consider a follow-up to #150 focused specifically on runtime lifecycle:

1. Add terminal Cortex task memory lifecycle tests:
   - terminal task clears/compacts prompt/result/error/progress/recentTools;
   - child runtime unregisters promptly;
   - `task_output()` still returns useful summary/tail/full behavior.

2. Add session runtime TTL/backpressure policy suitable for production parallelism:
   - separate long-lived user sessions from completed child task sessions;
   - keep UI semantics but avoid retaining child runtime state for 30 minutes.

3. Route external-agent tool output through the same truncation/artifact path as native tools.

4. Settle cancelled/aborted bash tools even when child process exit events are missed.

5. Add browser runtime idle cleanup after no observers remain.

6. Keep first-class memory instrumentation or expose a debug endpoint:
   - cgroup memory;
   - RSS/heap/external/arrayBuffers;
   - runtime counts;
   - running/queued/completed Cortex task counts;
   - prompt/message/tool-output character counts at invoke phases.

## Why This Is Separate From #150

#150 bounded large persisted session/history/snapshot data, which was necessary.

This follow-up is about long-lived runtime objects that remain reachable after work has completed, been cancelled, or lost its real child process. The observed restart behavior strongly suggests the memory is still tied to main-process runtime retention rather than only persisted data size.

yzxoi:
## Summary
Process output was retained up to 200K chars in the running registry and persisted for 30 minutes after completion. Reduced the per-process output cap to 64K chars and increased tail retention to 8K chars for better diagnostics while reducing memory pressure.

## What changed
- `process/registry.ts`: `MAX_OUTPUT_CHARS` from 200,000 → 64,000; `TAIL_CHARS` from 2,000 → 8,000

## Verification
- Text diff review confirms constants-only change

Fixes #241

### DIFF
--- a/packages/synergy/src/process/registry.ts
+++ b/packages/synergy/src/process/registry.ts
@@ -7,8 +7,8 @@ import { PerformanceMetrics } from "@/performance/metrics"
 const log = Log.create({ service: "process.registry" })
 
 const DEFAULT_TTL_MS = 30 * 60 * 1000 // 30 minutes
-const MAX_OUTPUT_CHARS = 200_000
-const TAIL_CHARS = 2000
+const MAX_OUTPUT_CHARS = 64_000
+const TAIL_CHARS = 8000
 
 export namespace ProcessRegistry {
   export type Status = "running" | "completed" | "failed" | "killed"


yzxoi:
Closing in favor of a more comprehensive fix. This PR's constant changes alone are insufficient for #241 — the real memory retention sources are Cortex task state lifecycle, session runtime TTL, external agent tool output truncation, and browser session cleanup.

PR #474, opened 2026-07-08
Fix worktree transition progress UI

EricSanchezok:
## Summary
- Replaces the full-session worktree transition overlay with centered confirmation dialogs plus bounded in-session progress cards.
- Splits new-session worktree startup so the session is created/navigated first, then worktree setup and first prompt progress render inside the session content.
- Adds the generated `worktree.enter` API/SDK route for binding an existing worktree, removes the obsolete global transition signal and modal start-progress path, and consolidates request error copy.

## Verification
- `bun test src/components/session/worktree-session.test.ts` from `packages/app`
- `bun test src/components/prompt-input/submit-intent.test.ts` from `packages/app`
- `bun run typecheck`
- `bun run quality:quick`

## Notes
- Manual browser verification was not performed because the running app/server must not be restarted; the work was verified through app build/typecheck and focused model/submit tests.


EricSanchezok:
Implemented the follow-up hardening pass for PR #474 and rebased it onto the current `dev`.

Key fixes:
- Added an immediate new-session submit lock covering session create, worktree setup, workflow setup, and first prompt dispatch.
- Restored rollback semantics for worktree new sessions when setup/workflow/first shell/slash/prompt dispatch fails, so we do not leave a usable empty main-checkout session behind.
- Treated `setupFailed` worktree create responses as failures for both new-session and existing-session transitions.
- Split progress data from UI actions and simplified existing-session enter progress to a truthful single request step.
- Hardened existing-session transition re-entry by session ID and forced session sync after enter/leave success.
- Cleaned up test/quality fallout found during the full pass: legacy note index normalization, stale runtime status handling, Windows shell/test portability, and Synergy Link state write retry on ENOENT.

Verification:
- `cd packages/app && bun test src/components/session/worktree-session.test.ts src/components/prompt-input/new-session-submit-lock.test.ts src/components/prompt-input/submit-intent.test.ts`
- `cd packages/synergy && bun test test/project/worktree.test.ts`
- `cd packages/synergy && bun test test/tool/bash.test.ts --bail=1`
- `cd packages/synergy && bun test test/tool/bash-github-token.test.ts`
- `bun run quality:quick`
- `bun run quality`
- `git diff --check` and bidi-control scan

GitHub CI for `0fcab510` is green across Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test.

yzxoi (approved):
## ✅ Approved with follow-up fix

Reviewed the full diff — the approach of moving worktree transition from full-screen overlay into in-session progress cards is sound and well-executed. The `NewSessionSubmitLock`, error rollback semantics, and three-phase progress UI are all solid.

### Fixes pushed to this branch (`5a90c7f6`):

1. **Success progress misses create/bind distinction** — `createNewSessionWorkspaceSuccessProgress` now accepts `selection` so the workspace step label matches: "Create checkout" for `mode: "create"`, "Bind worktree" for `mode: "existing"`.
2. **Mutable `client` scope** — replaced the void `useSessionScopeClient` side-effect with `resolveSessionClient()` returning a new client explicitly.
3. **Duplicate `releaseNewSessionSubmit`** — `.catch` returns `null` sentinel and we early-return, avoiding the redundant second release on the fall-through path.
4. **Missing session existence guard** — `worktree.enter` route now checks `Session.get` before `assertWorktreeSessionIdle`, returning a proper 404 on missing session.
5. **Card position** — moved `WorktreeTransitionCard` from message-list bottom to top (above timeline).
6. **Animation** — added enter (240ms slide+fade in) and dismiss exit (180ms slide+fade out) with `prefers-reduced-motion` override.

PR #475, opened 2026-07-08
fix(session): bound turn memory pressure

yzxoi:
## What this PR for?
Issue #473 reports that high-concurrency sessions still drive Bun/JSC Gigacage to MemoryHigh after processor disposal. Issue #467 narrows the remaining hotspot to `SessionProcessor.process()` tool streaming and settlement bookkeeping. This PR adds turn-level cleanup and pressure handling so each model/tool turn releases processor bookkeeping, records finer memory phase probes, and can force GC under critical RSS/ArrayBuffer/cgroup pressure before the next turn starts.

Fixes #473.
Fixes #467.
Related: #457.

## Summary
- Add `SessionMemoryPressure` with normal throttled GC, critical forced GC, Linux cgroup memory awareness, and compact memory probes.
- Dispose `SessionProcessor` tool bookkeeping after processing, clear stream-local text/reasoning references, and reduce hot-path settlement snapshot allocation.
- Run memory-pressure cleanup after post-turn jobs and document the GC tuning environment variables.

## Test
- [x] lint 通过 (`bun run lint`)
- [x] format 通过 (`bun run format:check`)
- [x] typecheck 通过 (`bun run typecheck`)
- [x] focused tests: `bun test test/session/processor.test.ts test/session/memory-pressure.test.ts test/session/invoke.test.ts`
- [x] pre-push hook 通过

## Before/After
**Before:** Session turns only exposed coarse `processor.process` memory growth, retained processor maps until the processor object became collectible, and normal GC could remain throttled during critical MemoryHigh pressure.
**After:** Processor-local state is explicitly cleared at turn completion, phase probes split the processor hot path, and critical memory pressure bypasses the normal GC interval.

Issue #400, opened 2026-07-07
Synergy dev service memory climbs to cgroup limit and retains long-lived tool child processes


marzich:
## Summary

On a production Synergy dev service, memory again climbed to the configured cgroup limit after several hours. This appears to involve both high main-process RSS and long-lived tool child processes remaining inside the `synergy-dev.service` cgroup.

## Environment

- Branch: `dev`
- HEAD: `e9c26276` (`origin/dev`)
- Service: `synergy-dev.service`
- Host bind: `100.117.3.114:18080`
- Guardrails: `MemoryHigh=10G`, `MemoryMax=11G`, `MemorySwapMax=0`, `TasksMax=1024`

## Initial local evidence

Captured at `2026-07-07 17:21 CST`:

- System memory: `15Gi total`, `12Gi used`, `195Mi free`, `2.9Gi available`
- System swap: `4.0Gi total`, `2.7Gi used`
- `synergy-dev.service` active for about `3h50m`
- systemd cgroup memory: `10.2G`, high `10.0G`, max `11.0G`, swap current `0`, available `0B`, peak `10.3G`
- Main Synergy `bun` process RSS: about `8.1G` (`8481700 KiB`)
- Service tasks: `259 / 1024`

The Synergy service cgroup also contained long-lived child processes launched through tool workflows, including:

- `uvicorn holos.main:app --host 0.0.0.0 --port 8000`
- `next dev -p 8090 -H 0.0.0.0`
- `clarus-agent-adapter ... daemon`
- several `oxlint --lsp` processes
- lingering `rg` searches and `python3 -` helper processes

Example service tree shape:

```text
bun run --cwd packages/synergy ... server --hostname 100.117.3.114 --port 18080
  |- node ... oxlint --lsp
  |- node ... oxlint --lsp
  |- node ... oxlint --lsp
  |- uv run uvicorn holos.main:app ...
  |- next dev -p 8090 -H 0.0.0.0
  |- clarus-agent-adapter ... daemon
  |- python3 -
  `- rg ...
```

## Why this matters

This machine runs Synergy as the main production workflow. When the service reaches `MemoryHigh`, reclaim pressure makes the UI and tool execution slow. Once it approaches `MemoryMax`, user retries and active sessions can compound the problem. The service has `MemorySwapMax=0`, but the host can still be under global memory pressure from other processes.

## Initial hypothesis

There may be two separate retention paths:

[...]

PR #476, opened 2026-07-08
fix(bash): block detached daemon launches

### LINKED ISSUE #432: Production Synergy stalls under MemoryHigh from main-process ArrayBuffers plus detached workflow daemons
## Summary

Production Synergy on the Huawei Ubuntu machine entered another severe MemoryHigh reclaim stall after about 4h22m uptime. The service was not OOM-killed, but the backend became effectively stuck: `/global/performance/summary` timed out and cgroup memory pressure was very high.

This is a follow-up to the previous memory pressure issue, but this incident has clearer cgroup-wide attribution.

## Incident timestamp

- Local time: `2026-07-08 00:53 CST`
- Service uptime: about `4h22m`
- Main PID: `3952461`

## Evidence

Systemd/cgroup state:

- `ActiveState=active`, `SubState=running`
- `NRestarts=0`
- `MemoryCurrent=12268158976`
- `MemoryPeak=12326481920`
- `MemoryHigh=11811160064`
- `MemoryMax=13958643712`
- `MemorySwapCurrent=0`
- `oom=0`, `oom_kill=0`, `oom_group_kill=0`
- `memory.events high=422478`
- `memory.pressure full avg10=69.17`, `full avg60=69.72`

Backend symptom:

- `GET /global/performance/summary` timed out after 8 seconds.

High-memory snapshot:

- Main process RSS: `9279549440`
- Main process heapUsed: `8849918883`
- Main process external: `8641993613`
- Main process arrayBuffers: `8426239416`
- `sessionMessageCache.totalBytes=0`, so the session message cache was not the immediate driver.
- `trackedToolProcesses` contained two long-running `rg` processes that had auto-backgrounded and not exited.

Cgroup-wide process attribution:

- `cgroupProcesses.count=29`
- `cgroupProcesses.rssBytes=12309757952`
- Top cgroup processes:
  - Synergy main process: `~9.28GB`
  - Holos Python child: `~1.87GB`
  - Next dev server: `~564MB`
  - Clarus adapter: `~176MB`, in `D` state at `mem_cgroup_handle_over_high`
  - Clarus worker: `~120MB`, in `D` state at `mem_cgroup_handle_over_high`

Detached daemon pattern found in Synergy cgroup:

- `tmux new-session -d -s clarus-rcm ...`
- `uvicorn ... --reload`
- `npm run dev / next dev`
- `clarus_agent_adapter ... daemon`
- `clarus-worker`

These were still charged to `synergy-dev.service`, so they competed with the Synergy main process for `MemoryHigh`/`MemoryMax`.

## Current hypothesis

This incident is a combination of:

1. Main-process retention of large ArrayBuffer/external memory during/after heavy sessions.
2. Background experience encoding and related model/embedding work still running on large turns, increasing memory and latency pressure.
3. Detached workflow daemons started from bash/tmux staying inside `synergy-dev.service` cgroup.
4. Long-running shell commands auto-backgrounding and remaining tracked but unfinished.

## Expected behavior

- Synergy should not let detached daemons silently consume the Synergy service cgroup by default.
- Library experience encoding should not run expensive intent/script/reward/embedding work while process memory is already above a production safety threshold.
- High-memory snapshots should make cgroup-wide ownership obvious; the new snapshot attribution confirms this is now possible.

## Immediate local remediation plan

- Add a hard default guard in the bash tool to block detached daemon launch patterns (`tmux new-session -d`, `nohup`, `setsid`, `disown`, shell backgrounding) unless explicitly allowed.
- Add a memory pressure guard to the library experience encoder so it skips/defer encoding when Synergy is already above safe RSS/ArrayBuffer thresholds.
- Restart the production service after tests so the cgroup is cleared and patches load.

yzxoi:
## What this PR for?
Issues #432 and #400 report that detached workflow daemons and shell-backgrounded processes remain charged to the Synergy service cgroup after the initiating bash tool call appears settled. This PR adds a default guard in local bash execution so shell-level detached daemon patterns are blocked unless an operator explicitly enables them for that runtime.

Fixes #432. It also covers the detached daemon / child-process retention path discussed in #400; #479 handles the remaining child-process attribution and stale-process diagnostics for #400. Complements #475, which handles the main-process turn memory-pressure path from the same incident family.

## Summary
- Block local bash commands that use `tmux new-session -d`, `nohup`, `setsid`, `disown`, `daemonize`, or shell backgrounding with `&` by default.
- Allow explicit operator-managed override with `SYNERGY_BASH_ALLOW_DETACHED_DAEMONS=1` or internal `shellAllowDetachedDaemons` context.
- Update bash tool guidance to prefer tracked `background` / `yieldSeconds` process lifecycle instead of detached shell daemons.

## Test
- [x] lint 通过 (`bun run lint`)
- [x] format 通过 (`bun run format:check`)
- [x] typecheck 通过 (`bun run typecheck`)
- [x] focused test: `bun test test/tool/bash.test.ts`
- [x] pre-push hook 通过

## Before/After
**Before:** A bash command could start detached daemons that stayed in the Synergy service cgroup without a tracked process lifecycle.
**After:** Detached daemon launch patterns are denied by default, and agents are directed to keep long-running work tracked through the process tool.


EricSanchezok:
Audit complete for #476.

What this PR was trying to fix:
- #432 is real: production hit severe cgroup memory pressure, with both Synergy main-process memory retention and service/workflow child processes contributing to the same `synergy-dev.service` cgroup pressure.
- The detached daemon part of the incident is also real: processes started from shell/tmux/nohup can remain charged to the Synergy service cgroup after the bash tool call appears settled.

Problem found in the original PR:
- The original branch turned that observation into a default runtime block for `tmux new-session -d`, `nohup`, `setsid`, `disown`, `daemonize`, and shell `&`.
- That is not the correct root-cause fix for #432. The issue discussion explicitly records that this guard was tested locally and reverted because it is not acceptable to prevent users/operators from starting required Holos/Next/adapter style workflow services from Synergy.
- The implementation also added a brittle shell-string regex classifier in the main bash path. It cannot reliably parse shell quoting/contexts, and it would have introduced a new compatibility/escape-hatch path (`SYNERGY_BASH_ALLOW_DETACHED_DAEMONS` / `shellAllowDetachedDaemons`) rather than a clean lifecycle model.

Fix I pushed:
- Removed the runtime detached-daemon detector, block, env override, internal context override, and tests that locked in the block.
- Kept the useful guidance in `bash.txt`, but changed it to: prefer tracked `background` / `yieldSeconds` when Synergy should own the process lifecycle; do not use detached daemons casually; only launch service-style detached processes when the user/operator explicitly wants that lifecycle and has a monitor/stop plan.
- Merged current `origin/dev` into the branch and resolved the bash test conflict by keeping the new `Shell` cache reset from `dev` plus the existing `LocalBashBackend` import.

Resulting assessment:
- This branch no longer claims a misleading runtime fix for #432. It now avoids adding dead/legacy/compatibility logic to bash execution and keeps the product behavior aligned with the actual incident follow-up.
- The real upstream fix direction for #432 remains separate: main-process retention fixes (hashline/session runtime, encoder pressure guard, etc.) and eventually an explicit safe service-launch path that can put operator-managed daemons outside the Synergy service cgroup. A regex block in bash is not that path.

Verification run locally:
- `cd packages/synergy; bun test test/tool/bash.test.ts` (29 pass)
- `bun run format:check`
- `bun run typecheck`
- `git diff --check`
- `bun run quality:quick`
- pre-push hook: format, lint, typecheck, monorepo check

Remote CI on head `c6943f218d3916189fbf2704e229cae0ee44ccfa` is fully green: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

yzxoi:
## 测试结果总览

### 拦截场景 ✅

| 模式 | 命令 | 结果 |
|---|---|---|
| `nohup` | `nohup sleep 10 &` | ✅ Blocked — 拦截成功 |
| `tmux new-session -d` | `tmux new-session -d -s test` | ✅ Blocked — 拦截成功 |
| `screen -dm` | `screen -dmS test` | ✅ Blocked — 拦截成功 |
| `setsid` | `setsid sleep 5` | ✅ Blocked — 拦截成功 |
| `disown` | `sleep 1 & disown` | ✅ Blocked — 拦截成功 |
| `&` (shell background) | `python3 -m http.server 0 ... &` | ✅ Blocked — 拦截成功 |

### 放行场景 ✅

| 场景 | 命令 | 结果 |
|---|---|---|
| 正常前台命令 | `echo "hello"` | ✅ Pass — 正常执行 |
| tracked background | `background:true, yieldSeconds:2` | ✅ Pass — 正常执行 |
| `docker run -d` | `docker run -d --name test nginx` | ⚠️ 通过 — Docker 自身的 daemon 模式未被拦截（预期内，不在检测范围） |

### 错误消息一致性 ✅

所有拦截返回统一的错误消息格式：

```
Error: Blocked detached daemon launch pattern: <pattern>
Detached shell daemons stay inside the Synergy service cgroup and can keep consuming MemoryHigh/MemoryMax after the tool call appears settled.
Use the bash tool's background/yieldSeconds flow for tracked processes, or set SYNERGY_BASH_ALLOW_DETACHED_DAEMONS=1 only for an operator-managed runtime that intentionally permits detached daemons.
```

### CI 状态

- `Test` job 的 `screen -dmS` regex 已在 CI 失败后修复（`\b` → `(?:\w|$)`），重新 push 后会重新触发

yzxoi (commented):
LGTM. 测试覆盖完整，所有 7 种 daemon 模式全部拦截成功，放行场景正常。CI regex 问题已修复。建议关注  场景——当前不在检测范围内，如果生产中有 Docker daemon 残留问题可后续迭代。

yzxoi (commented):
LGTM. 测试覆盖完整，所有 7 种 daemon 模式全部拦截成功，放行场景正常。CI regex 问题已修复。

PR #478, opened 2026-07-08
fix(session): reduce prompt budgeter peak allocations

### LINKED ISSUE #457: High RSS/ArrayBuffer growth during a single long tool-calling session
## Summary

A production Synergy daemon can grow to 6-10 GiB RSS even when only one parent session is visibly active. The persisted session data is only a few MiB, and the in-memory session message cache/hashline snapshot stores are also only MiB-scale. The growth appears to be in Bun/JavaScriptCore external/ArrayBuffer/native allocation while repeatedly rebuilding and estimating the prompt for a long tool-calling loop.

This is a follow-up to the prior memory/OOM work: the issue is improved but still reproducible under a long-running single-session workflow.

## Environment

- OS: Ubuntu on Huawei laptop
- Synergy branch: `dev`
- Local HEAD at observation time: `e9c26276`
- Service mode: user systemd service, Bun backend
- Observation time: 2026-07-08 18:10-18:20 CST
- Main active parent session: `ses_0c43714f1ffeZOdFTsVmI3vFtJ`
- Completed child session: `ses_0bec8c2e9ffdZ3zYdcBiui1g7I`
- Model/provider in the hot loop: `openai-codex/gpt-5.5`

### 1. Main process memory is anonymous JSC/Bun memory, not file cache or swap

At high memory pressure:

```text
Synergy MainPID: 1079866
VmRSS:      ~6.1-6.5 GiB
RssAnon:    ~6.0-6.4 GiB
VmSwap:     0
cgroup memory.current: ~8.9-9.8 GiB
memory.events: low=0 high=0 max=0 oom=0 oom_kill=0
```

`/proc/$PID/smaps` grouped by mapping showed:

```text
~4.7 GiB  [anon:JSGigacage]
~0.9 GiB  [anon:WKFastMalloc]
~0.3 GiB  [anon]
```

So the dominant memory is Bun/JavaScriptCore heap/ArrayBuffer/native allocator memory, not SQLite mmap, logs, file cache, or swap.

### 2. `process.memoryUsage()` shows external/ArrayBuffer growth

Recent high-memory snapshots showed the pressure comes from `external` / `arrayBuffers`:

```text
2026-07-08T09:58:49Z  rss 3.45 GiB, heapUsed 2.91 GiB, external 2.74 GiB, arrayBuffers 2.49 GiB
2026-07-08T10:03:49Z  rss 3.83 GiB, heapUsed 3.25 GiB, external 2.98 GiB, arrayBuffers 2.76 GiB
2026-07-08T10:08:49Z  rss 6.41 GiB, heapUsed 5.91 GiB, external 5.63 GiB, arrayBuffers 5.24 GiB
2026-07-08T10:18:49Z  rss 5.59 GiB, heapUsed 4.98 GiB, external 4.70 GiB, arrayBuffers 4.44 GiB
```

This pattern matches large temporary buffers or tokenization/provider stream allocations that are not reclaimed promptly.

### 3. Persisted session data is small

The active parent session was only ~5 MiB on disk:

```text
ses_0c43714f1ffeZOdFTsVmI3vFtJ: 5.0M on disk, 708 message/part JSON files
largest part JSON: ~115 KiB
text parts: 120
reasoning parts: 77
tool parts: bash 67, view_file 43, scan_files 15, task 9, task_output 7, revise_file 6
```

Aggregated persisted content for that session was also small:

```text
total text chars:       ~57 KiB
total tool output chars: ~465 KiB
total tool input chars:  ~104 KiB
```

The completed child session was only ~664 KiB on disk.

### 4. Existing runtime caches are not the root cause

High-memory snapshot values:

```text
sessionMessageCache.totalBytes: 0-1.9 MiB
hashlineSnapshotStore.totalBytes: 2-4.5 MiB
```

So the issue is not raw retained session metadata, `SessionMessageCache`, or hashline snapshots.

### 5. The parent session was still actively looping

Although only one visible workflow remained, the parent session was still running with `pendingReply=true` and repeatedly alternating tools and LLM calls. Recent traces around 18:17-18:20 CST showed repeated `session.turn` / `llm.request` spans for the same session:

```text
ses_0c437... llm.request 25.8s openai-codex/gpt-5.5
ses_0c437... llm.request 19.9s
ses_0c437... llm.request 12.9s
ses_0c437... llm.request 16.7s
ses_0c437... llm.request 22.8s
```

These were interleaved with small `bash`, `view_file`, and `scan_files` tool calls.

### 6. Hot code path likely amplifies memory per step

The loop appears to rebuild and re-estimate the whole prompt every step:

- `session/invoke.ts`: copies/project session messages each turn
- `MessageV2.toModelMessage(...)`: converts full history to AI SDK messages
- `PromptBudgeter.buildPlan(...)`: transforms messages
- `PromptBudgeter.decide(...)`: estimates prompt size
- `PromptBudgeter.estimateModelJSONCached(...)`: calls `JSON.stringify(value)` for cache key and again for token estimation
- `Token.estimateModel(...)`: uses `js-tiktoken` `tokenizer.encode(...)`
- `LLM.stream(...)`: passes `system + messages` to AI SDK `streamText(...)`

In a long tool-calling loop, this can repeatedly allocate large JSON strings, token arrays, provider message arrays, and stream buffers. The allocations show up as JSC `JSGigacage` / `WKFastMalloc` / `arrayBuffers`.

### 7. GC currently runs too late for long loops

`maybeCollectAfterSession()` is currently called from the deferred cleanup at the end of `SessionInvoke.loop()`. For a session that keeps continuing through tool calls, this means memory may grow for many turns before the GC probe runs.

## Secondary Issue: performance SQLite file bloat

The performance SQLite file had grown to 1.3 GiB even though the configured max is 250 MiB:

```text
~/.synergy/state/observability/performance/performance.sqlite: 1.3G
PRAGMA page_size: 4096
PRAGMA page_count: 321974
PRAGMA freelist_count: 315924
live bytes: ~24.8 MiB
free bytes: ~1.29 GiB
```

So retention/max-size enforcement deletes rows but does not shrink the main DB file. This is not the main process RSS root cause, but it can contribute to service cgroup/page-cache pressure and makes memory pressure harder to interpret.

## Root Cause Hypothesis

Strong evidence:

- High memory is in Bun/JSC anonymous memory, especially `JSGigacage` and `WKFastMalloc`.
- `process.memoryUsage()` reports multi-GiB `external`/`arrayBuffers`.
- Persisted session data, message cache, and hashline snapshots are all MiB-scale.
- The active parent session was still in a high-frequency tool-calling loop.

Likely cause:

- Repeated full-history prompt projection + JSON stringify + tokenization + AI SDK Responses streaming creates large transient allocations every step.
- Because the loop continues, after-session GC does not run until too late.
- Some JSC/Bun allocator memory is not returned quickly to the OS after each turn, so RSS remains high and can climb across turns.

### P0: Add per-turn memory pressure cleanup

Run a cheap memory pressure check after each `processor.process(...)` turn and before continuing to the next tool/LLM step. If RSS or `arrayBuffers` exceeds thresholds, clear turn-local references and call `Bun.gc(false)` or a configurable stronger GC mode.

This should not wait until the entire session loop exits.

### P0: Avoid repeated `JSON.stringify + tiktoken.encode` for every message every step

`PromptBudgeter.estimateModelJSONCached()` currently stringifies values for the cache key and token estimation. For long sessions this can allocate a lot even when values are unchanged.

Possible approaches:

- Cache token estimates by persisted message/part ids plus update timestamp/hash.
- Use cheap persisted size/hash metadata for unchanged messages.
- Estimate only changed tail messages after a calibrated API token count exists.
- Avoid `Promise.all(messages.map(...))` over the whole history when only the newest turn changed.

### P1: Cap or compact historical tool output before provider prompt projection

The active session had many tool parts, but only MiB-scale persisted content. Even so, the full projection of historical tool outputs is repeatedly sent through prompt construction and token estimation.

Options:

- Summarize or compact old tool outputs after N turns.
- Keep full outputs on disk/UI, but feed bounded summaries into model context.
- Prefer turn digest / result references for old completed tools.

### P1: Instrument prompt build memory deltas

Add per-step metrics around:

- `Session.messages`
- `MessageV2.toModelMessage`
- `PromptBudgeter.buildPlan`
- `PromptBudgeter.decide`
- `Token.estimateModelJSON`
- `streamText` initialization

Record RSS/heap/external/arrayBuffers deltas so the exact growth point is visible in production snapshots.

### P1: SQLite maintenance

After retention/max-size enforcement, run a bounded shrink strategy for performance SQLite:

- incremental vacuum if enabled, or
- periodic `VACUUM` under a size/free-page threshold and when the system is idle, or
- rotate/recreate the performance DB when freelist dominates the file.

## Expected Outcome

A single long-running tool-calling session should not push the daemon to 6-10 GiB RSS when its persisted data is only a few MiB. Memory should either remain bounded per turn or drop after each turn, without requiring a full service restart.

yzxoi:
## What this PR for?
Fixes #457. Long tool-calling sessions repeatedly rebuilt prompt budget estimates by stringifying and tokenizing every historical message in parallel. That creates avoidable peak JSON/token buffer allocation even when estimates are cached.

## Summary
- Estimate history messages serially instead of `Promise.all` over the whole prompt history.
- Serialize each estimate value once and reuse that string for both the estimate cache key and tokenization.
- Add regression coverage for cache reuse with serialized values and bounded estimator concurrency.

## Test
- [x] lint 通过
- [x] format 通过
- [x] typecheck 通过
- [x] `bun test test/session/prompt-budgeter.test.ts test/session/invoke-compaction.test.ts test/session/compaction.test.ts`
- [x] pre-push hook 通过

## Before/After
**Before:** Prompt budgeting could concurrently retain JSON strings/token buffers for every message in a long history and stringify non-string estimate values twice.
**After:** Prompt history estimates are processed one at a time, and each value is serialized once before cache lookup/tokenization.


EricSanchezok:
Audit complete for this PR.

What this PR is fixing:
- The issue is real. The production evidence in #457 points to high RSS/external/ArrayBuffer pressure during a long single-session tool loop, while persisted session data and explicit runtime caches remain MiB-scale. Repeated prompt rebuild and token estimation is a plausible hot allocation path.

Root-cause assessment:
- This PR correctly targets one concrete contributor: prompt budgeting estimated every historical message concurrently and stringified non-string values once for the cache key and again inside token estimation.
- This is not a complete fix for all of #457. It does not add per-turn GC, avoid full-history provider projection, summarize old tool output, or shrink performance SQLite files. It is a focused reduction of peak transient allocations in the prompt budgeter.

Patch-quality assessment:
- The change preserves estimator semantics. `Token.estimateModelJSON()` already tokenizes strings as-is and non-strings through `JSON.stringify(value)`; the PR just serializes once before cache lookup and tokenization.
- Serializing before the cache key also avoids duplicate large JSON strings without adding a parallel cache model.
- Processing history messages serially is a deliberate memory/latency tradeoff. It can reduce parallel tokenization throughput, but prompt budgeting is on the memory-pressure path here, and tool/schema estimates remain parallel where the fanout is bounded by tool count.
- I do not see redundant compatibility, wrapper, legacy, or dead-code paths introduced. The budgeter control flow remains straightforward.

Additional work I did:
- Merged current `origin/dev` into the branch in `124d0e96cdb33fd91ce2b4ebdf9fee8e562d5bac` so the branch uses the current SDK/package-check and kvcache baseline. No additional code changes were needed after audit.

Verification:
- `bun test test/session/prompt-budgeter.test.ts test/session/invoke-compaction.test.ts test/session/compaction.test.ts test/session/kvcache-measurement.test.ts` from `packages/synergy`
- `bun run quality:quick`
- pre-push hook on `git push origin synergy/fix-issue-457`
- GitHub CI on HEAD `124d0e96cdb33fd91ce2b4ebdf9fee8e562d5bac`: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

PR #479, opened 2026-07-08
fix(performance): attribute tool child process memory

### LINKED ISSUE #400: Synergy dev service memory climbs to cgroup limit and retains long-lived tool child processes

yzxoi:
## What this PR for?

Fixes #400. The issue reports Synergy reaching the service cgroup memory limit with both high server RSS and long-lived tool child processes contributing to MemoryHigh pressure. This PR adds first-class attribution for registered tool child process RSS, keeps server resource samples separate from child process samples, and settles stale process registry entries when their pid has disappeared.

## Summary

- Add ProcessRegistry child process inspection, stale-pid settling, and Linux `/proc/<pid>/status` RSS sampling for registered tool processes.
- Record child process resource samples and expose aggregate child RSS plus top child memory contributors in the Performance summary/UI.
- Keep server resource charts/cards bound to server samples so child process samples do not corrupt main RSS, CPU, or event-loop metrics.

## Test

- [x] lint 通过
- [x] format 通过
- [x] typecheck 通过
- [x] `bun test test/process/registry.test.ts test/performance/store.test.ts test/server/performance-route.test.ts`
- [x] `bun test src/components/performance/PerformanceDashboard.test.ts`

## Before/After

**Before:** Performance summary only exposed the main server process RSS and could not identify registered tool child process memory contributors; stale registered pids could continue to appear active.
**After:** Performance summary separates server resources from tool child resources, shows child process RSS contributors, and settles missing child pids into finished process history.


EricSanchezok:
Audit complete for this PR.

What this PR is fixing:
- The issue is real. The production report had two separate contributors to service-level memory pressure: main Synergy RSS and long-lived tool child processes still inside the service cgroup. Without first-class child process attribution, the Performance page could show server RSS but could not explain which registered tool children were consuming memory.

Root-cause assessment:
- The PR addresses the observability/root-cause gap rather than just adding another memory card. It keeps server resource samples and tool-child resource samples separate, uses the existing `ProcessRegistry` as the ownership source for child processes, and settles registered pids that have disappeared before presenting them as active.
- This does not yet solve every possible retention path from #400, especially main-process heap/native retention or lifecycle policy for intentionally persistent user-managed child processes. It does give the next production report the attribution needed to distinguish those paths.

Patch-quality assessment:
- The main-line data model is clear: server samples keep `process.role = server`; child samples use `tool` / `tool-background`; dashboard server RSS/event-loop/CPU reads filter back to server rows; child RSS appears as support diagnostics and top contributors.
- Linux RSS sampling is correctly best-effort through `/proc/<pid>/status`; unsupported hosts still expose active registered child counts without pretending to have RSS.
- Stale process settlement reuses `markExited`, so the registry moves missing pids into the finished path instead of adding a parallel lifecycle model.
- The UI additions are small and aligned with the Performance product contract. I do not see wrapper/legacy/dead-code paths added in the dashboard or registry flow.

Issue I found and fixed:
- The original patch changed `perf_resource_samples` by adding `process_id`, `process_role`, and `labels_json`, but `PerformanceStore` still had schema version `1` and relied on `CREATE TABLE IF NOT EXISTS`. Existing production SQLite databases would not get the new columns, so `insertResource()` and role-filtered dashboard queries could fail on upgrade.
- I added commit `576d6f7b094aba3522a07be27175c4505b80a412` with a v2 store migration: it adds missing resource-sample columns, classifies legacy rows as `server`, creates the role/time index only after the column exists, and updates the stored schema version. I also added a regression test that opens a v1-shaped resource table and verifies both legacy reads and new writes.
- I also merged current `origin/dev` into the branch to bring in the current Windows SDK build script baseline; the PR diff against `dev` remains focused on performance observability plus the migration fix.

Verification:
- `bun test test/performance/store.test.ts test/process/registry.test.ts test/server/performance-route.test.ts` from `packages/synergy`
- `bun test src/components/performance/PerformanceDashboard.test.ts` from `packages/app`
- `bun run quality:quick`
- pre-push hook on `git push origin synergy/fix-issue-400`
- GitHub CI on HEAD `576d6f7b094aba3522a07be27175c4505b80a412`: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

PR #480, opened 2026-07-08
fix(session): expose effective control profile on read

### LINKED ISSUE #470: bug: frontend permission mode reverts to 'Guarded' on page refresh, only corrects after session runs
## User report

> @chaos 现在每一次刷新前端，原来显示full的权限会变成gaurd，然而在显示gaurd的状态下开始跑，又会变成full

Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6bd7bbed64acc255151f9cc221c

## Triage

- Type: bug_report
- Related area: area/frontend, area/session

## Deduplication

No duplicate found in recent open/closed issues. Issue #240 (file_search requests permission even with full_access) and #384 (add controlProfile to agenda) are related but not the same.

## Static inspection

- `packages/app/src/components/prompt-input.tsx:934-939` — `selectedControlProfile` memo resolves permission mode: for existing sessions it reads `info()?.controlProfile` (from `sync.session.get`) and falls back to `sync.data.config.controlProfile` (global config default, which is usually undefined → `permissionModeVisual` defaults to "Guarded").
- `packages/synergy/src/session/index.ts:572-581` — `Session.get()` only auto-resolves `controlProfile` for child sessions (parentID exists). For parent sessions, if the stored data lacks `controlProfile`, it stays `undefined`.
- `packages/synergy/src/session/index.ts:672-718` — `Session.list()` reads sessions from storage without resolving `controlProfile` at all.
- `packages/synergy/src/session/index.ts:338` — `create()` stores `controlProfile` only if `input?.controlProfile` is provided; for child sessions (`parentID` set), it's explicitly set to `undefined`.

Likely root cause: On frontend refresh, the session info is loaded (via list or get) but the stored session data may not include `controlProfile`, or there is a timing/race condition where `info()` returns undefined before the sync completes. The memo then falls back to the global config default ("guarded"). Once the session starts executing (WebSocket events publish updated session info), the correct `controlProfile` becomes available and the display flips to "Full".

## Expected behavior

After refreshing the frontend, the permission mode displayed for a session should match the stored session's `controlProfile` value (e.g., "Full Access" if the session was created with `controlProfile: "full_access"`).

## Actual behavior

After refreshing the frontend, the permission mode displays "Guarded" instead of "Full Access" for sessions that have `controlProfile: "full_access"` stored. The display only corrects to "Full" after the session starts executing.

## Proposed fix

1. Ensure `Session.get()` and `Session.list()` resolve `controlProfile` for all sessions (not just child sessions) if it's missing from storage, by calling `resolveControlProfile` or `resolveEffectiveControlProfile`.
2. Alternatively, ensure the frontend `sync.session.sync()` properly waits for the session info to load before computing the permission mode visual, or add a loading state.

yzxoi:
## What this PR for?

Fixes #470. The issue reports that after refreshing the frontend, a session that should display Full Access can temporarily display Guarded until the session runs again. The read API could return stored session info without an effective `controlProfile` when the profile came from config defaults or parent inheritance, so the frontend fell back to its visual default.

## Summary

- Add a client-facing session info helper that resolves effective `controlProfile` for read responses without persisting inherited/config-derived values.
- Use that helper for `Session.get()`, `Session.list()`, search list results, and child session pages.
- Add regression coverage for root config-derived profiles and child inherited profiles in list/get reads.

## Test

- [x] lint 通过
- [x] format 通过
- [x] typecheck 通过
- [x] `bun test test/session/session.test.ts`

## Before/After

**Before:** Refresh-loaded session info could omit `controlProfile`, causing the frontend permission selector to render Guarded.
**After:** Session read APIs expose the effective control profile consistently, so refreshed UI shows the same permission mode the runtime will use.


EricSanchezok:
Audit complete for this PR.

What this PR is fixing:
- The issue is real. Refreshed session reads can surface the stored session record before runtime execution re-resolves the effective permission profile. When a root session has no stored `controlProfile` because it is meant to inherit from top-level config, or when a child session inherits from its parent, the read/list API could return `controlProfile: undefined`; the frontend then renders its visual fallback, typically Guarded.

Root-cause assessment:
- The root cause is at the session read boundary, not just in the frontend selector. Runtime execution already resolves the effective profile through `resolveControlProfile()`, but client-facing read paths were exposing raw persisted state in places where the UI needs the same effective session semantics.
- The fix uses the existing resolver instead of creating a second frontend-only interpretation, which keeps the runtime and read model aligned.

Patch-quality assessment:
- The patch is not just a superficial UI fallback. It adds a narrow `withClientInfo()` helper that composes the existing runtime info shape with the existing effective control-profile resolver.
- It does not persist inherited/config-derived profile values back to session storage; the tests explicitly check that root sessions using config defaults still have no stored session-level profile after `Session.get()`.
- The list/search/child-page read paths now share the same client-facing behavior, so there is no special-case drift between refreshed sidebar lists, direct session reads, search results, and child-session pages.
- I do not see redundant compatibility, wrapper, legacy, or dead-code branches introduced here. The one caveat is that list responses now resolve effective profiles for returned rows, but this happens after pagination/filtering and reuses the existing resolver, so the complexity is contained.

Additional work I did:
- The original CI failure was in `@ericsanchezok/synergy-link#test`, not in this PR's session tests. I merged current `origin/dev` in `78bfb600d79b28ef44f0c5d946bd36ae51a088cf`, which brings in the existing baseline fix for that unrelated async state-write race. The PR diff against `dev` remains focused on `packages/synergy/src/session/index.ts` and `packages/synergy/test/session/session.test.ts`.

Verification:
- `bun test test/session/session.test.ts` from `packages/synergy`
- `bun run test` from `packages/synergy-link` to cover the previously failing CI package
- `bun run quality:quick`
- pre-push hook on `git push origin synergy/fix-issue-470`
- GitHub CI on HEAD `78bfb600d79b28ef44f0c5d946bd36ae51a088cf`: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

PR #481, opened 2026-07-08
fix(frontend): preserve session model variant

### LINKED ISSUE #463: bug: model effort variant resets to Default in UI after sending the first message (new session)
## User report

> 现在初始的model effort选择好之后，发送消息又会自动切换回默认

Reporter: Feishu:ou_29e9348c6b7a9818a0458f24ed03fcfe
Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6bd7d47ee8a4c38c671da733f77

## Triage

- Type: bug_report
- Related area: area/frontend, area/session

## Deduplication

No duplicate found in recent open issues.

PR #301 (merged 2026-07-05, "Fix session-scoped model effort selection") previously added a restore effect in `session.tsx` that copied `msg.variant` to the local variant store when opening a session. However, a subsequent refactoring (composer intent layering for issue #318) removed that restore code and introduced a new `composer-intent.ts` system that handles `model` and `agent` inheritance but does **not** handle `variant`.

Issue #184 ("Add durable model variant/reasoning-effort controls") was a broader feature request for variant persistence and was closed when the per-session variant store was implemented, but the regression described here was not caught.

### Root cause

The composer intent system (`packages/app/src/context/composer-intent.ts`) has three layers for model/agent resolution:
1. `draft` — user's explicit per-session choice
2. `sessionDefault` — server modelOverride or last root message's value
3. `fallback` — agent default / recent / provider default

The `sessionDefaultModel()` function (line 55-61) reads the model from the last root message, and `sessionDefaultAgent()` (line 63-65) reads the agent. However, there is no `sessionDefaultVariant()` function — the `RootMessageLike` type (line 39) does not include a `variant` field, and neither does any derivation.

The variant is separately handled via `variantSession()` in `packages/app/src/context/local.tsx` (line 238), scoped by `(sdk.scopeKey, params.id)`. When a new session is created:

1. User selects a variant on the workspace-level composer → stored under `workspace:<scope>:model-variant` (since `params.id` is `undefined`, `loadVariantSession` uses the `WORKSPACE_KEY` constant)
2. The `selectedVariant` is correctly captured and sent with the first message (`packages/app/src/components/prompt-input/submit.ts`, line 244)
3. After `navigate(...)` to the new session, `params.id` becomes the new session ID → `variantSession()` now loads a **different** store key (`session:<new-id>:model-variant`)
4. The new session store is empty → `current()` returns `undefined` → UI shows "Default"

The previous fix (PR #301) restored variant from the last message via an effect:
```
if (msg.model) {
  local.model.set(msg.model)
  local.model.variant.set(msg.variant, msg.model)
}
```
This effect was removed during the composer-intent refactoring and was not replaced.

### Other affected files

- `packages/app/src/context/local.tsx` — `variantSession` scope switch on `params.id` change is the core mechanism of the bug
- `packages/app/src/components/prompt-input/submit.ts` — correctly captures `selectedVariant` at line 244 and sends it with the message; no step copies workspace variant to session variant after session creation
- `packages/app/src/context/model-variant.ts` — per-session variant persistence store

## Expected behavior

When a user selects a model effort/variant on the new-session composer and sends a message, the variant should be preserved in the UI after the session is created and navigated to. The top-bar should continue to show the selected effort, not "Default".

Similarly, when navigating to an existing session that has previous messages with a variant, the composer should inherit that variant.

## Actual behavior

After selecting a model effort on the initial composer and sending the first message, the top-bar variant indicator resets to "Default". The variant is still correctly sent with the message to the backend (the LLM call uses the right effort), but the frontend UI does not reflect it.

## Proposed fix

Two complementary approaches — either one alone may suffice, but both together provide robustness:

### Approach A: Migrate workspace variant to session on session creation

In `packages/app/src/components/prompt-input/submit.ts`, after a new session is created and before/after navigation, copy the workspace-level selected variant to the new session's variant store:

```typescript
// After session creation and navigation (around line 342)
if (selectedVariant && session) {
  local.model.variant.set(selectedVariant, {
    providerID: currentModel.id,
    modelID: currentModel.provider.id,
  })
}
```

### Approach B: Add variant to the composer-intent inheritance system

In `packages/app/src/context/composer-intent.ts`:
1. Add `variant?: string` to the `RootMessageLike` type
2. Add a `sessionDefaultVariant()` function similar to `sessionDefaultModel()`
3. In `local.tsx`, include the `sessionDefaultVariant` as a fallback layer for `local.model.variant.current()` when no draft variant is set

This mirrors how model/agent inheritance works and ensures variant is restored from history for existing sessions too.

### Approach C (minimal): Restore the PR #301 effect

Re-add the variant restore effect that was removed during the composer-intent refactoring:

```typescript
// In session.tsx, when a session's messages are available:
const lastRootMsg = lastRoot()
if (lastRootMsg?.model && lastRootMsg?.variant) {
  local.model.variant.set(lastRootMsg.variant, lastRootMsg.model)
}
```

yzxoi:
## What this PR for?

Fixes #463. The issue reports that selecting a model effort/variant on the new-session composer sends the right variant to the backend, but the UI switches back to Default after the session is created. The refactored composer intent layer restored model and agent from session history but no longer restored variant.

## Summary

- Add session-default variant derivation from the last root user message, scoped to the current effective model.
- Read variant as session store draft first, then history fallback, preserving user draft precedence.
- Copy the selected new-session variant into the created session's variant store so the top-bar survives the workspace-to-session key switch.

## Test

- [x] lint 通过
- [x] format 通过
- [x] typecheck 通过
- [x] `bun test src/context/composer-intent.test.ts src/context/model-variant.test.ts`

## Before/After

**Before:** After first send, the route changed to the new session and the variant store key changed, so the top-bar showed Default despite the message using the selected effort.
**After:** The new session store receives the selected variant, and existing sessions derive variant from the last root message when no draft variant exists.


EricSanchezok:
Audit complete for this PR.

What this PR is fixing:
- The issue is real. A model variant selected in the new-session composer can be used for the first submit, but after the session is created the UI state is keyed by the session id instead of the workspace/new-session key, so the top bar can fall back to `Default`. Existing sessions also need to recover the selected variant from the last root message after the composer-intent layering change.

Root-cause assessment:
- The root problem was not the submit path alone. Variant was missing from the same layered composer intent model used for provider/model resolution: explicit draft state first, then read-only session default derived from history, then fallback. Without that layer, the UI loses the session's effective variant even when the canonical root message carries it.
- The new-session case additionally needed a narrow handoff from the workspace-scoped draft to the created session-scoped draft after session creation.

Patch-quality assessment:
- The implementation is properly scoped. `sessionDefaultVariant()` derives from the last root user message only when its provider/model still matches the current effective model, so stale variants do not bleed across model changes.
- `variant.current()` checks `has(modelKey)` before reading, which preserves the important distinction between “no draft exists, use session history” and “draft explicitly selected Default”. That avoids a subtle compatibility-style fallback bug.
- The submit change only copies the variant when a session was just created and a variant is actually selected. It does not add a parallel persistence model.
- I do not see redundant, legacy, wrapper, or dead-code paths introduced by this fix. The main-line composer layering remains clear and aligned with the frontend data-sync design.

Additional work I did:
- Merged the branch up to current `origin/dev` in `cc99b690e439aef808a8143e05631e7bf42c491f`. This was needed because the branch carried an older SDK package script shape that made `bun run quality:quick` fail locally on Windows. After the merge, the PR diff against `dev` remains focused on the model-variant/composer files.

Verification:
- `bun test src/context/composer-intent.test.ts src/context/model-variant.test.ts` from `packages/app`
- `bun run typecheck`
- `bun run quality:quick`
- pre-push hook on `git push origin synergy/fix-issue-463`
- GitHub CI on HEAD `cc99b690e439aef808a8143e05631e7bf42c491f`: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

PR #483, opened 2026-07-08
fix(session): settle session_send delivery

### LINKED ISSUE #462: bug: session_send tool execution does not settle after successful delivery, and session_send-initiated sessions stop after subagent returns
## User report

> @chaos 当前message send tool存在已经成功发送到对话（或者是消息信箱）的情况下message send tool一直不停止，同时使用message send发起的对话会导致发起的对话中即使调用了subagent，subagent返回后无法重新激活该对话而是直接停止了

Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6bd717f6c4a0c2d6a01bcb2e73c

## Triage

- Type: bug_report
- Related area: area/session, area/tools

## Deduplication

No duplicate found in recent open/closed issues. Related but different:
- #444 (closed, duplicate): feature request to create primary sessions via session_send — not about tool lifecycle
- #244 (closed): subagent completion notification re-triggers primary agent — opposite direction (session re-triggered vs session stops)
- #202 (closed): queued inbox messages not processed after session execution ends — different mechanism

### 1. session_send tool execution doesn't settle after delivery

**File: `packages/synergy/src/tool/session-send.ts` (lines 35–84)**

The tool calls `SessionManager.deliver()` and returns a result. The tool itself completes synchronously, but the LLM in the calling session may keep calling `session_send` in a loop (doom loop) because:
- When delivering as `role: "assistant"`, `SessionManager.deliver` (line 408–421 in manager.ts) delivers with mode `"context"` which is silent — no response is triggered in the target session, so the calling agent sees no evidence of delivery
- The `mailbox: true` metadata sets `channelPush` to true (invoke.ts line 437), but the outbound channel push (channel/outbound.ts line 27) only fires when the **assistant message in the source session** completes — not when the delivery to the target completes
- The tool result metadata (`{ sessionID, role }`) may not be informative enough for the LLM to realize the message was fully delivered

**File: `packages/synergy/src/session/invoke.ts` (line 437)**
```
const channelPush = !!(userMetadata?.mailbox || userMetadata?.channelPush)
```

**File: `packages/synergy/src/session/manager.ts` (lines 408–421)**
Delivering as `role: "assistant"` calls `SessionInbox.deliver` with mode `"context"` and returns without waking the target session. The calling agent gets no confirmation that the target agent processed the message.

### 2. session_send-initiated conversation stops after subagent returns

**File: `packages/synergy/src/session/session-send.ts` (lines 46–51)**

The `session_send` tool sets `mailbox: true` and `source: "mailbox"` in the mail metadata. When delivered as `role: "user"`, this creates an inbox item with origin `{ type: "channel" }` (message-v2.ts line 739).

**File: `packages/synergy/src/session/message-v2.ts` (line 739)**
```
if (metadata.channelPush || metadata.mailbox) return { type: "channel" }
```

The origin type being "channel" affects the session's loop behavior in `invoke.ts` (line 437):
```
const channelPush = !!(userMetadata?.mailbox || userMetadata?.channelPush)
```

This causes the assistant response to be marked with `{ metadata: { channelPush: true } }`. When the session processes a user message from `session_send`, the LLM turnaround flow may behave differently because of this metadata.

**File: `packages/synergy/src/cortex/manager.ts` (lines 461–477)**

When a subagent (task) completes:
1. If in synchronous mode (the `task` tool waits for the result), waiter gets the result directly — no mail notification to parent session
2. The parent session continues processing the task result
3. After the turn completes, `hasRunnableItem` (inbox.ts line 338) checks for steer items with `latestRootID`
4. If there are steer items from the subagent completion, the loop re-enters

The potential issue: the `session_send`-originated session may not properly handle the `channelPush` metadata in subsequent loop iterations. The channel outbound push (channel/outbound.ts) fires on the completion of an assistant message with `channelPush` metadata — but if the steer item from the subagent completion triggers a new turn, the resulting assistant message may not have the `channelPush` metadata, causing the response to not be pushed to the channel.

## Expected behavior

1. After `session_send` delivers a message (either `role: "assistant"` or `role: "user"`), the tool execution should settle/completenormally and the LLM should receive the result and proceed to its next action (text response or next tool call).

2. In a session that received a user message via `session_send`, when the agent calls a subagent (`task` tool), the subagent's result should be consumed by the agent, and the session should continue processing until a terminal finish. Subsequent subagent completion notifications (steer items) should either be suppressed (for synchronous tasks already consumed) or properly handled to continue the conversation.

## Actual behavior

1. `session_send` tool execution stays in a running/pending state or the LLM keeps re-invoking the tool despite successful delivery.

2. A session that received a message via `session_send` stops responding after a subagent returns — the subagent result is consumed but the session does not continue processing (the loop exits prematurely or stuck in an unresponsive state).

## Proposed fix

1. For `session_send` tool completion:
   - Ensure the tool result clearly indicates successful delivery (already does at the code level, may need better LLM-facing feedback)
   - Verify the tool execution slot settles properly in the processor (processor.ts lines 875–888)
   - Check for doom loop patterns with `session_send` and add appropriate safeguards

2. For session lifecycle after subagent return in `session_send`-initiated sessions:
   - Review how `mailbox`/`channelPush` metadata flows through the session lifecycle
   - In `invoke.ts`, when processing steer items after a synchronous task completion, ensure the assistant message properly inherits the `channelPush` metadata
   - Review the interaction between `hasRunnableItem` and the loop re-entry logic in `loopBody` (invoke.ts lines 934–936)

yzxoi:
## What this PR for?

Fixes #462. The issue reports that `session_send` can appear to stay running after a successful delivery, and sessions started via `session_send` can diverge into mailbox/channel semantics after subagent work. The root cause is that user-role delivery awaited the target session processing, and `session_send` marked internal session delivery as mailbox metadata.

## Summary

- Queue `session_send` role=user deliveries asynchronously with `waitForProcessing: false` so the source tool call settles after enqueue/wake scheduling.
- Mark session_send mail as internal `source: "session_send"` with `sourceSessionID`, not mailbox/channel metadata.
- Make the tool result explicitly state that delivery is queued/delivered and the tool call is complete.

## Test

- [x] lint 通过
- [x] format 通过
- [x] typecheck 通过
- [x] `bun test test/tool/session-send.test.ts test/session/inbox.test.ts test/session/invoke.test.ts`

## Before/After

**Before:** role=user delivery could wait on the target session turn, and session_send-origin messages carried mailbox/channel metadata.
**After:** role=user delivery returns after asynchronous enqueue/wake scheduling, and target sessions process session_send messages with normal internal session origin semantics.


EricSanchezok:
Audit completed for #483.

Findings:

1. Problem addressed
   This PR fixes #462: `session_send` user-role delivery could make the source tool call wait on the target session turn, and messages sent through `session_send` were tagged with mailbox/channel-style metadata that could skew downstream session behavior.

2. Whether the problem is real
   Yes. The old tool path used `SessionManager.deliver()` without `waitForProcessing: false` for user-role mail, so an idle target session could be woken and processed before the source tool call settled. It also wrote `mailbox: true` / `source: mailbox`, which maps through existing message semantics to agenda/channel-like origins instead of an internal session-to-session origin.

3. Root-cause assessment
   The root cause is delivery semantics, not just weak tool-result wording. The PR fixes the two relevant layers: user-role delivery is queued and scheduled asynchronously, and the metadata now identifies the sender as `source: session_send` with `sourceSessionID` rather than mailbox/channel metadata.

4. Patch quality / mainline complexity
   The change is narrow and follows the existing `SessionManager.deliver(waitForProcessing: false)` path already used by Blueprint/Lattice delivery. It does not add a compatibility wrapper or special post-hoc handling in invoke. The remaining legacy mailbox/channel mapping stays where it already exists; this PR stops `session_send` from opting into it unnecessarily.

5. Additional changes
   I did not push additional code for this PR. The implementation and tests are sufficient after audit.

Verification:

- `bun test test/tool/session-send.test.ts test/session/inbox.test.ts test/session/invoke.test.ts` from `packages/synergy`
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run quality:quick`
- GitHub CI for head `b78f91ee540efda1eba579928162661cc196fb31`: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

PR #484, opened 2026-07-08
fix(frontend): preserve note scroll on metadata updates

### LINKED ISSUE #442: bug: Note document auto-refreshes and jumps to top when metadata-only updates arrive (BlueprintLoop status or other note.updated events)
## User report

> @chaos 现在的note系统啊，我打开一篇文档，这篇文档会时不时的自动刷新一下，导致我本来看到一半了，又自动跳到开头，体验很差

Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6beb98341054b3900bdc13f81f4

## Triage

- Type: bug_report
- Related area: area/frontend

## Deduplication

No duplicate found in recent open issues. Related to closed issue #303 (Notebook flickers during BlueprintLoop — closed 2026-07-06 as resolved by PR #214 and #329), but the root cause identified in #303 Hypothesis A — `applySnapshot()` unconditionally calling `ed.commands.setContent()` on metadata-only note updates — was never fixed. The #329 reconcile fix reduced the frequency of unnecessary store writes but did not eliminate the path: BlueprintLoop status changes (or any metadata-only note update like pin/rename) still trigger `note.updated` → `noteVersion++` → NoteEditor refetch → `applySnapshot()` → `setContent()` → scroll position reset to top.

## Static inspection

- `packages/app/src/components/note-panel.tsx` lines 1472-1491: `applySnapshot()` unconditionally calls `ed.commands.setContent()` and then `ed.commands.setTextSelection(from)`. The `setContent` causes the ProseMirror editor to rebuild DOM, losing viewport scroll. The cursor restoration (`setTextSelection`) does not restore vertical scroll.
- `packages/app/src/components/note-panel.tsx` lines 1589-1607: The effect watches `note()` resource changes, calls `applySnapshot()` whenever version is higher and editor is not dirty.
- `packages/app/src/components/note-panel.tsx` lines 1360-1381: Note resource is refetched on `globalSync.noteUpdate()` when a `note.updated` event arrives with higher version.
- `packages/app/src/context/global-sync.tsx` lines 784-796: Every `note.updated` event bumps `noteVersion` and fires `noteUpdate` signal.
- `packages/synergy/src/blueprint/loop-store.ts` lines 90, 176: BlueprintLoop status changes update note metadata (`activeLoopID`, `runCount`, `lastRunAt`) via `NoteStore.update()`, which increments note version and emits `note.updated`.

## Expected behavior

- Opening and reading a Note document should be stable. The editor should not re-render or reset scroll position unless the document content actually changes.
- Blueprint runtime metadata updates (activeLoopID, runCount, lastRunAt) should not cause a full editor snapshot reapplication.

## Actual behavior

- While viewing a note, periodic `note.updated` events (from BlueprintLoop status or other metadata changes) cause the editor to re-fetch the note, call `applySnapshot()`, which calls `ed.commands.setContent()` unconditionally, resetting the viewport scroll to the top of the document.

## Proposed fix

Make `applySnapshot()` content-aware — only call `ed.commands.setContent()` when the note content actually changed. If only metadata (title, tags, blueprint) changed, update `baseNote`, `title`, `tags` signals without touching the editor content.

```ts
function applySnapshot(snapshot: NoteInfo) {
  const current = baseNote()
  const contentChanged = !deepEqual(current?.content, snapshot.content)
  const titleChanged = current?.title !== snapshot.title
  const tagsChanged = !deepEqual(current?.tags ?? [], snapshot.tags ?? [])

  setBaseNote(snapshot)

  if (titleChanged) setTitle(snapshot.title)
  if (tagsChanged) setTags(snapshot.tags ?? [])

  const ed = editor()
  if (contentChanged && ed && !ed.isDestroyed) {
    const { from } = ed.state.selection
    ed.commands.setContent(snapshot.content as any, { emitUpdate: false })
    if (from > 0) {
      try { ed.commands.setTextSelection(from) } catch { /* position may be invalid */ }
    }
  }

  setConflict(null)
  setDirty(false)
}
```

This was originally proposed as Candidate Fix 1 in #303 but was not implemented. The #329 reconcile fix helped but is insufficient for this scenario.

yzxoi:
## What this PR for?
Fixes #442.

Metadata-only `note.updated` events from BlueprintLoop status changes, pin/global updates, or other note metadata writes could refresh the note resource and force the open Tiptap editor to reset its document content. That rebuilt the editor DOM and sent the reader back to the top of the note.

## Summary
- Add a note snapshot delta helper so metadata-only refreshes update note state without calling `setContent()`.
- Keep `DocumentEditorCore` editor construction mount-only, matching its initial-content contract.
- Cover metadata-only, title/tag-only, and real content-change cases with focused tests.

## Test
- [x] `bun test src/components/note/note-snapshot.test.ts src/components/note/blueprint-run-session.test.ts` from `packages/app`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run --cwd packages/app build`
- [x] `bun run quality:quick`

## Before/After
**Before:** metadata-only note refreshes could rebuild the editor and jump the note viewport to the top.
**After:** metadata refreshes update note metadata without resetting editor content or remounting the Tiptap editor.

### DIFF
--- a/packages/app/src/components/note-panel.tsx
+++ b/packages/app/src/components/note-panel.tsx
@@ -32,6 +32,7 @@ import {
   isActiveBlueprintLoopStatus,
   type BlueprintRunMode,
 } from "@/components/note/blueprint-run-session"
+import { getNoteSnapshotDelta } from "@/components/note/note-snapshot"
 import "./note-panel.css"
 
 type LoopStatus = BlueprintLoopInfo["status"]
@@ -1470,22 +1471,23 @@ function NoteEditor(props: { id: string; directory: string; onBack: () => void; 
   }
 
   function applySnapshot(snapshot: NoteInfo) {
+    const delta = getNoteSnapshotDelta(baseNote(), snapshot)
     const ed = editor()
     setBaseNote(snapshot)
-    setTitle(snapshot.title)
-    setTags(snapshot.tags ?? [])
+    if (delta.titleChanged) setTitle(snapshot.title)
+    if (delta.tagsChanged) setTags(snapshot.tags ?? [])
     setConflict(null)
     setDirty(false)
-    if (ed && !ed.isDestroyed) {
-      const { from } = ed.state.selection
-      ed.commands.setContent(snapshot.content as any, { emitUpdate: false })
-      const docSize = ed.state.doc.content.size
-      if (from > 0 && from < docSize) {
-        try {
-          ed.commands.setTextSelection(from)
-        } catch {
-          /* position may be invalid */
-        }
+    if (!delta.contentChanged || !ed || ed.isDestroyed) return
+
+    const { from } = ed.state.selection
+    ed.commands.setContent(snapshot.content as any, { emitUpdate: false })
+    const docSize = ed.state.doc.content.size
+    if (from > 0 && from < docSize) {
+      try {
+        ed.commands.setTextSelection(from)
+      } catch {
+        /* position may be invalid */
       }
     }
   }
--- a/packages/app/src/components/note/document-editor-core.tsx
+++ b/packages/app/src/components/note/document-editor-core.tsx
@@ -1,4 +1,4 @@
-import { createEffect, createSignal, onCleanup, Show } from "solid-js"
+import { createSignal, onCleanup, onMount, Show } from "solid-js"
 import { Editor } from "@tiptap/core"
 import StarterKit from "@tiptap/starter-kit"
 import Placeholder from "@tiptap/extension-placeholder"
@@ -489,7 +489,7 @@ export function DocumentEditorCore(props: DocumentEditorCoreProps) {
   let bubbleRef!: HTMLDivElement
   const [editorInstance, setEditorInstance] = createSignal<Editor>()
 
-  createEffect(() => {
+  onMount(() => {
     const instance = new Editor({
       element: editorRef,
       extensions: createDocumentEditorExtensions({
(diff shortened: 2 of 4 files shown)


EricSanchezok:
Audit completed for #484.

Findings:

1. Problem addressed
   This PR fixes #442: metadata-only `note.updated` refreshes could cause the open Tiptap note editor to call `setContent()`, rebuild the editor DOM, and jump the reader back to the top of the document.

2. Whether the problem is real
   Yes. The event path is real: note metadata changes, including BlueprintLoop note metadata writes, bump note version and trigger the editor resource refresh. Before this PR, `applySnapshot()` unconditionally called `ed.commands.setContent()` when applying a newer snapshot, even when the actual document JSON was unchanged.

3. Root-cause assessment
   The root cause was content and metadata being treated as one undifferentiated snapshot update. The PR addresses that at the right layer by adding a content-aware delta and only touching the Tiptap document when the content actually changes. This is not just masking the scroll position after the fact.

4. Patch quality / mainline complexity
   The implementation is clean and narrowly scoped. `getNoteSnapshotDelta()` is a small pure helper, and `applySnapshot()` keeps the single main flow while skipping only the expensive editor content reset on metadata-only updates. I did not find redundant compatibility logic, wrapper layers, dead code, or legacy fallback paths.

5. Additional changes
   I did not push additional code for this PR. The existing implementation is sufficient after audit.

Verification:

- `bun test src/components/note/note-snapshot.test.ts src/components/note/blueprint-run-session.test.ts` from `packages/app`
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run quality:quick`
- GitHub CI for head `6a3eb17d5d767610320670d78ad5fa735753b151`: Quality, Typecheck, Test, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test all passed.

yzxoi:
Already fix in #488

PR #488, opened 2026-07-08
Fix Note editor sync stability

### LINKED ISSUE #442: bug: Note document auto-refreshes and jumps to top when metadata-only updates arrive (BlueprintLoop status or other note.updated events)

EricSanchezok:
## Summary

Fixes #442.

This PR fixes the Note/Blueprint editor instability at the root instead of adding compatibility branches or legacy fallbacks.

### Root cause

- `NoteEditor.applySnapshot()` treated metadata updates like content updates, so autosave echoes, pin/global/tag changes, and BlueprintLoop status updates could call `setContent()` and reset selection/scroll.
- `DocumentEditorCore` initialized Tiptap from a reactive effect that read `props.content`, so content object reference churn could recreate the editor instance.
- Notes list/detail sync used coarse `globalSync.noteVersion()` invalidation and duplicate loop resources, causing full refetches and races between note metadata and BlueprintLoop state.
- Autosave tracked one coarse dirty boolean, so in-flight saves could clear a field that changed again during the request.

## Changes

- Make `DocumentEditorCore` mount-only and read initial content with `untrack`; parent code now controls explicit snapshot application.
- Add field-level dirty revisions for `title`, `content`, and `tags`; save responses only clear fields captured by that request and unchanged since.
- Make snapshot application content-aware: metadata-only and content-identical updates update base state/header/Blueprint state without touching Tiptap DOM.
- Merge remote updates by changed field: non-overlapping remote Blueprint/metadata changes merge into local dirty drafts; overlapping content/title/tag changes produce a conflict.
- Replace Notes panel coarse invalidation with local event patching for note create/update/archive/unarchive/delete and BlueprintLoop create/update.
- Remove `globalSync.noteVersion()` / `noteUpdate()` and the duplicate NoteEditor loop resource.
- Expand Note events with `scopeID`, `meta`, `changed`, and archive `metas`; regenerate SDK/OpenAPI types.
- Make batch archive/unarchive/delete resolve the actual note scope so project Notes views can operate on visible home/global notes cleanly.
- Add tests covering scoped note event metadata, BlueprintLoop note metadata updates, frontend sync helpers, autosave dirty revision semantics, local list patching, and cross-scope global note archive/unarchive.
- Document the Note/Blueprint detail stability rule in `packages/app/PRODUCT.md`.

## Migrations

No migration. The persisted Note schema is unchanged; this only changes event payloads, sync behavior, and frontend state management.

## Verification

Passed:

- `bun test test/note/store.test.ts test/blueprint/loop-store.test.ts`
- `cd packages/app && bun test src/components/note/note-sync.test.ts`
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run monorepo:check`
- `bun packages/sdk/js/script/build.ts`

`bun run quality:quick` still fails at the existing Windows package-check path for `@ericsanchezok/synergy-sdk`: `bun run build` tries to execute `./script/build.ts` directly and reports `bun: command not found: ./script/build.ts`. The same SDK build logic passes when invoked explicitly with Bun via `bun packages/sdk/js/script/build.ts`, and the push hook passed format/lint/typecheck/monorepo checks.


yzxoi (dismissed):
Reviewed — this is a solid fix at the root cause. The field-level dirty tracking, event-driven incremental patching, and content-aware snapshot application are the right design direction. Extracting `note-sync.ts` as a pure-function module is good engineering practice.

Three minor observations (non-blocking, can be follow-up PRs):
1. `store.ts:336` — `structuredClone(current)` does a full deep copy just for `changedFields` comparison; could copy only the compared fields for large documents.
2. `store.ts:71` vs `note-sync.ts:67` — Backend uses `NoteDocument.hash` (canonical key-sorted SHA-256), frontend uses `isDeepEqual` (key-order-sensitive). Semantically equivalent in practice because Tiptap/structuredClone preserve key order, but worth unifying the comparison strategy later.
3. `NoteEditor` has two independent remote-update entry points (`createEffect` on resource + `sdk.event.listen`); the version guard prevents double-apply, but it's a subtle coupling worth documenting.

Thanks for the thorough work on this!

yzxoi (approved):

PR #504, opened 2026-07-09
fix(bus): rate-limit streaming publish debug logs

SII-jake:
## Summary

- Aggregate streaming event publish counts over a 5-second window and log the rate (events/sec) instead of logging every individual publish
- Prevents log flooding from high-frequency streaming events while retaining visibility into publishing activity
- Adds a `recordStreamingPublish()` helper with a per-event-type stats map


yzxoi (approved):
Thanks for this PR, @SII-jake! The rate-limited streaming debug logging is a clean improvement that prevents log flooding during high-frequency part-delta publishing.

I pushed one follow-up commit (`66da1f1`) that addresses two things:

1. **Removed the immediate "first event" log** — the original code logged `{ count: 1, windowMs: 0, ratePerSecond: 0 }` on the first streaming event for a type, which was a useless log line and excluded that event from the first aggregate rate window. Now the first event is silently accumulated into the first 5-second window.

2. **Added test coverage** in `test/bus/publish.test.ts`:
   - 200 rapid streaming publishes don't crash, all events arrive at subscribers
   - 10 state event publishes still work correctly (non-streaming path untouched)
   - Two streaming event types have independent rate counters

All 24 bus tests pass, and pre-push hooks (format, lint, typecheck, monorepo check) are green.

Nice work on this — simple, focused, correct optimization.

PR #506, opened 2026-07-09
feat(provider): add proxy/noProxy env override options

SII-jake:
## Summary

Solve `ECONNRESET` issues with internal providers (e.g. `sii.edu.cn`) behind ClashX system proxy.

Two new provider options to control proxy behavior:

| Option | Type | Behavior |
|---|---|---|
| `proxy` | `string` | Set `https_proxy` for this provider's requests |
| `noProxy` | `boolean` | Set `NO_PROXY=*` to bypass system proxy |

## How it works

Temporarily overrides `process.env` proxy variables before each fetch, then restores them in `finally`. Bun's native fetch respects these env vars — no extra subprocess needed.

## Usage

```jsonc
{
  "provider": {
    "sii": {
      "options": {
        "noProxy": true
      }
    }
  }
}
```

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -1,0 +1,2 @@
+1580842cd4b7af1decf47ffc6af2375010f76a80:packages/synergy/test/observability/redaction.test.ts:curl-auth-header:36
+1580842cd4b7af1decf47ffc6af2375010f76a80:packages/synergy/test/observability/diagnostics.test.ts:curl-auth-header:83
--- a/docs/open-source-quality.md
+++ b/docs/open-source-quality.md
@@ -101,12 +101,12 @@ bun run package:check
 
 ## Workflow Validation
 
-The `workflow:check` script downloads actionlint and runs it against all workflow files in `.github/workflows/`, then runs zizmor for GitHub Actions security analysis.
+The `workflow:check` script uses an installed `actionlint` binary when available, otherwise downloads the pinned actionlint release, then runs zizmor for GitHub Actions security analysis.
 
 Run locally:
 
 ```bash
-bun run workflow:check      # requires network access
+bun run workflow:check      # install actionlint locally to avoid the actionlint download fallback
 ```
 
 ## Secret Scanning
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -44,6 +44,7 @@
     "tree-sitter-bash",
     "xlsx",
     "turbo",
+    "undici",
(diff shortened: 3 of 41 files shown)


yzxoi:
Thanks for this PR — the ClashX + internal API proxy issue is a real pain point.

I pushed a follow-up commit that replaces the `process.env` modify/restore approach with undici's `Agent`/`ProxyAgent` passed as a per-request `dispatcher` to Bun's `fetch`. Two reasons: (1) `process.env` mutation is a global side effect with a race condition when multiple provider requests overlap, and (2) the import probe path (`createSDKFromSpec`) was not covered.

### How to use

After merge, add `proxy` or `noProxy` to the provider or model options in `20-providers.jsonc`:

```jsonc
// Bypass system proxy for an internal API broken by ClashX:
{
  "provider": {
    "openai-compatible": {
      "api": "https://sii.edu.cn/v1",
      "options": {
        "noProxy": true,
        "apiKey": "sk-xxx"
      }
    }
  }
}

// Route all requests through a specific proxy:
{
  "provider": {
    "openai": {
      "options": {
        "proxy": "http://127.0.0.1:7890",
        "apiKey": "sk-xxx"
      }
    }
  }
}
```

Model-level override also works (`models.<id>.options.proxy`).

Added 11 behavioral tests in `test/provider/proxy.test.ts`. The existing `.catchall(z.any())` on the provider options schema already accepts these fields, so no schema migration is needed.

yzxoi (dismissed):

yzxoi:
Fixed the provider-level `noProxy` handling so it now applies to the actual AI SDK request path, not just the exposed SDK fetch helper.

What changed:
- Inject proxy-aware `fetch` at provider SDK construction time, so `generateText`, `streamText`, and OpenAI `responses()` all use the configured transport.
- Route `noProxy: true` through the direct transport and give it precedence over explicit `proxy`.
- Include provider/model transport options in the model cache key to avoid reusing stale language model instances after proxy config changes.
- Added regression coverage for direct fetch, explicit proxy precedence, `generateText`, `streamText`, and OpenAI responses paths.

Why: Bun 1.3.14 can honor startup proxy env at the fetch layer, and `dispatcher` is not enough to force a no-proxy path. The fix makes `noProxy` an explicit Synergy transport decision for model requests while documenting that system-level/TUN transparent proxying still has to be handled by network rules.

yzxoi (approved):

Issue #328, opened 2026-07-05
feat: add force-refresh button to 'No messages yet' empty state


yzxoi:
## User report

> @chaos 前端在出现no message yet的信息后应该增加一个强制刷新的按钮

Reporter: Feishu:ou_29e9348c6b7a9818a0458f24ed03fcfe
Source: group:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6b938bef90acc2cc72483bf6dbc

## Triage

- Type: feature_request
- Target repo: SII-Holos/synergy
- Related area: area/frontend, area/session

## Deduplication

Related open issue #316 ("Investigate intermittent 'No messages yet' session view until hard reload") investigates the root cause of the session appearing empty until hard reload. This issue proposes a complementary UX improvement: adding a **force refresh button** directly in the empty-state view so users can recover without manually reloading the page.

No duplicate found for this specific feature request.

## Static inspection

Location: `packages/app/src/pages/session.tsx` lines 988-1005

The "No messages yet" empty state is rendered as a fallback in a `<Show>` block (line 989-1005):

```tsx
<Show
  when={activeMessage() || (timeline()?.length ?? 0) > 0}
  fallback={
    <Show when={messagesReady()} ...>
      <div class="synergy-workbench-canvas flex h-full items-center justify-center bg-background-stronger">
        <span class="text-sm text-text-weak">No messages yet</span>
      </div>
    </Show>
  }
>
```

When `activeMessage()` is `undefined` AND `timeline().length === 0`, this empty state is shown. The known intermittent issue (#316) can cause a newly opened/refreshed session to land here even when messages exist, forcing the user to hard-reload the page.

## Expected behavior

When the "No messages yet" empty state is rendered, the UI should also display a **refresh / reload button** (e.g. an icon button or text link) that triggers a session data re-fetch without requiring a full page reload. This provides a recovery path for the intermittent empty-state issue tracked in #316.

## Proposed fix

In `packages/app/src/pages/session.tsx`, update the empty-state block (lines 1000-1003) to include a refresh button:

```tsx
<div class="synergy-workbench-canvas flex h-full flex-col items-center justify-center gap-3 bg-background-stronger">
  <span class="text-sm text-text-weak">No messages yet</span>
  <button
    class="text-sm text-accent hover:text-accent-hover underline cursor-pointer"
    onClick={() => {/* re-fetch session messages */}}
  >
    Refresh
  </button>
</div>
```

[...]

PR #534, opened 2026-07-10
feat: unify loading and empty states into single spinner with 10s timeout recovery

EricSanchezok:
## Summary

Merge the two conversation-area placeholders — the "Loading conversation…" spinner and the "No messages yet" plus Refresh button — into one unified loading state.

- Always shows a spinner ("Loading conversation…") while messages haven't loaded yet
- After 10 seconds without data, a "Refresh" button fades in below the spinner so the user can recover
- Removes the "No messages yet" text entirely, along with its Refresh button, the `emptyRefreshing` signal, and the `sync.session.refresh()` method

## First Principles

Showing "No messages yet" on a session that should have messages is a user-hostile experience: it forces the user to manually refresh and question whether the app is broken. The app knows it hasn't completed loading — it should stay in a loading state, not pretend the session is empty. A single spinner with a timed recovery button communicates clearly: "we're loading; if it takes unusually long, here's a way to retry."

## Changes

| File | What |
|---|---|
| `packages/app/src/pages/session.tsx` | Drop `emptyRefreshing` signal, unify fallback into single `Show when={!messagesReady()}` block, add 10s timer that fades in a Refresh button |
| `packages/app/src/context/sync.tsx` | Delete `sync.session.refresh()` method (only caller was the old Refresh button) |
| `packages/app/src/index.css` | Add `@keyframes fade-in` for the timed Refresh button appearance |

## Verification

- [ ] `rg "emptyRefreshing" packages/app/src/` → 0 matches
- [ ] `rg "No messages yet" packages/app/src/` → 0 matches
- [ ] `rg "sync\.session\.refresh" packages/` → 0 matches
- [ ] Format, lint pass on changed files

### DIFF
--- a/packages/app/src/context/sync.tsx
+++ b/packages/app/src/context/sync.tsx
@@ -353,19 +353,6 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
           await Promise.all(requests)
           refreshPlanBlueprintOfferFromLoadedParts(store, setStore, sessionID)
         },
-        // Force a fresh re-fetch of a session's messages and volatile state,
-        // bypassing the "already loaded" short-circuit in sync(). Used by the
-        // empty-state Refresh button to recover if the initial load missed
-        // messages or session metadata such as derived rollback state (issue
-        // #328 / #316).
-        async refresh(sessionID: string) {
-          const limit = meta.limit[sessionID] ?? chunk
-          await Promise.all([
-            loadSession(sessionID, { force: true }).catch(() => {}),
-            loadMessages(sessionID, limit),
-            refreshVolatile(sessionID),
-          ])
-        },
         async diff(sessionID: string) {
           if (store.session_diff[sessionID] !== undefined) return
 
--- a/packages/app/src/index.css
+++ b/packages/app/src/index.css
@@ -123,6 +123,15 @@ a:active,
   }
 }
 
+@keyframes fade-in {
+  from {
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+
 .statusbar-glass {
   position: relative;
   background: var(--workbench-control-bg, var(--surface-raised-base));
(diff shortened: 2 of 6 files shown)


yzxoi (changes requested):
The loading-state direction is reasonable, but this implementation introduces two blocking regressions that need to be fixed before merge.

## 1. Loaded-empty sessions render a completely blank canvas

`messagesReady()` only checks whether `sync.data.message[id] !== undefined`. `loadMessages()` writes an empty array into that bucket when the API returns no messages, so a successful empty response produces:

- `messagesReady() === true`
- no active message
- an empty timeline

The new inner `<Show when={!messagesReady()}>` has no fallback, so that state renders nothing at all: no spinner, empty-state text, or recovery action. This is reachable for a genuinely empty workspace session and for the missed/temporarily empty snapshot scenario this change is intended to improve.

Please preserve an explicit loaded-empty branch rather than rendering nothing. At minimum, the UI needs three observable states: loading, loaded-empty/recovery, and conversation.

## 2. The new Refresh action no longer performs a real message refresh

The removed `sync.session.refresh()` deliberately bypassed the loaded-state short circuit and re-fetched session data, messages, and volatile state. The replacement calls:

```ts
await sync.session.sync(id, { refreshVolatile: true })
```

But `sync()` treats any existing message bucket—including a stale or accidentally empty `[]`—as `hasMessages === true` and skips `loadMessages()`. Once an initial request stores an empty result, the new action cannot recover the missing messages. If the original request is still in flight after 10 seconds, the call also reuses the same inflight promise rather than initiating a fresh request.

This removes the recovery path introduced for #328/#316. The retry action needs explicit force-refresh semantics that bypass the message/session cache short circuit, whether by restoring `refresh()`, adding a well-defined `forceMessages`/`forceSession` option, or introducing a dedicated `forceReload()` operation.

## Additional requested cleanup

- Handle rejection from the async Refresh click path so a failed retry does not become an unhandled promise rejection.
- Restore a disabled/loading state while refresh is running to prevent repeated clicks and provide feedback.
- Add focused coverage for at least: unloaded → loading, loaded-empty, failed/missed initial load followed by a forced retry, and loaded-with-messages.

CI is green, but there is currently no test covering these loading/empty-state transitions. Please address the two blockers above before merging.

yzxoi:
Implemented in #554

PR #538, opened 2026-07-10
fix: key provider catalog cache by includeLive to prevent stale static data

### LINKED ISSUE #521: Provider catalog static cache suppresses authenticated live model discovery
## Summary

`ProviderCatalog.resolve()` uses one shared in-memory cache and one shared in-flight request for both static catalog resolution (`includeLive: false`) and authenticated live model discovery (`includeLive: true`).

If a static lookup happens first, later live discovery returns the cached static catalog without calling the provider's live model endpoint. The provider/model UI can therefore omit models that the authenticated account can use.

This is reproducible on current `dev` at `cb6a5123`.

## Observed behavior

For an authenticated `openai-codex` account:

- Codex CLI's current model catalog contains:
  - `gpt-5.6-sol`
  - `gpt-5.6-terra`
  - `gpt-5.6-luna`
- An isolated Synergy resolution with the same config/auth and `forceRefresh: true, includeLive: true` returns those three models.
- The running Synergy `GET /provider` response only contains the static fallback models through `gpt-5.5`.

The stale local GPT-5.5 context override was removed and is not responsible: live discovery still returns GPT-5.6 models with no provider override.

## Reproduction

```ts
ProviderCatalog.reset()

const staticCatalog = await ProviderCatalog.resolve({
  config,
  includeLive: false,
})

const liveCatalog = await ProviderCatalog.resolve({
  config,
  includeLive: true,
})
```

Expected:

- The second call performs authenticated live discovery and returns the live model set.

Actual:

- The second call returns the first call's cached static result.
- The live discovery callback is never invoked.

The normal provider route makes this order likely:

1. `GET /provider` calls `ProviderCatalog.resolve({ includeLive: false })`.
2. It then calls `Provider.list()`, whose state requests `ProviderCatalog.resolve({ includeLive: true })`.
3. The second call hits the static cache.

## Root cause

`packages/synergy/src/provider/catalog.ts` stores:

```ts
let inFlight: Promise<...> | undefined
let memoryCache: { value: ...; createdAt: number } | undefined
```

Neither value is keyed by `includeLive`. The default cache TTL is one hour, so a static result can suppress live discovery for the whole cache window. Restarting is not a reliable workaround because the same request order can repopulate the static cache.

This affects every provider profile with `fetchModelCatalog` or `fetchModels`, not only OpenAI Codex.

## Suggested fix

Keep separate cache and in-flight entries for static and live resolution:

```ts
const memoryCache = new Map<boolean, CacheEntry>()
const inFlight = new Map<boolean, Promise<Catalog>>()
const cacheKey = input?.includeLive === true
```

Add a regression test that:

1. Resolves the static catalog.
2. Resolves with `includeLive: true`.
3. Verifies the live discovery callback runs once and returns live models.
4. Resolves live again and verifies the live result is cached independently.

A local implementation of this targeted change passes:

- `bun test test/provider/framework.test.ts`: 15 passed, 0 failed
- `bun run typecheck`
- Prettier and `git diff --check`

## Notes

GPT-5.6 is currently account/rollout dependent, so hard-coding preview model IDs into `DEFAULT_MODEL_IDS` is not the right general fix. The authenticated live catalog should remain the source of truth; static fallback models should only be used when live discovery is unavailable.

yzxoi:
## Summary

- Fixes #521: `ProviderCatalog.resolve()` used one shared in-memory cache for both static catalog resolution (`includeLive: false`) and authenticated live model discovery (`includeLive: true`)
- If a static lookup populated the cache first, later live discovery returned the stale static catalog without calling the provider's live model endpoint
- Replaced the single `memoryCache` / `inFlight` variables with Maps keyed by a `cacheKey(includeLive)` function ("static" | "live"), giving each variant independent caching and in-flight deduplication

## Changed Files

- `packages/synergy/src/provider/catalog.ts` — 19 insertions, 11 deletions

## Test Plan

- `bun test test/provider/` — existing provider tests pass (4 pass, failures are pre-existing environment issues with missing `node_modules` in this worktree)

### DIFF
--- a/packages/synergy/src/provider/catalog.ts
+++ b/packages/synergy/src/provider/catalog.ts
@@ -53,8 +53,8 @@ export namespace ProviderCatalog {
     .strict()
   type RemoteCatalog = z.infer<typeof RemoteCatalog>
 
-  let inFlight: Promise<Record<string, ModelsDev.Provider>> | undefined
-  let memoryCache: { value: Record<string, ModelsDev.Provider>; createdAt: number } | undefined
+  const memoryCache = new Map<string, { value: Record<string, ModelsDev.Provider>; createdAt: number }>()
+  const inFlightMap = new Map<string, Promise<Record<string, ModelsDev.Provider>>>()
 
   function fallbackModel(provider: ModelsDev.Provider, modelID: string): ModelsDev.Model {
     return {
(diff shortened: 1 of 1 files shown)


yzxoi:
Already Implemented in 0542892f

PR #547, opened 2026-07-10
refactor(browser): unify Protocol v2, CDP controller, and Host runtime

EricSanchezok:
## What does this PR do?

This PR replaces Synergy's Browser subsystem as one atomic migration. It removes the split Playwright/WebContents behavior, legacy browser tools, per-page Electron processes, screenshot presentation fallback, and protocol v1 implementation. There are no deprecated aliases, feature flags, dual implementations, or compatibility fallbacks left in the runtime.

### Shared Browser package and Protocol v2

- Adds the internal `@ericsanchezok/synergy-browser` package as the single owner of:
  - strict Zod Protocol v2 command/result/event schemas
  - structured locators and errors
  - opaque generation-bound snapshot refs
  - `CdpPageController`
  - actionability, screenshots, reads, inspection, waits, console/network/performance/audit/emulation
  - redaction, filename safety, staging leases, navigation policy, and native leases
- Uses the same controller through a `CdpTransport` for Playwright `CDPSession` and Electron `webContents.debugger`.
- Rejects unsupported selectors such as `button:has-text("Continue with Holos")` before execution with `browser_invalid_selector` and semantic locator guidance.
- Enforces unique matches, visibility/stability/enabled/editable/receives-events checks, bounded obstruction errors, stale refs, strict screenshot modes, and readonly eval with CDP side-effect rejection.

### Single-page session ownership

- Makes `BrowserSessionManager` the sole page owner with `empty`, `suspended`, `active`, `migrating`, and `failed` states.
- Adds owner-serial command queues, required `commandId`, replay deduplication, conflict detection, abort handling, and one-page invariants.
- Persists backend-neutral descriptors/checkpoints and lazily resumes only from an explicit resume or a page-requiring command.
- Keeps `GET /browser/session`, event WebSocket connections, and WebRTC signaling passive.
- Migrates legacy storage once, removes old persisted fields/profile paths, and preserves structured recovery failures.
- Captures and restores URL, cookies, origin storage, viewport, scroll, and form state during explicit backend migration, with rollback to the source backend on failure.

### Final tool and permission surface

- Keeps the non-overlapping Browser tools:
  - navigation, snapshot, action, wait, read, inspect, screenshot
  - eval, console, network, performance, audit, emulate
  - dialog, upload, downloads, clipboard, assets, annotate, view
- Deletes the old navigate/click/type/scroll/health/list/download/viewport tools and all registrations, prompts, taxonomy entries, renders, icons, and source-string tests tied to them.
- Splits capabilities for coordinates, readonly/trusted eval, clipboard, upload, downloads, private network, and emulation.
- Requires file read for uploads and destination write authorization for download/asset/trace exports.
- Preserves guarded/autonomous/full_access semantics without adding ask paths to autonomous mode.

### Host broker, presentation, and security

- Replaces one Electron process per page with one server-level Host broker that multiplexes owner pages.
- Uses the Desktop main process as the local native broker and a minimal `browser-host-main` entry for WebRTC hosting.
- Adds short-lived, single-use, owner/page/role-bound tickets; registration secrets; message/rate/concurrency limits; origin/version validation; offer generations; connection IDs; and ICE sequencing.
- Captures video only and denies page camera, microphone, location, notifications, display capture, and clipboard permissions by default.
- Adds a server-owned HTTP/HTTPS CONNECT gateway with DNS resolution at the gateway, DNS-rebinding defense, metadata/link-local/multicast denial, private-network capability checks, and controlled localhost development ports.
- Routes direct, click-triggered, popup, redirect, subresource, and download traffic through the shared policy/gateway boundaries.
- Stages uploads by content rather than client paths and stores downloads in owner-isolated managed storage with size, containment, symlink, normalization, and exclusive-create checks.
- Redacts credentials, cookies, authorization headers, secret query values, clipboard content, and upload content from Browser diagnostics.

### Web, Desktop, release, and documentation

- Makes Web use one minimal WebRTC Host daemon, local Desktop use the native broker, managed Desktop use the same broker with its packaged server, and remote Desktop use WebRTC.
- Dynamically loads the Browser workspace and developer panels; opening a viewer performs no build.
- Adds the dedicated Browser Host electron-builder target, signed manifest generation, exact version/protocol matching, Ed25519 signature verification, SHA-256 verification, safe atomic extraction, and release assets per OS/architecture.
- Regenerates OpenAPI and the TypeScript SDK; the App uses generated Browser control/session APIs.
- Updates release workflow, runbooks, README, architecture docs, PRODUCT guidance, AGENTS files, and the develop-synergy skill.

### Runtime issues found during final acceptance

- Fixed an OpenAPI recursive-locator `$ref` boundary by keeping strict recursive Protocol schemas internally and using an API-safe session response schema for SDK generation.
- Removed a process-wide Bun Glob failure from custom tool discovery by replacing that single scan with deterministic `readdir` discovery; this eliminated full-suite NUL-suffixed cwd failures without changing Config discovery.
- Fixed Host broker WebSocket wrapper identity handling. Hono can provide a different wrapper object across callbacks; the route now retains the registered socket identity for subsequent messages and close cleanup. Before this fix Web mode left a stale ready broker, rejected reconnects, and timed out `page.create`.

## Why?

The previous Browser subsystem mixed Playwright selector syntax, DOM `querySelector`, WebContents commands, workspace routing fallbacks, and multiple presentation/control paths. That produced failures such as:

- invalid CSS passed to `Document.querySelector`
- trusted eval rejected inconsistently
- 30-second locator timeouts for schema/selector mistakes
- overlays intercepting input without bounded structured errors
- invalid `browser_wait` discriminators
- different pages or behavior between `bun dev web`, `bun dev desktop`, and `bun dev desktop --managed`
- repeated Desktop builds and per-page Electron lifecycle problems

The new architecture has one protocol, one page owner, one CDP behavior implementation, explicit presentation ownership, and shared network/file security boundaries.

### Full repository gates

- `bun run quality`
  - 4,659 local Synergy tests passed, 1 platform-conditional symlink test skipped, 0 failed
  - all package tests, builds, typechecks, formatting, lint, monorepo checks, and package validation passed
- `bun run deadcode`
- `bun run workflow:check`
- `bun run secrets:check`
- `./script/generate.ts`
  - generated SDK successfully
  - verified every OpenAPI component `$ref` resolves
- `git diff --check`
- removal scans for legacy tool names, `sameHost`, `policyOverride`, protocol v1 runtime references, per-page Host process, and old fallback paths

### Browser-specific and runtime coverage

- Shared `packages/browser` protocol/controller/security tests: 27 passed
- Synergy Browser/backend/server/enforcement/tool tests: 304 targeted tests passed after isolated reruns; the same coverage is included in the full suite
- App Browser workspace tests: 27 passed
- Desktop Browser/profile/IPC/release tests: 16 passed
- `SYNERGY_DESKTOP_RUNTIME_TEST=1 bun test test/browser-runtime-smoke.test.ts`
  - real Electron Host executed Protocol v2 through the shared CDP controller
- Host artifact install tests cover signature verification, digest tampering, exact protocol/version matching, and archive traversal
- Shared controller contract covers fill/press/select/check/drag, iframe, shadow DOM, readonly/trusted eval, obstruction, stale/ambiguous/zero-match/invalid selectors, screenshots, reads, inspection, waits, and performance

### Isolated mode acceptance

Used separate `SYNERGY_HOME` directories and ports for each mode:

1. `bun dev web --no-open --server-port 4297 --app-port 3201`
   - passive state was `empty`, Host ready, presentation WebRTC
   - first navigation created one page
   - replaying the same `commandId` returned the same page/result without repeating navigation
   - Web Host broker recovered cleanly after the WebSocket identity fix
2. `bun dev desktop --server-port 4298 --app-port 3202`
   - passive state was `empty`
   - broker advertised native + WebRTC
   - the external/remote client correctly selected WebRTC and created one page
3. `bun dev desktop --managed --server-port 4299 --app-port 3203`
   - packaged-style server and local broker started successfully
   - passive state was `empty`
   - first navigation created one page

All three process trees were stopped and checked for orphan server, Vite, Electron, Chromium, and Browser Host processes; none remained.

### Packaging

- `bun run desktop:pack`
  - built the Web app, Desktop main/preload, minimal Browser Host entry, managed runtime, Electron app directory, and completed local macOS ad-hoc signing

## Review guide

Recommended review order:

1. `packages/browser/src/protocol.ts` and `controller.ts`
2. `packages/synergy/src/browser/session.ts`, `command-service.ts`, `broker.ts`, and `network-gateway.ts`
3. `packages/synergy/src/server/browser-route.ts` and Browser permission classification
4. `packages/desktop/src/browser-host-*`, native page pool, and WebRTC host
5. App Browser workspace, generated SDK, release workflow, and documentation
6. Behavioral contract/security/runtime tests

## Checklist

- [x] Relevant narrow tests pass; commands are listed above
- [x] `bun run workflow:check` passes
- [x] SDK regenerated after route/schema changes
- [x] Documentation, AGENTS, PRODUCT, and `.synergy` development skill updated
- [x] No secrets, local auth files, unrelated cleanup, placeholder scripts, deprecated aliases, redundant compatibility wrappers, or temporary build/test artifacts included

### Final dev rebase and CI hardening

- Rebased the complete Browser change onto `dev` commit `9673cd8cf` after PR #546 merged; the final branch is not behind `origin/dev`.
- Final CI run [29122066678](https://github.com/SII-Holos/synergy/actions/runs/29122066678) passed all 8 jobs: Quality, Typecheck, Test + coverage, Package Validation, Workflow Validation, Secret Scan, Desktop Checks, and Smoke Test.
- Fixed Linux Electron screenshot hangs by making remote Host pages true offscreen renderers with background frame throttling disabled. The Ubuntu/Xvfb runtime smoke now exercises locator screenshot through the same Protocol v2 CDP controller without a platform skip or screenshot fallback.
- Aligned the runtime smoke request envelope with the production Host broker timeout instead of allowing the test harness to terminate CDP screenshot commands early.
- Removed a cross-file test race by assigning Browser event sequencing tests a unique owner and cleaning up only that owner; the invariant remains strict contiguous per-owner sequencing.
- Refreshed and frozen the lockfile after the latest dev rebase, removing the obsolete Desktop direct dependency entry left by lockfile merge.


yzxoi:
👋 We're reviewing this PR now — will analyze the design approach, look for bugs, and resolve any merge conflicts with `dev`.

yzxoi:
## Review Summary

✅ Merged `dev` (21 commits ahead / 5 behind) — resolved 8 merge conflicts across AGENTS.md, README.md, architecture docs, and Browser workspace files.

### Architecture Assessment

The new Browser subsystem is well-structured: unified Protocol v2, clean state machine (`empty→active→suspended/migrating/failed`), `CdpTransport` abstraction, Host broker ticket system, network gateway defense-in-depth, and command replay dedup. Test coverage is strong (304+ browser tests).

### Bug Found & Fixed

**Native Browser ticket owner does not match** — The frontend computes `ownerKey` using `params.dir` (URL route directory segment), while the server-side `routeState()` uses `ScopeContext.current.scope.id` (a scope hash). These are different values, so native ticket consumption always fails for Desktop users. Agent tools work fine because they use `scope.id` consistently.

**Fix** (2 files, +20/-2):

- `native-presentation.ts`: `consume()` computes an alternate `ownerKey` using the route directory as `scopeID`, accepting either match
- `browser-route.ts`: passes `c.req.param("directory")` to `consume()`

### Other Findings (non-blocking)

- `broker.ts`: page.create timeout cleanup closes the entire broker connection if `page.close` also times out — overly aggressive
- `controller.ts`: `timer.unref?.()` is a no-op in Bun (setTimeout returns number, not Timeout)
- `command-service.ts`: outer dedup check outside the serialized queue has confusing atomicity semantics
- `session.ts`: `saveTail` swallows persist errors silently

### Testing

- `bun test test/browser/event.test.ts` — 2 pass, 0 fail
- Pre-push gates: prettier ✓ oxlint ✓ typecheck (12 packages) ✓ sherif ✓

yzxoi:
## Review Summary (Updated)

✅ Merged `dev` (21 commits ahead / 5 behind) — resolved 8 merge conflicts across AGENTS.md, README.md, architecture docs, and Browser workspace files.

### Architecture Assessment

The new Browser subsystem is well-structured: unified Protocol v2, clean state machine (`empty→active→suspended/migrating/failed`), `CdpTransport` abstraction, Host broker ticket system, network gateway defense-in-depth, and command replay dedup. Test coverage is strong (304+ browser tests).

### Bugs Found & Fixed

**1. Native Browser ticket owner does not match** (critical, Desktop only)
Root cause: frontend computes `ownerKey` using `params.dir` (URL route directory segment), server-side `routeState()` uses `ScopeContext.current.scope.id` (a scope hash). These never match → native ticket consumption always fails for Desktop users. Agent tools work fine because they use `scope.id` consistently.

Fix: `native-presentation.ts` — `consume()` accepts an optional `routeDirectory` parameter, computes an alternate `ownerKey` using the route directory as `scopeID`. Either match passes. `browser-route.ts` — passes `c.req.param("directory")` through.

**2. Black browser panel on first open** (WebRTC Host startup timing)
When the user opens the browser panel before the WebRTC Host (Playwright/Chromium headless shell) is ready, the ticket endpoint returns 400 with `retryable: true`. The frontend was retrying every 1s and showing a generic `error` state, looking permanently broken.

Fix:

- `remote-browser-surface.tsx` — attach `retryable` flag from the SDK error response to the thrown Error
- `browser-webrtc.ts` — exponential backoff (1s → 2s → 4s → 8s → 16s → 30s cap), reset counter on success, show `host_pending` status for retryable errors instead of `error`

### Other Findings (non-blocking, to address in follow-up)

- `broker.ts`: `page.create` timeout cleanup closes entire broker connection if `page.close` also times out — overly aggressive
- `controller.ts`: `timer.unref?.()` is a no-op in Bun (`setTimeout` returns number, not Timeout object)
- `command-service.ts`: outer dedup check outside serialized queue has misleading atomicity semantics
- `session.ts`: `saveTail` swallows persist errors silently
- `migration.ts`: silently drops invalid annotations without logging

### Commits on this branch

```
14aae0b8 fix(browser): show pending state with backoff during WebRTC host startup
a979f6bb fix(browser): accept native ticket ownerKey from route directory
d6319335 Revert bad fix: use params.dir for ownerKey in BrowserPanel (reverted erroneous `sdk.scopeID` approach)
3fe9b721 chore: merge dev into pr-547-review, resolve conflicts
```

### Testing

- `bun test test/browser/event.test.ts` — 2 pass, 0 fail
- Pre-push gates: prettier ✓ oxlint ✓ typecheck (12 packages) ✓ sherif ✓

EricSanchezok:
Implemented the Browser runtime convergence and rebased this PR onto the latest `dev`.

The central architectural decision is now explicit: Synergy provides a real Chromium page plus the thinnest control/ownership layer. Chromium owns webpage-origin security, CORS, TLS, mixed content, Local Network Access, and renderer isolation. Synergy continues to protect the boundaries it actually owns: authenticated proxy access, owner isolation, page/profile lifecycle, workspace file containment, and Browser tool permissions.

## What changed

### Network path

- Removed the unpublished `browser_private_network` capability, grant propagation, DNS pre-resolution, IP-range classification, localhost port allowlists, and Fake-IP/private/metadata/benchmark/documentation address branches.
- The Gateway now validates only the proxy protocol, hostname syntax, and port, then passes the hostname to the system network stack. This makes TUN/Fake-IP, localhost, LAN, direct IP, redirects, and ordinary multi-resource pages follow the machine's actual routing.
- Kept loopback-only listening, random owner-scoped credentials, per-owner connection limits, revoke cleanup, and socket cleanup.
- Replaced O(n) owner credential scans with username-indexed lookup plus constant-time password verification.
- Fixed an additional root cause found during the managed smoke test: Chromium uses proxy challenge-response. The Gateway previously returned `403` for the initial unauthenticated CONNECT, so Chromium never submitted its owner credentials and surfaced `ERR_TUNNEL_CONNECTION_FAILED`. It now returns the standard `407 Proxy-Authenticate` challenge while invalid/missing/cross-owner credentials remain rejected.
- Electron Browser content sessions allow only Chromium's local-network/loopback-network permission names; camera, microphone, location, devices, and unrelated permissions remain denied. Web security is not disabled by Synergy.

### Canonical ownership and page lifecycle

- Added `ownerKey` to the Browser session-state API and regenerated the SDK.
- Frontend events, native presentation tickets, Desktop profiles, broker pages, and native attachment now use only the server-provided canonical owner. The route-directory compatibility owner was removed.
- Session bootstrap is passive. Empty and suspended sessions do not create a page. An existing active headless page sends one `resume`, migrates to the selected Host backend, and then attaches the native/WebRTC presentation.
- `hostStatus` is now computed for the current owner/page instead of treating a globally ready broker as proof that this page exists.
- Interactive surfaces wait for page-specific Host readiness before attaching.

### Reliability and errors

- Added one Browser SDK-boundary error normalizer and applied it consistently to session bootstrap, control, diagnostics, annotations, viewer tickets, and WebRTC. Structured `code/message/retryable` data is preserved; `[object Object]` is no longer shown.
- WebRTC retries only retryable ticket failures, signaling closure, and failed peer connections. Backoff resets only after Host/media readiness. Permanent protocol/host errors stop retrying, and stale tickets, sockets, peers, and timers are discarded.
- Fixed Browser page disposal ordering so CDP detaches before the Playwright page/context closes; shutdown no longer reports a false `Target page, context or browser has been closed` cleanup failure.

### Workspace files and checkpoints

- Explicit `file:` navigation is supported only for existing real paths inside the workspace.
- Traversal, escaping symlinks, hidden segments, `.git`, `.synergy`, and `node_modules` remain denied. Page-initiated file navigation is still blocked.
- Checkpoints now support valid workspace `file:` pages, revalidate containment before restore/migration, and tolerate documents where `localStorage` or `sessionStorage` throws `SecurityError` instead of failing an otherwise usable navigation.
- Fixed the shared path-containment helper so legitimate contained names beginning with `..` are not misclassified as parent traversal.

### Cleanup and documentation

- Removed obsolete Browser tool registry references and the duplicate Browser architecture document.
- Consolidated the current model in `docs/architecture/browser-runtime.md` and synchronized the product Browser doc, App product contract, package reference, Browser development skill, core/Desktop agent guides, and a new `packages/browser/AGENTS.md` ownership guide.
- Kept one page model, one protocol, one control path, and the two first-class presentation modes (`WebContentsView` native and WebRTC remote). No iframe, screenshot-stream, pseudo-tab, compatibility adapter, or fallback Browser implementation was added.

## Verification

Passed:

- `bun run quality:quick`
- `bun run deadcode`
- Browser package: 30/30 tests, typecheck, build
- App Browser workspace: 31/31 tests, typecheck
- Core Browser/enforcement/route/path regression set: 309/309 tests
- Desktop: 59/59 tests (1 environment-gated skip), typecheck
- Real Electron Browser Host runtime smoke: 1/1
- Managed Desktop startup with an isolated `SYNERGY_HOME`: app build, server startup, global runtime, and Browser Host broker registration all succeeded
- Real headless Browser command path: random localhost development port loaded, then the same page navigated to `https://www.google.com/` and returned title `Google`
- Real workspace `file:` navigation loaded a workspace file successfully
- Proxy authentication, cross-owner rejection, 64-connection limit, revoke socket closure, owner mismatch, passive empty/suspended bootstrap, active headless resume, WebRTC retry/stop behavior, structured errors, symlink/path containment, and storage-denied checkpoint capture are covered by regression tests

I also ran the full Turbo test matrix twice. All packages and all Browser-related tests passed. The second run reached 4,724 passing / 1 skipped, but the nine `daemon.spec` cases hit the repository test harness's shared temporary SQLite database with `SQLITE_IOERR_READ`; running that complete file immediately in isolation passed 9/9. The first full run showed the same SQLite failure pattern plus one timing-sensitive summary timeout, whose file also passed in isolation. I did not hide or bypass those full-suite environmental failures.

## Manual reviewer flow

1. Run `bun dev desktop --managed`.
2. Open Browser without navigating: no page should be created.
3. Enter a localhost frontend/backend URL on any valid port, then navigate to Google; both should load.
4. Ask the agent to call Browser `goto`; it should reuse the same session page without a private-network/Fake-IP denial.
5. Let the agent create a headless page first, then open Browser; the same page should migrate once into native presentation.
6. Try a workspace file and an escaping symlink/hidden project directory; only the contained ordinary file should open.

The updated commit is `e6bd34d63`.

EricSanchezok:
CI follow-up: the complete GitHub Actions matrix is now green, including the full package test run and Synergy coverage. This confirms the local `SQLITE_IOERR_READ` failures documented above were isolated to the local shared temporary test environment.

EricSanchezok:
## Final rebase and upgrade-safety follow-up

This PR has now been rebased onto the current `dev` plus the dedicated upgrade fix in #555. The rebase completed without conflicts, and `git range-diff` matched all 10 Browser commits one-for-one: no Browser behavior was dropped or silently rewritten by the rebase.

### Why #555 is part of the current base

The latest `dev` observability migration referenced both `trace_id` and `traceId`, while released user databases contain only the canonical `trace_id` column. That caused managed startup to exit before health became ready. The dedicated fix keeps legacy handling inside Synergy migration boundaries:

- detects the released snake-case schema and the known camel-case variant at the migration boundary;
- streams legacy telemetry and redacts canonical rows in bounded batches;
- writes schema version 4 through a new central migration;
- marks `ensureMigrations()` complete only after success, allowing same-process retry after transient failure;
- adds released-schema, rollback, idempotence, batching, metadata, and retry regression coverage.

#555 is intentionally a separate PR and its CI is 8/8 green. It should merge before #547; once it lands in `dev`, the shared migration commit automatically disappears from this PR diff.

### Final verification after rebase

- `bun run quality:quick` passed.
- Browser package: 30/30 passed.
- App Browser workspace: 31/31 passed.
- Desktop: 59/59 passed, with 1 existing environment-gated Electron Host skip.
- Core Browser, server route, observability, and migration set: 144/144 passed.
- Isolated upgraded-state server startup reported all 10 migration domains current and `/global/health` returned healthy.
- GitHub Actions for this rebased head is 8/8 green: Quality, Test + coverage, Typecheck, Desktop, Package Validation, Workflow Validation, Secret Scan, and Smoke Test.

The rebased head is `231436af7`.

EricSanchezok:
Follow-up from managed Desktop validation:

The reported 400 was not a network failure. The native Chromium page had already loaded Google, but a newly created Electron WebContentsView still had a 0×0 viewport. The control route captures a checkpoint before returning navigation success, and the checkpoint correctly rejects width/height below 1. That left the frontend waiting for a native surface even though the backend page was active.

Commit e4998d6a5 fixes the lifecycle invariant by assigning a 1280×720 viewport before the first load. It also preserves the attached view x/y when later viewport commands change width/height, preventing the native page from jumping to the app window origin.

Added regression coverage for both cases. Verified Desktop typecheck, all Desktop tests (61 pass, 1 expected skip), the real Electron Browser Host runtime contract, and the full repository quality:quick gate.

EricSanchezok:
Second managed Desktop follow-up (commit 6e737cbff):

The first viewport fix exposed two distinct lifecycle gaps:

1. Electron does not give an unattached WebContentsView a non-zero CSS viewport from setBounds alone. Native page creation now initializes Chromium through the shared setViewport/CDP command before the first navigation, so checkpoint capture sees a real 1280×720 viewport.
2. Successful Host page creation and headless-to-Host migration emitted page.created/page.updated but no page-scoped host.status ready event. The canonical Browser session now publishes ready after the Host page becomes canonical, allowing the native surface to attach immediately instead of only after a reload/session snapshot.

I verified this with a fresh isolated managed Desktop instance through the real renderer + preload IPC + native ticket path: opened Browser from a new session, entered Google in the address bar, observed both control requests return 200, confirmed the waiting state disappeared without reload, inspected the attached native WebContentsView at the panel-sized CSS viewport, and confirmed Google reached document.readyState=complete with title/content present.

Checks:
- Core Browser: 56 pass
- App Browser: 31 pass
- Desktop: 61 pass, 1 expected skip
- Electron Browser Host runtime contract: pass
- bun run quality:quick: pass

EricSanchezok:
Implemented and pushed the full Browser review follow-up in `a1c3d9f28`.

### What changed

- Kept the runtime contract at one canonical page per owner while making Desktop native and WebRTC peer presentation paths.
- Fixed page-scoped Host readiness so both timing orders work:
  - open Browser workspace, then navigate;
  - create/use a Browser page through tools, then open the workspace.
- Hardened native `WebContentsView` lifecycle:
  - non-zero initial viewport;
  - valid rounded attach/resize bounds;
  - viewport resize preserves the Desktop-assigned presentation origin;
  - asynchronous idempotent close, dead-page cleanup, and same-owner recreation;
  - background rendering remains active while detached.
- Removed native interaction stalls by using Electron input for mouse move/wheel while retaining the shared CDP controller for canonical actions.
- Improved controller reliability and atomicity:
  - bounded actionability stability checks outside renderer animation frames;
  - deterministic nearest-container scrolling;
  - select validates value/label/index before changing the page;
  - readonly side-effect failures have a specific structured error.
- Improved Agent-facing Browser tools:
  - precise locator, select, scroll, close, console, and network guidance;
  - `includeSnapshot` returns the resulting snapshot and accessibility tree directly;
  - structured action summaries and errors;
  - screenshots enter visual context for image-capable models, while text-only models receive a real asset path and `look_at` guidance.
- Updated the Browser architecture, Web product contract, and Browser runtime development skill.
- Removed overlapping readiness coverage and retained focused creation/migration contracts.

### Validation

- `packages/browser`: 33 passed
- Browser workspace UI: 32 passed
- Core Browser routes/runtime/tools: 58 passed
- Desktop unit tests: 61 passed
- Real Electron native/Host runtime: 2 passed
- `bun run quality:quick`
- `bun run deadcode`
- pre-push format, lint, typecheck, and monorepo checks

The real Electron regression covers native viewport/origin, hover, click and modal opening, targeted scrolling, console retrieval, checkpoint capture, close, and same-owner recreation.

### Manual verification

- Google loads successfully through the system/TUN/Fake-IP environment.
- An Agent-created page appears in the already-open Browser workspace.
- A tool-created page is visible when the Browser workspace is opened later.
- First address-bar navigation from an empty workspace remains inside the right Browser panel; viewport fitting no longer moves the native page to the window origin.
- Screenshots remain available while the user sees the same live Chromium page in the workspace.

Issue #523, opened 2026-07-10
Add an Agent Organization Runtime with Boss control plane, primary sessions, and governed A2A


marzich:
## Problem or motivation

Synergy already provides strong primitives for long-running agent work: primary sessions, subagents, DAGs, BlueprintLoop, Lattice, worktrees, notes, memory, session control, and `session_send` / `session_read`.

However, when a project becomes large enough to require many stages, repeated parallelism, review/rework loops, and long-lived context, the bottleneck shifts from individual agent capability to **organizational complexity**.

Today, a human may need to:

- remember what each session is responsible for;
- find and inspect many sessions one by one;
- manually route messages and artifacts between sessions;
- audit whether sessions with the same prompt actually followed the same process;
- coordinate several waves of parallel investigation, implementation, review, and testing;
- recover role boundaries and workflow state after context compaction;
- distinguish durable team members from temporary review/investigation sessions;
- decide when a workflow should continue, pause, escalate, or stop.

This does not scale for workloads such as continuously resolving Synergy issues. The problem is no longer simply “run several subagents in parallel.” It is:

> How can Synergy run a persistent, auditable organization of agents while the human interacts with only one control-plane session?

## Proposed principle

Add a first-class **Agent Organization Runtime** built around the following hierarchy:

```text
Human
  │
  ▼
Boss Session
  Human-facing control plane
  │
  ▼
WorkflowRun / Organization Runtime
  Durable automated state machine
  │
  ├── Primary Sessions (persistent employee agents)
  │     ├── investigation / issue research
  │     ├── implementation
  │     ├── PR review
  │     ├── integration test
  │     └── release / operations
  │
  └── External Engagements (temporary contractor sessions)
        ├── independent third-party review
        ├── isolated investigation
        ├── security/performance audit
        └── one-shot verification
```

The design should follow four rules:

1. **Human authority is centralized in the Boss Session.**
2. **Workflow authority is centralized in a durable state machine.**
3. **Execution is distributed across persistent Primary Sessions.**
4. **Temporary expertise is delegated to External Sessions without making them permanent organization members.**

A concise formulation is:

> Centralized governance, distributed execution; shared state, peer coordination; temporary delegation, durable accountability.

## Boss Session: human control plane

The human should interact only with the Boss Session.

The Boss Session should:

- accept the human goal and start an organization/workflow run;
- observe progress, health, blockers, costs, and risk;
- request pause/resume/cancel operations;
- receive escalations and human checkpoints;
[...]

Issue #512, opened 2026-07-09
Add Super Loop orchestration for linked multi-session workflows


marzich:
## Problem

Synergy has strong primitives for multi-session work (`session_send`, session control, DAGs, notes, subagents, long-running sessions), but it lacks a first-class orchestration layer that binds many sessions into a persistent role-based workflow.

In the Clarus Runtime V1 workflow we have these roles:

```text
Boss / Coordinator session
Execution sessions
Dedicated per-PR review sessions
Final test/merge session
Runtime testing session
```

The intended flow is:

```text
Coordinator defines graph/policy/state
  -> Execution Session implements issue and opens PR
  -> Dedicated PR Review Session reviews PR
  -> Final Test/Merge Session handles final gate
  -> User final checkpoint
```

But today the coordinator session often becomes a manual message relay. For example, after a dedicated PR review session passes a PR, the review session should send the final test/merge handoff directly to the final test/merge session and only notify the coordinator. Instead, without workflow-level routing, the coordinator may manually copy/paste the handoff.

This creates several failure modes:

1. Execution sessions can stop silently at blockers (dirty checkout, missing decision, ambiguous scope) unless the coordinator manually inspects every session.
2. Execution sessions can open PRs and produce handoff notes but fail to create/send the dedicated PR review session.
3. PR review sessions can pass PRs but fail to route directly to the final test/merge session.
4. The coordinator becomes the default relay for artifacts instead of owning only graph, policy, state, and exceptions.
5. Context compaction can break the workflow because sessions may forget the next required downstream action.
6. There is no shared dashboard showing which issue is executing, which PR is in review, which PR is queued for final merge, and which blocker needs coordinator/user decision.

## Desired Principle

```text
Coordinator owns the graph, not every message.
```

Each role should own its downstream handoff:

```text
execution.PR_OPENED      -> create/send dedicated PR review session; notify coordinator
pr_review.REVIEW_BLOCKED -> send rework to linked execution session; notify coordinator
pr_review.REVIEW_PASSED  -> send final handoff directly to final merge session; notify coordinator
final_merge.MERGED/HELD  -> notify coordinator/user
```

## Proposed Solution

Add a first-class **Super Loop / Workflow Run** system for linked multi-session workflows.

A workflow run should define:

- coordinator / boss session;
- role definitions (`execution`, `pr_review`, `final_merge`, `runtime_test`, etc.);
- session bindings;
- structured events;
- transition routing rules;
- prompt templates;
[...]

PR #549, opened 2026-07-11
docs: rewrite product and architecture documentation

yzxoi:
## Summary

- Reorganize documentation around canonical product, architecture, reference, plugin, operations, migration, and research ownership, with a new documentation index.
- Rewrite repository entry points and subsystem guides to match the current Scope/session model, Cortex and workflow lifecycle, Browser and File workbenches, provider recovery, Holos connectivity, and plugin contracts.
- Refresh repository commands and development Skills, including focused workflows for frontend work, LLM integration, server APIs, persistence, testing, runtime isolation, and safe worktree/PR publication.

## Why

Core product and runtime behavior has evolved across several iterations, leaving overlapping historical documents and executable repository guidance out of sync. This change replaces that drift with current-state documents and removes obsolete duplicates.

## Validation

- `bun run quality:quick`
- Pre-push format, lint, typecheck, and monorepo checks
- Structural validation for the updated repository Skills
- Equivalent assertions for the relocated optional performance-tooling documentation

The existing `bun test script/performance-tools.test.ts` entry remains blocked by the root Bun test configuration, which points root test discovery at a disabled directory; the same file and documentation invariants were executed directly and passed.


marzich (changes requested):
Overall, this is a strong rewrite of Synergy's current-state documentation.

The separation between product, architecture, reference, operations, plugins, migrations, and executable repository Skills is substantially clearer. The new Session/Message, Cortex, Scope, Workflow, Inbox, and Compaction documents also establish much more precise ownership boundaries than the previous documentation.

I agree with keeping future features such as #512 and #523 out of the canonical product and architecture documents until those features are implemented. My review is therefore focused on current-state correctness, internal consistency, preservation of useful material, and whether the documented ownership boundaries leave a clean extension point for future architecture.

## Blocking: the research ownership destination does not exist

`docs/README.md` defines `docs/research/` as the canonical home for investigations, measurements, and historical design material, and links to `docs/research/README.md`. That path does not exist in this branch.

At the same time, this PR deletes the KV-cache research documents and `docs/architecture/runtime-performance.md` without relocating or summarizing them under the new research ownership boundary.

Could we either:

1. add `docs/research/README.md` and migrate or summarize the still-useful research material there; or
2. remove the broken research link for now and explicitly track the research migration as follow-up work?

As written, the new documentation ownership contract points to a missing destination while existing research material is removed. The material does not necessarily need to remain in its original long form, but the PR should make an explicit preservation/archive decision rather than leave the new canonical destination missing.

## Non-blocking: clarify that the workflow documents are session-local

The workflow documents accurately describe the currently implemented Plan, BlueprintLoop, Light Loop, and Lattice behavior. These are all workflows owned by or bound to one session.

It may be useful to state that scope explicitly near the beginning of:

- `docs/product/workflows.md`
- `docs/architecture/workflows.md`

For example:

> This document covers workflows owned and executed by a single session.

Or, more precisely:

> This document covers session-local workflow modes. Plan, Light Loop, and Lattice are owned by one session, while BlueprintLoop is a separate execution lifecycle bound to a session.

This remains a current-state statement rather than a roadmap statement, while making the ownership boundary explicit. It would also prevent the current `Workflow Runtime` wording from being interpreted as a claim that all future orchestration must fit inside `session.workflow`.

## Non-blocking: qualify workflow-state ownership in the architecture overview

`docs/architecture/README.md` currently says that a durable session owns “workflow state.” Since the implementation documented here is specifically session-owned workflow state, consider changing this to “session-local workflow state.”

That wording is more precise for the current implementation and leaves a clean extension point if Synergy later adds a separate cross-session orchestration layer without changing the existing session workflow contract.

## Future compatibility assessment

With the ownership scope made explicit, the current architecture described here appears compatible with future higher-level orchestration:

- DAG remains session-local coordination state rather than a durable outer workflow.
- Cortex remains bounded parent-child delegation with explicit output and lifecycle contracts.
- SessionInbox remains the delivery and scheduling substrate through `task`, `steer`, and `context` modes.
- Plan, Light Loop, Lattice, and BlueprintLoop remain session-local execution lifecycles.
- Compaction remains a recoverable context boundary with durable history and a `session.compacted` event.
- A future cross-session runtime can therefore be added above sessions rather than being forced into the existing `session.workflow` union.

No future Boss Session, organization runtime, governed A2A, or cross-session state-machine behavior needs to be documented in this PR before it exists. The important requirement here is that the current-state documents describe their ownership scope precisely enough that those future layers can be added without contradicting the contracts established now.

## Terminology note for future work

The rewritten current-state docs also establish two existing meanings that future design work should preserve:

- `primary agent` is an agent catalog/execution mode;
- `external agent` means a local adapter such as Codex, Claude Code, or OpenClaw.

Future persistent organization members and temporary contractor sessions should therefore use distinct names rather than overloading those current terms. This does not require a change in this PR, but the terminology established here should be treated as part of the compatibility boundary.

Once the missing research destination is resolved, I think this documentation structure provides a solid current-state foundation and a clean layering model for future architecture.

EricSanchezok (approved):
Sehr gut. Es gibt jedoch einige Punkte, die nach der Optimierung eine Überarbeitung der Dokumentation erfordern:

1. Der Pull Request https://github.com/SII-Holos/synergy/pull/547  hat den Browser umfassend überarbeitet. Nach der Validierung müssen auch die browserbezogenen Dokumente entsprechend angepasst werden.

2. Lokal habe ich ebenfalls ein umfassendes Upgrade für das Plugin durchgeführt. Sobald ich diesen Teil auf den `dev`-Branch rebase, werde ich die entsprechenden Dokumente aktualisieren.

3. Was den Test-Code betrifft, so gibt es derzeit viele vereinzelte Testdateien, die direkt neben dem Code verteilt sind. Ordnungsgemäße Testdateien sollten im entsprechenden `tests/`-Verzeichnis abgelegt werden. Nachdem die Tests insgesamt neu strukturiert wurden, müssen auch die betreffenden Dokumente angepasst werden.

4. Die Quellen für das Frontend-Theme sind derzeit nicht einheitlich. Dieser Punkt sollte so schnell wie möglich bereinigt und vereinheitlicht werden, um zu vermeiden, dass die zukünftige Frontend-Entwicklung ohne klare Richtlinien bleibt und zunehmend unübersichtlich wird. Nach der Vereinheitlichung ist auch die dazugehörige Dokumentation zu aktualisieren. Der Maßstab für die Vereinheitlichung lautet: **Alle Color Tokens werden im Theme zusammengeführt (konvergiert) und lassen sich problemlos um weitere Themes neben Synergy erweitern, wobei die Gesamtoptik ansprechend bleibt und keine Farbtoken übersehen werden.**

PR #571, opened 2026-07-12
fix: publish cortex completion before parent wake

EricSanchezok:
## Summary
- enqueue Cortex completion notices without synchronously processing the parent turn
- publish the terminal in-memory task before the parent can read task output
- add a regression test covering the idle-parent completion wake ordering
- sync generated LearningConfig timeout and output-limit contracts

## Verification
- `bun test test/cortex` (113 passed)
- `bun test test/session/manager.test.ts` (14 passed)
- `bun run quality:quick` (passed)
- `bun run test:changed` selected 2,041 tests: 2,026 passed, 1 skipped, and 14 failed under broad Windows parallel resource contention; the affected Cortex and Session Manager suites passed independently

### DIFF
--- a/packages/sdk/js/src/gen/types.gen.ts
+++ b/packages/sdk/js/src/gen/types.gen.ts
@@ -2098,6 +2098,14 @@ export type LearningConfig = {
    */
   encoderRetries?: number
   /**
+   * Wall-clock deadline for a single encoder LLM call in milliseconds (default: 60000)
+   */
+  encoderTimeoutMs?: number
+  /**
+   * Maximum characters collected from one encoder model stream before abort (default: 16000)
+   */
+  encoderMaxOutputChars?: number
+  /**
    * Max estimated tokens for tool output in turn digest (default: 800)
    */
   digestToolOutputBudget?: number
--- a/packages/sdk/openapi.json
+++ b/packages/sdk/openapi.json
@@ -21163,8 +21163,7 @@
             "type": "string"
           }
         },
-        "required": ["id"],
-        "additionalProperties": false
+        "required": ["id"]
       },
       "ChannelInfo": {
         "type": "object",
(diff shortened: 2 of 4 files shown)


yzxoi:
Initial review: the direction of this fix is correct, and the new idle-parent regression covers the original self-wake ordering. However, the completion boundary is still not atomic: the Cortex completion notice is written and published to the parent inbox before the terminal task is exposed in the in-memory task map. If the parent starts concurrently from another trigger, it can still observe `Status: running` after the completion notice exists.

We are working on a follow-up fix that makes the terminal task visible before any completion notice can be consumed, together with a regression test for the concurrent-parent-start path. We will push the additional commit to this PR once the focused and repository quality checks pass.

yzxoi:
Follow-up fix pushed in `19e6c231`.

The finalization path now registers the parent notification as pending before publication, exposes the terminal task and result to task readers, and only then flushes the completion notice when the parent is idle. This preserves the release-time no-lost-wake guarantee while preventing inbox observers or concurrently starting parent turns from seeing stale `running` output.

The added regression subscribes to `session.inbox.updated`, starts the parent at the completion-notice boundary, and verifies that `task_output` already reports `completed` with the final result.

Local verification:

- `bun test test/cortex` — 114 passed
- `bun test test/session/manager.test.ts` — 14 passed

The PR CI checks are now running on the updated head.

yzxoi (approved):
LGTM. I verified the updated completion-ordering flow with a real end-to-end scenario: the parent observes the terminal task and final output before processing the completion notice. The concurrent-start regression now covers the race, and the focused Cortex, SessionManager, and repository quality checks pass.

Thank you for the substantial effort on this PR and for carefully working through the Cortex lifecycle, notification ordering, and regression coverage. The resulting implementation is clearer and more robust. Approved.

PR #605, opened 2026-07-14
fix(runtime): share local production memory mitigations

marzich:
## Purpose

This is a review-only bundle of the production patches currently carried locally. It is **not a merge request**. The branch intentionally preserves the deployed patch stack so maintainers can inspect the mitigations and decide which ideas belong upstream.

The stack was developed against an earlier `dev` revision and may conflict with current `dev`; resolving those conflicts is intentionally out of scope for this reference PR.

## Patch areas

| Area | Local mitigation |
|------|------------------|
| Session streaming | Release completed-turn prompt/tool references, cancel the unused AI SDK stream branch, and detach per-turn abort resources |
| Message history | Add per-session message-cache bounds and preserve the cache across detached prune jobs |
| Processor lifecycle | Clear tool execution bookkeeping and streamed state after settlement |
| Hashline | Bound retained session snapshots and prune stale snapshot state |
| Optional background work | Skip experience encoding and reap idle LSP clients under memory pressure |
| Attribution | Record process-family RSS/PSS so main-process and child-process growth can be separated |
| Transport | Tighten provider cancellation, event backpressure, and slow-client handling |
| Recovery | Stop failed inbox wake loops from repeatedly recreating session work |
| Linux runtime | Preserve required sandbox environment and local helper compatibility |

## Verification

- `bun test test/session/invoke.test.ts test/session/manager.test.ts test/session/processor.test.ts` — 71 passed, 0 failed
- `git diff --check` — passed
- No stress test or live delegated task was run

## Review guidance

Please treat each commit as an independent mitigation proposal. The useful outcome of this PR is feedback on ownership and upstream direction, not merging the branch as-is.

### DIFF
--- a/packages/synergy/src/hashline/snapshots.ts
+++ b/packages/synergy/src/hashline/snapshots.ts
@@ -83,6 +83,18 @@ export class InMemorySnapshotStore extends SnapshotStore {
     return hash
   }
 
+  stats(): { paths: number; versions: number; totalBytes: number } {
+    let paths = 0
+    let versions = 0
+    let totalBytes = 0
+    for (const [, history] of this.#versions.entries()) {
+      paths++
+      versions += history.length
+      for (const version of history) totalBytes += version.text.length
+    }
+    return { paths, versions, totalBytes }
+  }
+
   recordSeenLines(path: string, hash: string, lines: Iterable<number>): void {
     const version = this.#versions.get(path)?.find((snapshot) => snapshot.hash === hash)
     if (version) mergeSeenLines(version, lines)
--- a/packages/synergy/src/hashline/store-session.ts
+++ b/packages/synergy/src/hashline/store-session.ts
@@ -3,43 +3,144 @@
  * Wraps InMemorySnapshotStore with ScopedState + Bus SessionEvent.Deleted cleanup.
  */
 import { Bus } from "../bus"
-import { ScopeContext } from "../scope/context"
 import { ScopedState } from "../scope/scoped-state"
 import { SessionEvent } from "../session/event"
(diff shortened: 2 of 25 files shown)


yzxoi:
Thank you for sharing this production patch stack and, especially, for documenting the runtime evidence behind it. The separation into independent mitigations made it much easier to compare the ideas with the current `dev` implementation.

After reviewing the commits, we found that several of the core ideas have since landed upstream through owner-level implementations, including model-stream ownership and residual-branch cleanup, per-turn abort cleanup, prune/history-cache reuse, bounded provider/tool input, permission telemetry bounds, idle LSP/process-tree cleanup, child-process memory attribution, and slow-client backpressure handling.

We do not plan to merge this draft branch as a bundle, both because it targets an older `dev` revision and because some local deployment changes—such as provider-specific sandbox environment passthrough and a machine-specific Linux helper hash—should not become upstream defaults.

There are two proposals here that we do want to carry forward in a focused follow-up PR:

1. Make Experience re-encoding pause under memory pressure and resume once pressure subsides, rather than permanently skipping work.
2. Break the failed inbox wake cycle so a failed session run leaves its durable inbox item recoverable without immediately scheduling the same failing work again.

For the Experience path, we will reuse the current canonical memory-pressure machinery and preserve resumability instead of adopting fixed deployment-specific thresholds or a permanent skip. Other re-encode improvements such as grouping and request-level scheduling are being handled separately and will not be included in this follow-up.

Thank you again for contributing the production findings and reference patches. Even where the exact patch is no longer the right upstream shape, the evidence and failure modes were useful in identifying the remaining work.

PR #611, opened 2026-07-14
fix(workflows): coordinate Agenda-owned loop wakeups

EricSanchezok:
## Summary

- suppress shared Light Loop, BlueprintLoop, and Lattice continuation while wake-capable Agenda items own the session cadence
- add loop-aware Agenda delivery guidance, stop-tool blockers, and safe continuation recovery after the final Agenda blocker is released
- document the workflow contract and add regression coverage for project/global Agenda items, cleanup guidance, stop guards, and delivery-before-resume ordering

## Verification

- `cd packages/synergy && bun test test/agenda/ test/session/continuation-kernel.test.ts test/session/inbox.test.ts test/session/light-loop-continuation.test.ts test/session/blueprint-continuation.test.ts test/tool/loop-stop.test.ts test/tool/blueprint-loop-tools.test.ts` — 107 passed, 0 failed
- `bun run quality:quick`
- `git diff --check`


yzxoi:
I reviewed the current approach against Synergy’s session lease, persistent Inbox, Cortex completion delivery, and workflow continuation invariants. I don’t think the Agenda-specific blocker plus `ContinuationKernel.kick()` is safe to merge as-is.

The main issue is architectural: the implementation infers continuation ownership from static Agenda item fields (`active`/`pending`, `wake`, `silent`, and the original session ID), while actual delivery can target a rebound endpoint and can still wake the session through `SessionManager.deliver()`. It also creates a second continuation trigger alongside the asynchronous `SessionEvent.Idle` path. This produces concrete failures including:

- `wake=false` items still waking the session while bypassing the blocker;
- endpoint rebinding blocking the archived origin session but waking the new session;
- triggerless or expired Agenda items permanently blocking workflow continuation;
- concurrent Idle evaluation and `kick()` delivering duplicate continuations because deduplication is not atomic;
- cancellation of an in-flight Agenda releasing ordinary continuation before the Agenda result is delivered;
- post-restart `kick()` running before continuation policies have been initialized.

I’m going to rework the PR around the existing Synergy ownership boundaries instead of adding more Agenda-specific guards:

1. Keep `SessionManager` as the single session-loop owner and use durable `SessionInbox` as the only ingress for external results.
2. Add a thin, per-session single-flight drive/arbitration step. It processes runnable Inbox work first, then explicit external waits, and only then asks workflow policies for a synthetic continuation.
3. Make Cortex completion notifications durable and Inbox-backed instead of relying on the in-memory deferred notification map.
4. Give Agenda explicit wait and run-delivery semantics so a wait is released only after its result is durably enqueued or explicitly cancelled; ordinary recurring schedules will not implicitly block workflow continuation.
5. Keep the event bus observational rather than correctness-critical, and add stable delivery identities for retry/restart idempotency.
6. Update the workflow/Cortex/Agenda architecture docs and add coverage for endpoint rebinding, passive wake, in-flight cancellation, restart recovery, duplicate arbitration, and non-fireable schedules.

I’ll push the revised implementation and tests to this PR branch once the focused and repository quality gates pass.

yzxoi (dismissed):

yzxoi (approved):

PR #646, opened 2026-07-16
feat(ui): animate active compaction cards

EricSanchezok:
## Summary

- replace the active compaction card’s small icon opacity pulse with a low-contrast diagonal shimmer across the card surface
- scope the shimmer, stacking context, and compositor hint to `data-status="running"` so completed historical cards remain static and lightweight
- preserve interaction layering and disable the shimmer under `prefers-reduced-motion`
- add a production-build CSS contract proving the running shimmer and reduced-motion override survive bundling and minification

## Implementation details

The effect is implemented entirely in `packages/ui/src/components/compaction-card.css` through a `::before` pseudo-element. It uses only `transform` and `opacity`, does not add DOM or JavaScript timers, and keeps the existing compaction status/state flow unchanged. The old `compaction-card-running` icon keyframe and its only reference were removed to avoid redundant concurrent motion.

The shimmer uses existing semantic theme tokens for its low-contrast gradient and is clipped by the card’s existing overflow boundary. The content wrapper is raised above the effect only while the card is running, and `pointer-events: none` ensures the animation cannot interfere with the collapsible trigger.

## Accessibility and performance

- `prefers-reduced-motion: reduce` changes the shimmer’s computed animation to `none`
- the active effect animates compositor-friendly `transform` and `opacity` only
- `will-change` and the stacking context exist only during the running state
- completed compaction cards do not retain animation layers or additional stacking contexts

## Verification

- red/green production build CSS contract: confirmed failure before implementation, then passing after implementation
- `bun run --cwd packages/ui test` — 294 tests passed
- `bun run --cwd packages/ui typecheck` — passed
- `bun run --cwd packages/app test` — 614 tests passed, 5 existing skips
- `bun run --cwd packages/app typecheck` — passed
- `bun run --cwd packages/app build` — passed
- `bun run lint` — passed with 0 warnings and 0 errors
- `bun run typecheck` — 11/11 package tasks passed
- `bun run monorepo:check` — passed
- `bun run package:check` — passed
- `TMPDIR=/private/tmp bun turbo test --env-mode=loose` — 14/14 tasks passed; core suite reported 4,453 passed and 1 existing platform-dependent skip
- independent maintainability review — passed with no blockers; confirmed no dead keyframes or references
- independent performance review — no blockers; confirmed running-only compositor animation
- browser visual inspection — normal mode reports `compaction-card-shimmer` at 2.8s; dark reduced-motion reports `animation-name: none`

## Notes

The macOS default `/var` symlink caused an existing plugin-kit temporary fixture to resolve across `/var` and `/private/var` during the full test graph. Running the same graph with canonical `/private/tmp` and Turbo loose environment propagation passed completely. This does not affect Linux CI.


yzxoi:
I reviewed the shimmer change against the compaction state lifecycle and the production CSS contract. The stacking and interaction design itself looks sound, but I found two issues I’m going to address in a follow-up commit:

1. **Completed mechanical fallbacks can remain in the `running` presentation forever.** The fallback path writes an earlier `compaction_recovery` part with `validated: false`, then writes another recovery part later. Parts are sorted by ascending ID and the frontend selects the first recovery part, so the completed card still resolves to `data-status="running"`. This PR turns that stale state into a full-card infinite shimmer and persistent compositor hint.
2. **The reduced-motion build assertion can pass without a reduced-motion media query.** It joins every matching selector body and only checks that `animation:none` appears somewhere. A second ordinary root rule would satisfy the test just as well, so the test does not prove that the override remains under `@media (prefers-reduced-motion: reduce)`.

I’ll also reset `will-change` when reduced motion disables the animation, and tighten the production contract around the keyframes/selector handling. I’ll follow up here with the commit and verification results.

yzxoi:
Correction after fast-forwarding this worktree to the exact PR head: the current compaction implementation already uses `usedMechanicalFallback` and does **not** write a second `compaction_recovery` part on that path. My earlier explanation mixed the pre-fix dependency state into the PR-head review.

The user-visible bug still stands, with a simpler cause: a successful mechanical fallback commits the assistant message as completed while its single recovery part intentionally remains `validated: false`; `CompactionCard` currently treats `validated === true` as the definition of completion. The completed fallback therefore remains `data-status="running"` and receives the new infinite full-card shimmer.

I’m fixing the presentation state to derive lifecycle completion from the completed message plus the presence of a recovery part, while keeping `validated`/`mechanical` as summary-quality metadata. The reduced-motion contract issue described above is unchanged.

yzxoi:
Implemented the follow-up fix in this PR.

### Changes

- Derive compaction-card lifecycle from the presence of recovery data plus terminal message state, rather than from `validated`, so completed mechanical fallbacks no longer retain the running shimmer.
- Preserve `validated: false` as summary-quality metadata and keep the existing mechanical-fallback warning behavior.
- Reset `will-change` under reduced motion.
- Strengthen the production CSS contract to distinguish root animation rules from reduced-motion overrides and verify the shimmer keyframes.

### Verification

- UI component suite: 297 passed
- Compaction-card presentation model: 3 passed
- Compaction lifecycle regressions: 10 passed
- Production CSS contract: 2 passed
- Repository TypeScript typecheck: passed
- Skill/package guide, monorepo, and package checks: passed
- Changed-file Prettier and oxlint checks: passed

Commit: 94d293d5

PR #653, opened 2026-07-16
fix(app): prevent parent session Back loops

EricSanchezok:
## Summary

- distinguish ordinary session opens from hierarchical returns to a parent session
- restore a verified parent history entry, otherwise replace the child route without creating a Back loop
- cover both parent-return controls with focused navigation tests and document the product invariant

## Verification

- `bun test src/composables/use-navigate-to-session.test.ts src/components/session/prompt-dock-model.test.ts`
- `bun run format:check`
- `bun run quality:quick`

## Manual check

- launched an isolated Web runtime on alternate ports and confirmed the app loaded without change-related console errors
- parent/child fixture creation was unavailable under the active autonomous policy, so exact history branches are covered by the deterministic navigation tests


yzxoi:
Thanks for the PR. I pulled the branch into an isolated worktree and found a few navigation-history gaps that are not covered by the current helper tests:

1. `from` is captured from the full browser pathname, while session targets are Router-relative. With a non-empty Router base/proxy prefix, the parent comparison never matches and parent returns always fall back to `replace`.
2. Session scope correction and message-anchor `replaceState` calls can discard the navigation state used to verify the parent history entry.
3. Common parent → child entry points, especially the Subagent Dock and StatusBar child selector, bypass the session navigation helper, so returning to the parent can leave duplicate parent entries in browser history.
4. When resolving an unloaded parent through the async session lookup, the response can arrive after the user has navigated elsewhere and apply a stale `navigate(-1)`.

I’m going to update this PR with a focused follow-up that:

- uses one canonical Router-relative representation for stored and compared paths, including proxy-prefix deployments;
- routes hierarchical child-session entry points through the shared session navigation owner;
- preserves session navigation metadata across canonical-route and hash updates;
- adds a freshness guard for async session resolution;
- adds integration-oriented regression coverage for the real parent → child → parent history flow, proxy bases, state preservation, and stale async responses.

I’ll keep the existing intent-based API and `replace` fallback, since that part of the design is sound, then push the fixes and verification results to this PR.

yzxoi:
Implemented and pushed the follow-up fix in `31ed9b39`.

### What changed

- Canonicalize session routes against the stable proxy base before storing or comparing Back targets.
- Preserve Solid Router history state through scope correction and message-hash URL replacements.
- Route Subagent Dock and StatusBar child-session entry through the shared session navigation owner.
- Ignore stale async session lookups when the route, history entry, or newer navigation request has changed.
- Add focused coverage for proxy-base round trips, history-state preservation, and async navigation freshness.

### Verification

- `bun run quality:quick`
- PR CI: all checks pass.

The Test job initially hit two unrelated backend fixture flakes on separate attempts (`Snapshot.track()` returning no snapshot, then Git fixture init exiting 141). Both focused tests passed three consecutive local runs; the final CI rerun is green.

yzxoi (approved):

PR #656, opened 2026-07-16
feat(frontend): generalize session transition progress

EricSanchezok:
## Summary

- replace the worktree-only progress card with a reusable session transition model, lifecycle, and conversation card
- preserve startup transition state across new-session navigation and publish ordinary/worktree first-message progress before exposing the target route
- treat queued first messages and transition progress as renderable content, with deterministic success dismissal and persistent error actions
- remove obsolete worktree-only card paths and document the generalized product contract

## Verification

- `bun run quality:quick`
- isolated runtime exercise confirmed loading → success → timed dismissal across the new-session route transition

## Notes

- success remains visible for three seconds before the 180 ms exit animation
- error transitions do not auto-dismiss
- reduced-motion styling disables card and step animation


yzxoi:
Review found three blocking correctness/design issues:

1. **Transition ownership is lost across `:dir` navigation.** `SessionTransitionProvider` is mounted inside the keyed `params.dir` subtree in `packages/app/src/pages/directory-layout.tsx`, while `submit.ts` publishes into the current provider before navigating to `sessionScopeKey`. Creating a session from a sandbox/worktree can change the route directory to the canonical root, which destroys the store, hides the transition, and resets the pending gate while startup is still running.

1. **`status: "started"` is presented as “message sent”.** The frontend marks the transition successful as soon as `client.session.input()` resolves, but `submitInput()` starts `SessionInvoke.invoke()` fire-and-forget and returns before `createUserMessage()` completes. A later invocation failure is only logged, while the UI still says “Your first message was sent.” The UI must either describe acceptance accurately or wait for a durable message-created signal.

1. **The new-session failure flow does not implement the documented persistent error model.** `PRODUCT.md` says errors remain until retry or dismissal, but every post-create failure clears the transition, deletes the session, navigates away, and leaves only a toast. The chosen product behavior is the persistent error model, so startup errors need a stable owner plus real retry/dismiss actions instead of immediate rollback.

The added unit tests pass, but they only cover pure progress/context/lifecycle helpers. Please add integration coverage for submit → route/provider handoff, accepted-vs-created message semantics, and persistent new-session error/retry behavior.

I will follow up on this PR branch with fixes for these issues.

yzxoi (approved):

PR #658, opened 2026-07-16
fix: settle turn diffs before enrichment

EricSanchezok:
## Summary

- settle each user turn diff before title/body enrichment and preserve every queued turn with FIFO processing
- expose an additive `summary.diffState` contract while keeping `summary.diffs` as the single turn-diff source
- transition the existing diff panel through delayed pending, ready, and low-emphasis error states with reduced-motion support
- document the settlement and frontend synchronization contract

## Verification

- `bun test test/session/summary.test.ts`
- `bun run test:changed` from `packages/synergy`
- `bun run quality:quick`
- isolated Web visual checks in light/dark themes, 375 px width, keyboard navigation, and reduced-motion mode

## Compatibility

- `summary.diffState` is optional and additive
- legacy messages without `diffState` continue to render non-empty `summary.diffs` as ready
- no route, event, migration, config, or persistent store was added


yzxoi:
I reviewed the turn-diff settlement flow and found four correctness issues that should be addressed before merging:

1. **Same-root continuations can lose diffs.** The per-session queue deduplicates work by the root user `messageID`. If another terminal assistant is produced for the same root while settlement is running—for example after additional tool steps, steer, or context—the second trigger is dropped. The active worker uses a message snapshot captured at startup, so it cannot include the later assistant's snapshots. `Turn.collectOne()` also stops at the next user message, which excludes assistants produced after a materialized steer/context message.

1. **The timeout does not cancel the settlement worker.** `withTimeout(summarizeNow(...))` only times out the wrapper promise. The old worker continues running while the queue advances to the next item, so settlement is no longer FIFO and a late completion can overwrite newer turn or session diff state.

1. **A timeout or process restart can leave durable `pending` state indefinitely.** The inner diff failure path writes a terminal error, but the outer summary timeout only logs. There is no restart repair/retry path for expired persisted `diffState.pending`; the UI merely projects it as an error locally.

1. **`deadlineAt` compares clocks from different machines.** The server persists an absolute epoch deadline, while the Web client compares it with browser `Date.now()`. With an external server or clock skew, a live calculation can immediately appear failed, or an expired one can remain pending much longer than intended.

I am preparing a focused follow-up that will add regression coverage for same-root continuation, timeout late completion, terminal timeout persistence/recovery, and the remote-clock contract before changing the implementation.

yzxoi:
Implemented the fixes for the four review findings:

- Queue identity now includes the terminal assistant revision, so a later continuation of the same root turn is summarized instead of being deduplicated away.
- Per-run cancellation now propagates through snapshot diff and LLM work. The queue advances only after the timed-out worker settles, preventing late writes from overwriting newer results.
- Timeout settlement now updates only a still-`pending` turn. If the diff already reached `ready` while session aggregation or enrichment was still running, the completed diff is preserved.
- Expired persisted `pending` settlements are projected to `error/timeout` on backend reads, including after restart.
- The frontend now renders the server-owned settlement state and no longer compares the server's `deadlineAt` with the client clock.

Regression coverage was added for same-root continuations, timeout/late-write ordering, preserving an already-ready diff across a later timeout, stale pending recovery, and client clock skew.

Validation completed:

- `bun test test/session/summary.test.ts` (16 passed)
- `bun test src/components/session-turn-timeline.test.ts src/components/turn-change-summary-panel.test.ts`
- `bun run typecheck`
- `bun run format:check`
- `bun run quality:quick`

yzxoi (approved):

PR #663, opened 2026-07-16
fix(file): bound workspace watcher fan-out

### LINKED ISSUE #638: [P1] Synergy-managed worktree checkout triggers unbounded watcher/status fan-out and stalls the backend
## Summary

Creating a Git worktree under a project-local `.synergy/worktrees/<name>` can make the shared Bun backend unresponsive for one to three minutes even though the `git worktree add` command itself finishes in seconds.

The failure is an unbounded amplification chain:

```
worktree checkout
  -> root workspace watcher sees every checked-out file
  -> dedicated .synergy watcher sees the same internal worktree files
  -> one async workspace-file pipeline per event
  -> repeated Git-status builds, VCS branch probes, sequenced Bus events,
     WebSocket fan-out, reload classification, and per-file logging
  -> event-loop/filesystem starvation across all Scopes
```

A concurrent `Snapshot.track()` can then report a very large wall-clock duration even though its tree is unchanged and `.synergy` is excluded from snapshot contents.

No dedicated issue currently tracks this exact `.synergy/worktrees` watcher/status fan-out.

## Impact and redacted incident evidence

A production-like Desktop managed runtime reproduced the following behavior while two worktrees were created concurrently in separate project Scopes:

- the two worktree commands checked out approximately 2,700 and 3,900 files;
- both Git commands completed successfully in roughly 5–8 seconds;
- over the next approximately 3.5 minutes the backend emitted 7,735 application records, including:
  - 2,856 `file.watcher.updated` publishes;
  - 2,054 project `.synergy` watcher events;
  - 2,054 reload-classification skips for files with no reload target;
- a snapshot that normally completed in 58–77 ms took 81,398 ms;
- the slow snapshot produced the same tree hash as the snapshots before it;
- unrelated HTTP requests completed together after 65–165 seconds, all with HTTP 200;
- the backend process stayed alive and recovered without a restart.

This combination demonstrates shared event-loop / I/O starvation rather than an Electron-only failure or one locked HTTP route.

### 1. The root watcher does not exclude Synergy metadata

`packages/synergy/src/file/ignore.ts` excludes `.git`, build outputs, caches, and dependencies, but not `.synergy`.

`packages/synergy/src/file/watcher.ts` subscribes the project root with `FileIgnore.PATTERNS`. As a result, every file checked out below `.synergy/worktrees/**` enters the normal workspace-file path even though the File workbench hides dot paths by default.

### 2. A second watcher subscribes the whole `.synergy` tree

The dedicated project `.synergy` subscription ignores only `node_modules`. It therefore receives every internal worktree checkout event and emits `global.config.file.changed`.

Most internal worktree paths have no runtime-reload target, but each event still crosses GlobalBus, performs classification, and writes an INFO-level “no targets detected” record.

The root and dedicated subscriptions therefore duplicate work for an internal runtime directory that should not be treated as File workbench content or project configuration.

### 3. The 50 ms coalescing is not backpressure

The current parent-based timers delay events for 50 ms, then run:

```ts
for (const item of batch) {
  void publishWorkspaceFileEvent(...)
}
```

They do not deduplicate paths, bound queue depth, bound concurrency, or await one drain before starting another. A checkout burst therefore creates thousands of independent async pipelines.

### 4. Every event can trigger repeated Git-status work

`publishWorkspaceFileEvent()` currently:

1. calls `WorkspaceFileStatus.invalidate()`;
2. calls `WorkspaceFileIndexer.applyChange()`, which resolves a node;
3. resolves the node again for the event payload.

`WorkspaceFileService.node()` resolves Git status by default. `WorkspaceFileStatus.summary()` has a TTL but no single-flight/in-flight promise. When thousands of invalidations and node resolutions race, many callers can simultaneously observe `fetchedAt === 0` and each start a full status build:

- `git diff --numstat HEAD`;
- `git diff --name-status -M HEAD`;
- `git ls-files --others --exclude-standard`;
- optional untracked-file reads.

This is the same missing single-flight protection proposed in #160, but it is still absent from the current implementation.

The frontend does not consume `event.properties.node`. `packages/app/src/context/file/index.tsx` refetches the affected parent directory instead, so computing Git-enriched node metadata per watcher event is unnecessary for the current Web consumer.

### 5. VCS probes the branch for every non-HEAD file event

`packages/synergy/src/project/vcs.ts` subscribes to `file.watcher.updated`, skips paths ending in `HEAD`, and runs `git rev-parse --abbrev-ref HEAD` for every other file event.

A checkout burst can therefore add thousands of redundant Git processes. Branch refresh should be driven by VCS metadata or coalesced once per watcher batch, not once per workspace file.

### 6. Every file event is a sequenced state event

Each `file.watcher.updated` publish receives a Scope sequence number, enters the replay journal, synchronously emits through GlobalBus, and is serialized for connected WebSocket clients.

This can consume replay capacity and add network/serialization pressure even when the File workbench ultimately performs only one in-flight parent-directory refresh.

## Why the 81-second snapshot is a victim, not the primary cause

Current snapshots exclude every path containing a `.synergy` segment. The slow snapshot retained the exact pre-incident tree hash, so it did not spend 81 seconds indexing the newly created internal worktrees.

`Snapshot.track()` does have 10-second abort and 15-second hard deadlines for each Git subprocess, but it has no total wall-clock deadline. `refreshIndex()` also contains multiple sequential phases:

- `diff-files`;
- `ls-files --others`;
- bounded candidate `lstat`;
- optional `git rm`;
- `git add`;
- `write-tree`.

Under event-loop starvation, child-exit callbacks and the timeout timers themselves can be delayed. Current logging records only total track duration, so the exact delayed phase cannot be attributed from existing telemetry.

This is distinct from the permanent untimed Bun Shell deadlock fixed by #177 / #178.

## LSP timing

LSP was a secondary load wave, not the cause of the 81-second snapshot:

- there were no LSP events during the watcher/snapshot stall window;
- after the snapshot completed and agents began reading the new worktrees, approximately 6,100 worktree LSP records appeared over the next 15 minutes, mostly TypeScript/oxlint diagnostics.

Per-worktree LSP roots and INFO-level per-file diagnostic logging can amplify later CPU/log load, but that should not be conflated with the initial watcher root cause. LSP log aggregation can be handled as a follow-up if needed.

## Regression boundary

The vulnerable combination became broadly reachable when #546 enabled the project workspace watcher by default for the File workbench and added incremental node/status handling.

The current remote `dev` commits after the affected branch base do not change:

- `packages/synergy/src/file/watcher.ts`;
- `packages/synergy/src/file/ignore.ts`;
- `packages/synergy/src/workspace-file/**`;
- `packages/synergy/src/project/vcs.ts`;
- `packages/synergy/src/session/snapshot.ts`;
- `packages/synergy/src/lsp/**`.

The snapshot and LSP fixed-cost improvements from #363 / #365 are already present and do not close this watcher boundary.

### P0 — stop internal worktrees at the subscription boundary

1. Add `.synergy` to the **root workspace watcher subscription ignore list**, without necessarily adding it to the shared File workbench ignore catalog.
   - This keeps hidden-file browsing semantics separate from watcher ownership.
   - The dedicated Synergy config watcher remains responsible for supported project runtime/config files.
2. Exclude `worktrees` and other non-config internal/high-churn directories from the native `.synergy` subscription, at minimum:
   - `.synergy/worktrees/**`;
   - dependency/cache/temp/data directories that cannot own runtime configuration.
3. Share an explicit canonical classifier for supported `.synergy` reload inputs (domain config, legacy config, agents, commands, skills, tools, plugins) so watcher filtering and `RuntimeReload.detectTargetsForFile()` cannot drift.
4. Do not emit or INFO-log one reload event per unsupported internal file.

Filtering must happen in the Parcel watcher ignore configuration where possible, not only after callbacks have already been delivered.

### P0 — remove Git status from the per-file watcher hot path

1. Resolve a changed path at most once with `resolveGitStatus: false`.
2. Reuse that cheap node for index mutation and the optional event payload, or omit the optional node when no consumer needs it.
3. Add a bulk `WorkspaceFileIndexer.applyChanges()` path so one batch mutates/sorts index state once rather than copying and sorting arrays once per file.
4. Invalidate `WorkspaceFileStatus` once per normalized batch, not once per raw event.

This preserves the existing public event schema and frontend convergence behavior.

### P0 — make workspace status single-flight

Add per-Scope generation-aware single-flight behavior to `WorkspaceFileStatus`:

- at most one `build()` runs at a time;
- invalidations during a build mark it dirty;
- after the current build finishes, perform at most one follow-up build for the newest generation;
- all concurrent callers await the same promise;
- force refresh retains explicit semantics without creating parallel builders.

This is a general safety boundary even after internal worktrees are filtered.

### P1 — bound and coalesce general watcher bursts

Replace the unbounded fire-and-forget fan-out with one per-Scope drain:

- normalize and deduplicate event transitions by path;
- keep one in-flight drain per Scope;
- process filesystem metadata with a small fixed concurrency;
- bound queue depth;
- on overload, invalidate the workspace index/status once and use an explicit bounded resync/parent-refresh path rather than launching one task per raw event;
- ensure timer cleanup and Scope disposal cancel pending drains.

If an explicit batch/resync event is added, update Agenda, File workbench, VCS, OpenAPI event schemas, and generated SDK types together.

### P1 — remove per-file VCS process fan-out

Refresh branch state only for relevant VCS metadata changes, or debounce it once per normalized watcher batch. A burst of ordinary file additions must not run `git rev-parse` once per file.

### P2 — add aggregate diagnostics

Add redacted aggregate measurements rather than per-file noise:

- raw, ignored, normalized, and published event counts per watcher batch;
- maximum queue depth and active worker count;
- status-build count, coalesced callers, and duration;
- VCS refresh count;
- snapshot phase durations (`diff-files`, untracked listing, candidate stat, add/remove, write-tree);
- event-loop lag during an overload regression test.

Do not record paths, session IDs, or private configuration.

## Acceptance criteria

- Creating a 5,000-file Git worktree below `.synergy/worktrees` emits:
  - zero `file.watcher.updated` events for its contents;
  - zero `global.config.file.changed` events for its contents;
  - no runtime reload attempts for its contents.
- Edits to every supported project `.synergy` config/runtime root still trigger the correct reload target.
- A synthetic 5,000-file ordinary workspace burst:
  - never exceeds the configured watcher worker limit;
  - runs at most one status build concurrently;
  - causes at most one follow-up status build for invalidations arriving during that build;
  - performs a bounded number of VCS branch refreshes;
  - does not create one unresolved promise pipeline per raw event.
- File workbench directory children and open documents converge after add/change/delete/rename and after focus/reconnect recovery.
- Agenda file triggers retain add/change/delete/rename semantics.
- A concurrent snapshot either completes near its normal bounded cost or reports a bounded failure; unrelated health and lightweight HTTP requests remain responsive.
- Existing snapshot rollback/diff behavior and hidden-file browsing semantics remain unchanged.

## Regression tests

1. Pure path-classification tests for root watcher versus dedicated `.synergy` watcher ownership.
2. A fake Parcel event batch containing thousands of `.synergy/worktrees` paths and supported config paths.
3. A status single-flight test with a controlled build barrier and invocation counter.
4. A watcher queue test proving maximum concurrency, deduplication, overload fallback, and disposal.
5. A VCS test proving thousands of file events cause one bounded branch refresh.
6. File workbench reducer tests for coalesced add/change/delete/rename behavior.
7. A temporary real Git repository regression test where practical; keep native watcher timing out of deterministic unit tests.

## Related

- #160 — workspace status fan-out and proposed in-flight deduplication
- #177 / #178 — permanent snapshot freeze from an untimed Bun Shell call
- #350 — broader runtime/LSP resource lifecycle work
- #363 / #365 — snapshot and concurrent-session fixed-cost reductions
- #515 / #546 — File workbench and default project watcher rollout
- #606 / #609 — renderer lifecycle delivery; related symptom surface, different root cause

yzxoi:
## Summary

- separate workspace-file watching from supported project runtime inputs so internal worktrees and other high-churn `.synergy` data never enter the workspace event pipeline
- coalesce normalized file changes through a bounded single-concurrency drain with explicit resync recovery, batch index updates, single-flight status builds, and HEAD-only VCS refreshes
- update the frontend recovery path, generated watcher event contract, regression coverage, and architecture documentation

## Testing

- `bun run quality:quick`
- `cd packages/synergy && bun test` — clean full-suite pass before review fixes; after the two completion-boundary regression tests were added, 4,511 passed and two unrelated config-import tests flaked (one 5-second timeout and one redirect test)
- `cd packages/synergy && bun test test/config/import.test.ts` — 19 pass, 0 fail on immediate isolated rerun
- `cd packages/synergy && bun test test/file/watcher-events.test.ts test/project/vcs-branch-watcher.test.ts test/workspace-file/status-cache.test.ts test/workspace-file/service.test.ts` — 29 pass, 0 fail
- `bun run --cwd packages/app test` — 638 pass, 5 pre-existing skips, 0 fail

Closes #638


yzxoi:
## Real-environment verification

I verified this PR in an isolated source runtime using `deepseek/deepseek-v4-flash`, with the repository root opened as an active Scope before generating the watcher load.

### Internal worktree burst

To reproduce the checkout write pattern deterministically, I extracted two copies of the current commit under `.synergy/worktrees/**` after the project watcher was active:

- 3,905 tracked files per copy
- **7,810 files total**
- extraction completed in **1.39 s**

During the following 60-second observation window, the health endpoint remained responsive:

| Metric          | Result    |
| --------------- | --------- |
| ---             | ---:      |
| Requests        | 236       |
| Successful      | **236**   |
| Failed/timeouts | **0**     |
| p50             | 1.17 ms   |
| p95             | 36.15 ms  |
| p99             | 115.33 ms |
| Maximum         | 270.45 ms |

The valid load-window logs contained:

- **0** `file.watcher.updated` publishes from the internal worktree payload
- **0** project `.synergy` runtime events from that payload
- **0** reload-classification `no targets detected` records
- **0** references to either generated internal worktree payload

This is the desired subscription-boundary behavior: the files are excluded before entering either the workspace event pipeline or project runtime reload pipeline.

### Concurrent model responsiveness

The runtime log confirmed `providerID=deepseek` and `modelID=deepseek-v4-flash` for all requests.

| Stage                  | Response               | End-to-end time |
| ---------------------- | ---------------------- | --------------- |
| ---                    | ---                    | ---:            |
| Before load            | `PR663_ROOT_SCOPE_OK`  | 13.10 s         |
| During 7,810-file load | `PR663_DURING_LOAD_OK` | 12.91 s         |
| After load             | `PR663_POST_OK`        | 11.67 s         |

The concurrent model request completed normally during the burst, with no backend stall or material latency regression.

### Compatibility checks

I also verified that the filtering does not disable valid watcher behavior:

- touching an ordinary workspace document produced one `file.watcher.updated` event;
- touching an existing `.synergy/command/*.md` file produced a project runtime event;
- runtime reload selected `config` + `command` and completed successfully;
- the source worktree remained clean after verification.

### Result

**PASS.** Under a 7,810-file `.synergy/worktrees` write burst, the backend remained responsive and emitted no internal-worktree watcher/reload fan-out, while ordinary workspace watching, supported `.synergy` reloads, and DeepSeek V4 Flash requests continued to work.

PR #669, opened 2026-07-17
feat: establish frontend localization

EricSanchezok:
## What does this PR do?

- adds the App-owned Lingui runtime with `system`, `en`, and `zh-CN` preferences, global config synchronization, bootstrap persistence, locale-aware formatting, and production-safe catalog loading
- migrates Synergy-owned App and shared UI chrome to semantic runtime descriptors while preserving user, plugin-author, path, identifier, and raw diagnostic pass-through boundaries
- adds deterministic App/UI catalog extraction, strict Simplified Chinese completeness and source-contract checks, CI wiring, tests, and localization architecture/development documentation
- resolves dynamic prompt, session, status-bar, Marketplace, safety, navigation, and tool-card labels reactively so language changes update without reload

## Why?

Synergy previously had no single frontend localization contract, no persisted language preference, and no enforceable boundary between translatable product chrome and verbatim external content. This establishes that foundation and completes the product-wide English/Simplified Chinese migration without introducing parallel locale state or runtime language branches.

## How was it tested?

- `bun run quality:quick`
- `bun run deadcode`
- `bun run workflow:check`
- `bun run secrets:check`
- `bun test` from `packages/synergy` — 4,537 passing; one parallel-only `ExperienceReencode` failure passed 8/8 in isolation
- isolated Web runtime — verified `system` / `en` / `zh-CN` switching, persisted cold refresh, 375 px layout, light/dark system themes, Escape close, accessible language control, zero console errors, and zero 5xx responses

## Checklist

- [x] `bun run workflow:check` passes
- [x] SDK regenerated for the additive locale config schema
- [x] Documentation, AGENTS, and `.synergy` help are updated
- [x] No secrets, local auth files, unrelated cleanup, placeholder scripts, or redundant compatibility wrappers are included

Production builds keep English as the initial/fallback catalog, lazy-load Simplified Chinese, and exclude the development pseudo-locale. The new locale config field is additive and optional; existing configurations continue to resolve through `system`.


yzxoi:
Thanks for the substantial localization integration. I reviewed the PR from the runtime, catalog, Settings synchronization, shared UI package, and production-build perspectives. The overall direction is sound—especially the two-phase catalog activation and ticket-based stale-load protection—but there are several correctness and compatibility issues that should be addressed before merge.

## Confirmed issues

1. **Reactive translations are bypassed across much of the App.** Many components call `useLocale().i18n._(...)` on the stable raw `I18n` object. Lingui Solid only creates a reactive dependency through `useLingui()._()`, so those components can remain in the previous language after an in-place locale switch and produce a mixed-language UI.
2. **Dynamic default messages are silently omitted from extraction.** Confirmation and attention descriptors interpolate runtime values directly into `message` template literals. The extractor skips static IDs whose messages cannot be resolved, so dozens of IDs never enter the PO catalogs and always fall back to English.
3. **A failed background config save can leave locale authority permanently pending.** The controller only clears `pendingPreference` when a matching config value arrives. The Settings save failure path does not reject or roll back the pending attempt, after which different authoritative server values are ignored until refresh or another manual switch.
4. **System-locale semantics are incomplete.** The resolver chooses Chinese if any preferred language is Chinese instead of honoring `navigator.languages` order, and `system` mode does not listen for `languagechange`.
5. **The localStorage mirror is treated as a hard dependency.** A storage read can prevent App startup; a storage write can fail after catalog activation but before controller-state commit, leaving i18n, formatter, and controller state inconsistent.
6. **Superseded switches are reported as failures.** A rapid zh-CN → en switch makes the stale request return `false`, which the Settings UI presents as a catalog-load failure even though the newer switch succeeded.
7. **The shared UI package has source-compatibility regressions.** Public `session-status` exports were removed, `ToolInfo.title` widened from `string`, and Lingui Solid became a new provider/peer contract for shared UI consumers.
8. **PO files are being scanned by Tailwind.** Production CSS contains selectors derived from catalog metadata and the build warns that `file` is not a known CSS property.

The existing locale unit tests, localization check, CI, and production build pass, but they do not exercise real component reactivity, failed-save settlement, catalog completeness for dynamic descriptors, or generated-CSS pollution.

## Planned fix series

I am going to follow up on this PR branch with focused commits:

1. **Unify translation ownership:** use a single App i18n runtime, route reactive rendering through `useLingui()`, reserve a `translateNow()` helper for imperative callbacks, and stop exposing raw i18n through the locale context.
2. **Make descriptors statically extractable:** convert runtime-interpolated defaults to static ICU messages plus `values`, make extraction fail closed for static IDs with dynamic messages, regenerate catalogs, and add catalog-completeness tests.
3. **Complete the preference transaction:** distinguish `applied`, `superseded`, and `failed`; identify pending attempts; explicitly confirm or reject them from the save result; roll back on failure; and synchronize external config without leaving stale Settings state.
4. **Harden runtime semantics:** honor ordered browser language preferences, react to `languagechange`, and make localStorage a best-effort bootstrap mirror.
5. **Preserve shared-package compatibility and build cleanliness:** add a UI translation adapter/fallback boundary, retain deprecated public exports, avoid widening the existing string title field, and exclude PO catalogs from Tailwind scanning.

I will add failing behavioral tests before each fix and report the exact verification commands with the pushed commits.

PR #685, opened 2026-07-17
fix(provider): avoid incompatible coding plan reasoning variants

EricSanchezok:
## Summary

- keep catalog reasoning effort metadata as model capabilities while suppressing automatic Anthropic reasoning variants for Kimi For Coding and MiniMax Anthropic-wire providers
- preserve OpenAI-compatible GLM variants, official Anthropic behavior, and explicit user-configured model variants
- document the distinction between wire-protocol compatibility and provider-option semantic compatibility

## Verification

- `bun test test/provider/transform.test.ts test/provider/provider.test.ts`
- `bun test test/provider`
- `bun run test:changed`
- `bun run format:check`
- `bun run lint`
- `bun run deadcode`
- `bun run typecheck`
- `bun run monorepo:check`
- `bun run package:check`
- `bun run quality:quick`
- `bun turbo test --filter='!synergy'`

## Local full-suite note

The complete Synergy test suite was executed three times locally. Each run completed 4,604 tests with one skipped test and one unrelated test timing out under full-suite load; each timed-out test passed immediately in isolation. The pull request CI uses the repository's Linux test matrix as the final full-suite signal.


yzxoi:
I reviewed the provider-variant changes and found three issues that we will fix in a follow-up commit on this PR:

1. The provider-wide MiniMax Anthropic suppression also removes the supported `thinking: { type: "adaptive" }` control for MiniMax-M3. On MiniMax's Anthropic endpoint, omitting `thinking` disables M3 reasoning, while M2.x reasoning remains always on, so these model generations cannot share one empty-variant policy.
2. The fixed provider-ID allowlist can be bypassed by custom or newly registered MiniMax provider IDs. Those aliases fall through to the generic Anthropic fallback and receive unsupported `thinking.type = "enabled"` / `budgetTokens` options.
3. Removing the old MiniMax guard exposes generic `reasoningEffort` variants on the OpenAI-compatible Chat Completions path. The SDK serializes these as `reasoning_effort`, but MiniMax Chat Completions uses `thinking` / `reasoning_split`; its Responses API reasoning contract is a different endpoint and should not be applied here.

The follow-up will replace the provider-wide behavior with model- and transport-specific reasoning controls, preserve explicit user-configured variants, add regression coverage for MiniMax M2.x, M3, custom aliases, and OpenAI Chat, and run the focused provider suites plus the repository quality gates before pushing.

Issue #582, opened 2026-07-13
feat: Project-Session mapping with stable Project Sessions and routing


yzxoi:
## User report

> [需求 4] Project 与 Session 映射。定义 Clarus Project、Runtime Task、Synergy Session 和本地目录之间的关系。每个 Project 应有稳定的 Project Session 和 workspace；Runtime Task 可创建独立执行 Session。还需要明确普通消息、mention 和 task 的路由及唤醒策略，以及 Project 是否可以指定不同的 Primary Agent。

Reporter: Feishu:ou_25b44caac15ab235465df821bd7c8f2a
Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6a7ac5d70c7cc42f17e2290d1c1

## Triage

- Type: feature_request
- Target repo: SII-Holos/synergy
- Related area: area/session

## Deduplication

No duplicate found in recent open/closed issues. Issue #512 (Super Loop orchestration) and Issue #523 (Agent Organization Runtime) touch on multi-session orchestration but do not define the fundamental Project→Session mapping relationship. No existing issue defines the routing strategy for messages/mentions/tasks in the Clarus Project context.

## Static inspection

- `packages/synergy/src/session/` — the session runtime system. Currently sessions are created per-conversation (via frontend, channel, or `session_control` tool) without a stable "Project Session" concept.
- `packages/synergy/src/session/inbox.ts` — message inbox handling with agent name resolution via `Agent.defaultAgent()`.
- `packages/synergy/src/session/input.ts` — input routing based on agent name and channel type.
- `packages/synergy/src/task/` — task execution, which creates sub-sessions.
- `packages/synergy/src/tool/session-control.ts` — session_control tool, which creates sessions with optional directory binding.
- `packages/synergy/src/runtime/reload.ts` — runtime reload, includes `default_agent` watch.

## Expected behavior

1. Each Clarus Project should have a stable "Project Session" — a persistent session tied to the project lifecycle, not to individual user conversations.
2. The Project Session has a dedicated workspace directory (see Issue 3).
3. Runtime Tasks can create independent execution sessions (sub-sessions) distinct from the stable Project Session.
4. Clear routing and wake-up strategy:
   - Normal messages in a project context → routed to Project Session
   - @mention of a specific agent → routed to that agent's session or sub-session
   - Task execution → creates independent sub-session in project workspace
5. Projects should be able to specify a different Primary Agent (via `default_agent` override at project level).
[...]

PR #686, opened 2026-07-17
feat: add native Clarus integration

### LINKED ISSUE #585: feat: Holos workspace clarus field adaptation for Synergy-side project identification
## User report

> @chaos 关于holos ws中clarus字段的synergy端适配好像忘记开issue了，没有的话请你开一个

Source: feishu:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6a7b46cd34b0dfa9c87e63f5445

## Triage

- Type: feature_request
- Related area: area/session, area/auth-model

## Deduplication

No duplicate found in recent open/closed issues. Issue #582 (Project-Session mapping) covers the high-level Project→Session mapping and routing, but does not specifically address the Holos workspace `clarus` field adaptation on the Synergy side — i.e., how Synergy reads and interprets a `clarus` field from the Holos workspace definition to identify which Clarus Project a workspace belongs to.

## Static inspection

- `packages/synergy/src/config/schema.ts` — Holos config schema (lines 193-206): currently defines `enabled`, `apiUrl`, `wsUrl`, `portalUrl`. No `clarus` field or workspace-level project identification.
- `packages/synergy/src/holos/` — Holos runtime integration, including `runtime.ts`, `auth.ts`, `envelope.ts`, `accounts.ts`. The runtime connects to Holos via WebSocket but doesn't parse a `clarus` workspace field.
- `packages/synergy/src/config/config.ts` — exposes the Holos config type.
- `packages/synergy/src/config/migration.ts` — legacy Holos config migration, no `clarus` handling.
- `packages/synergy-link/src/holos/` — synergy-link Holos client, likely receives workspace information from Holos but doesn't forward a `clarus` field to Synergy config/session routing.
- `packages/synergy/src/session/` — session runtime, would need to use the parsed `clarus` field for project routing.

## Expected behavior

Synergy should be able to read and interpret a `clarus` field from the Holos workspace (WS) definition. When a Holos workspace carries a `clarus` field (e.g., indicating which Clarus Project the workspace belongs to), Synergy should:
1. Parse the field from the Holos workspace data received via WebSocket or API.
2. Map it to the corresponding Clarus Project for session/agent routing purposes.
3. Expose it in the Synergy config or runtime state so that downstream systems (session routing, agent assignment) can use it.

## Actual behavior

Currently the Holos config schema only defines basic connectivity fields (`enabled`, `apiUrl`, `wsUrl`, `portalUrl`). There is no handling or parsing of a `clarus` workspace field anywhere in the codebase.

## Proposed fix

1. Add a `clarus` field (or a broader workspace metadata field) to the Holos-related types/schema to capture the workspace's Clarus Project identity.
2. Update the Holos runtime integration (e.g., `packages/synergy/src/holos/runtime.ts` or the WebSocket handler) to parse and surface this field when received from the Holos API/WS.
3. Ensure the parsed value is accessible to session routing so it can be used for Project→Session mapping (complementing #582's planned routing logic).

Suggested labels: enhancement, area/session, area/auth-model, P2

marzich:
# feat: add native Clarus integration

## What does this PR do?

- Adds native `clarus.*` operations and events to Synergy's existing authenticated Holos Agent Tunnel instead of introducing a second transport.
- Persists Clarus projects, activity, task bindings, outbox state, reconciliation cursors, and migration state with bounded inputs and crash-safe recovery.
- Creates dedicated Home Scope task sessions and project workspaces, bundles the Clarus participation Skill, and exposes explicit result submission through `clarus_submit_task_result`.
- Adds global Clarus routes, generated SDK contracts, project/task navigation, composer flows, status handling, and product/architecture documentation.
- Recovers offline assignments through REST backfill and authenticated payload snapshots while preserving failed cursors and surfacing `sync_failed` reconciliation state.

Closes #585.
Related to #582; this PR implements native Clarus task-session routing and ownership, but does not introduce the stable Project Session model described by that issue.

## Why?

Synergy previously treated Holos as a generic connection and had no native owner for Clarus projects, assignments, task sessions, result delivery, or offline recovery. Agents could not reliably receive production assignments, reconstruct compacted historical payloads, submit results exactly once, or navigate their Clarus work after reconnects and restarts.

This change makes Clarus a first-class Synergy capability while preserving one Holos identity, one authenticated WebSocket, and explicit persistence and trust boundaries.

## How was it tested?

- `./script/generate.ts` — regenerated OpenAPI, SDK, and config contracts after rebasing onto the latest `origin/dev`; zero generated diff remained.
- `bun test test/clarus` — 748 passing tests.
- `bun test test/holos` — 143 passing tests.
- `bun run typecheck` — all workspace typecheck tasks passed.
- `bun run quality:quick` — passed.
- `env -u SHELL bun run quality` — passed: 14/14 Turbo tasks; backend suite 5580 passed, 1 skipped, 0 failed across 430 files.
- Isolated production runtime acceptance against the real Holos service: connected, reconciliation completed without error, 4 projects and 21 historical tasks recovered, and status remained stable across multiple reconciliation intervals.

## Review notes

- Security review: passed with no blockers. Snapshot hydration preserves authenticated same-origin entry, exact redirect handling, public HTTPS destination checks, byte limits, checksum verification, and bounded JSON parsing.
- API compatibility review: passed with no blockers. Clarus routes and generated contracts are additive; migrations cover persisted state upgrades.
- Maintainability review: passed after removing a dead import, stale commentary, and a duplicate shallow payload hash implementation.
- The complete diff is large because tests are the largest category and generated SDK/OpenAPI artifacts are included; the implementation was reviewed for duplicate transports, dead compatibility paths, and unrelated cleanup.

## Checklist

- [x] bun run workflow:check passes
- [x] Reachable PR history secrets scan passes with no leaks
- [x] SDK regenerated after server route and schema changes
- [x] Documentation and built-in Skill help updated


yzxoi (changes requested):
## Request changes

The high-level Clarus direction is sound: reusing the authenticated Holos tunnel, keeping Clarus projects as navigation containers, binding each task to one standard Home-Scope session, using event invalidation plus snapshots instead of polling, and requiring explicit result submission are all good boundaries. The deferred Composer, Project Activity UI, Continue Locally UI, full Task Detail, and recovery CTAs are also reasonable follow-up scope rather than blockers for this PR.

However, the current minimum navigation/reconciliation/submission loop still has several correctness gaps that can present the wrong project, lose live updates, hide historical tasks, archive valid bindings, or lose the audit state of a result submission. I am requesting changes for the following issues.

### Blocking correctness and state-model issues

1. **Preserve the canonical Clarus identity in navigation.**

   `docs/architecture/clarus.md` defines project identity as `(agentId, projectId)` and task identity as `(agentId, projectId, taskId)`. The navigation endpoint scans persisted bindings from up to 16 agents, but `NavigationProjectDto` and `NavigationTaskDto` omit `agentId`, while the frontend groups and selects solely by `projectId` / `taskId` (`packages/synergy/src/clarus/navigation.ts`, `packages/synergy/src/server/clarus-route.ts`, `packages/app/src/context/clarus/clarus-model.ts`, and `packages/app/src/components/clarus/hierarchy.ts`). Two agents with the same project ID are therefore silently merged, and the UI can select or open the wrong bound task session.

   Please either scope the endpoint strictly to one agent or carry `agentId` through the DTO, generated SDK, store, hierarchy keys, selection state, and route actions. The latter matches the persisted canonical identity and is preferred.

1. **Return retained tasks for inactive History projects.**

   The navigation composer includes inactive projects with `lastProjectActivityAt`, but it only calls `listTaskBindingsBounded` for `activeProjects`. The frontend then includes an inactive project in History only when `tasks.length > 0`. In combination, an inactive project is returned without its tasks and is filtered out of the rendered History hierarchy. This contradicts the current product contract that inactive projects with durable task history remain reviewable.

   Please query tasks for the inactive-history project set as well, and add a route-to-hierarchy test proving that an inactive project with a retained task appears in History and opens its existing session.

1. **Do not run negative reconciliation after an incomplete discovery.**

   In `packages/synergy/src/clarus/runtime.ts`, hitting `MAX_NON_PROGRESSING_PAGES` breaks discovery without marking the sweep incomplete. Because `discoveryTruncated` remains false, `archiveMissing(agentId, knownProjectIds)` still runs against a partial result set. A transient stuck cursor can consequently archive active projects that simply occurred after the last observed page.

   `archiveMissing` must run only after the API has explicitly reached the end of pagination. Page-budget exhaustion, repeated cursors, aborts, and request failures must all preserve existing bindings.

1. **Do not treat an attempted subscription as a confirmed subscription.**

   `reconcileProjectSubscription` persists the epoch/generation subscription index immediately after `markDispatched`, before awaiting the response. If dispatch succeeds but the response rejects or the connection drops before acknowledgement, the outbox becomes failed/ambiguous while the persisted index causes all later reconciliation in the same generation to skip the project. Periodic discovery reuses `connectedGeneration`, and the public runtime status does not verify confirmed subscription state, so the UI can later report `connected` although the project has no confirmed live subscription.

   Subscription is desired-state reconciliation, not the same kind of one-shot side effect as result submission. Please model attempted versus confirmed subscription explicitly (or otherwise provide an idempotent recovery path), persist the successful generation only after confirmation, and ensure `connected` cannot mean “transport connected but required subscriptions unconfirmed.” Add a test for dispatch success followed by response rejection.

1. **Isolate a failed payload snapshot from other backfill progress.**

   `backfillAssignment` logs and rethrows snapshot hydration errors. The error aborts the remainder of the page, all later projects in the reconciliation rotation, and the cursor update. A permanently bad assignment therefore blocks every healthy message behind it forever.

   Please retain a retry/deferred record for the failed assignment while allowing later messages and at least other projects to make progress. Add a test with one poison snapshot followed by a valid assignment and another valid project.

1. **Keep the task binding result state consistent with the durable outbox.**

   Both synchronous dispatch failure and asynchronous response failure settle the outbox as `ambiguous` or `rejected`, then call `revertSubmitting`, which changes the task to `needs_attention + idle` and clears `resultOutboxRequestID`. The binding store already provides `markResultAmbiguous` and `markResultRejected`, but these transitions are not used on the runtime failure path. The result is two contradictory durable truths: a terminal outbox record and a retry-looking task binding with no audit link.

   This also makes the user guidance impossible to follow: refreshing navigation cannot reveal whether the submission was ambiguous or rejected, and the current task rows render only `status`, not `resultState`.

   Please make the outbox outcome authoritative for the binding projection: preserve the request association, transition the task to `needs_attention + ambiguous/rejected` as appropriate, and expose non-idle `resultState` in the minimum task-row UI. Ambiguous outcomes must remain non-retryable without external confirmation.

1. **Distinguish local not-dispatched failures from remote rejection.**

   `HolosRuntime.sendNativeRequest` reports local `NOT_CONNECTED` and other pre-dispatch failures with `disposition: "rejected"`. Clarus persists that as terminal rejected, and `clarus_submit_task_result` tells the user that “The Clarus server definitively rejected this result,” although no request reached the server.

   Please preserve at least these three meanings across the transport, outbox, binding, and tool error layers:

   - known local/not-dispatched failure — safe to retry after the local condition is repaired;
   - dispatched but unknown (`ambiguous`) — do not retry automatically;
   - explicit remote rejection — terminal remote refusal.

   `NOT_CONNECTED` must not be presented or audited as a Clarus server rejection.

### Important implementation issues

1. **Remove the O(n) activity scan and duplicate live-message upsert.**

   `ClarusProjectActivityStore.upsert` scans every timeline key to find the existing sort key for one message, making an n-message backfill O(n²). The live project-message path also upserts the same activity twice: the first write includes `senderId`, then `deliverProjectMessage` overwrites the canonical record without it. This affects current reconciliation and History eligibility even though the full Activity UI is deferred.

   Please keep a direct message-ID-to-sort-key index (or otherwise avoid the full scan), perform one canonical upsert per event, and preserve `senderId`. A large-timeline regression test or focused benchmark would be useful.

1. **Shut Clarus down with `GlobalRuntime.stop()`.**

   `GlobalRuntime.stop()` only clears the configured REST port and never calls `ClarusRuntime.shutdown()`. Periodic reconciliation, Holos status subscriptions, tunnel observers, and initialization state can survive an in-process stop/start. Current tests manually call `ClarusRuntime.shutdown()` before `GlobalRuntime.stop()`, which masks the ownership gap.

   Please make the global runtime own the full Clarus lifecycle and add a stop/start test that does not manually clean Clarus up first.

1. **Complete the minimum connection-state presentation.**

    The backend exposes `connection.error`, but the Clarus panel only renders the five status labels, so `sync_failed` does not show the reason. In addition, the model marks data stale only when the navigation HTTP request fails; a successful persisted snapshot while the tunnel is `reconnecting` is displayed without a stale indication. Please represent transport health and data freshness as separate concepts, show the synchronization error, and ensure reconnecting retained data is not presented as current.

1. **Fix keyboard operation of the sidebar Clarus disclosure.**

    The section header is a focusable `div` with `role="button"` and `onClick`, but no Enter/Space handler. The nested project disclosures correctly use real buttons. Please make the section header a button as well (preferred) or implement the full keyboard behavior required by its role.

1. **Do not silently discard retry lineage.**

    The Clarus wire schema accepts `attempt_mode` and `retry_of_task_id`, but `toSemanticDTO` drops both before the runtime event. Either carry these values into the semantic/task model for diagnostics and task-detail follow-up, or remove them from the claimed supported contract until they are intentionally consumed.

### Required verification

Please add focused regression coverage for at least:

- two agents with the same `projectId`;
- inactive project with retained task history;
- non-progressing discovery cursor without archival;
- subscription dispatch followed by response failure;
- poison snapshot followed by healthy messages/projects;
- rejected and ambiguous result outcomes remaining visible and linked to their outbox record;
- local `NOT_CONNECTED` being classified as not dispatched;
- `GlobalRuntime.stop()` releasing Clarus runtime state;
- keyboard operation and reconnecting/stale presentation.

The existing CI checks pass, but the current tests primarily cover local happy paths, synchronous send failures, and isolated store transitions; they do not exercise these cross-stage failure windows. Once the identity, reconciliation readiness, and result-outcome semantics are made consistent across the backend and frontend, the current phased product scope will be a solid minimum Clarus integration.

yzxoi:
Thanks for the review context. We are working through the requested changes now, including the navigation identity and History projection, reconciliation/subscription correctness, result/outbox state consistency, activity performance, lifecycle cleanup, frontend status/accessibility gaps, regression coverage, and documentation updates. We will also resolve the current conflicts with `dev` and push the fixes to this PR for another review pass.

EricSanchezok:
## Architectural concern: Clarus should extend the existing Channel runtime, and each Clarus Project should be a Synergy Project Scope

I reviewed the current head after the correctness fixes. Reusing the authenticated Holos tunnel, standard Sessions, the persistent Inbox/LLM loop, bounded payload handling, durable reconciliation, and explicit result submission are all strong choices. However, I think the PR still places the main product boundary in the wrong location: a Clarus Project is modeled as a parallel `ClarusProjectBinding` container while every Clarus Task Session is created in Home Scope.

If the product invariant is that a Clarus Project is a long-lived work context with its own managed directory, it should be represented directly by a normal Synergy Project Scope. Clarus should extend the existing Channel system rather than establish a separate project/session/navigation/runtime stack beside it.

The desired shape is:

```text
Channel: clarus
└─ Account: agentId
   ├─ Synergy Project Scope: project A
   │  ├─ stable project conversation Session
   │  ├─ Task Session A
   │  └─ Task Session B
   └─ Synergy Project Scope: project B
      ├─ stable project conversation Session
      └─ Task Session C
```

There is no additional local `Project` object between the account and Scope. The remote Clarus Project is represented by the Scope itself.

### Why Channel is the extensible product boundary

Channel already owns most of the generic runtime needed by an external system that communicates with Synergy:

- provider registration;
- multiple configured accounts per provider;
- account connection state and lifecycle;
- reconnect scheduling and terminal failure state;
- normalized inbound message contexts;
- stable endpoint-based Session reuse;
- unattended/autonomous Session invocation;
- replies, proactive pushes, streaming output, tool progress, and reactions;
- outbound delivery and shutdown through the global Channel runtime.

Feishu is currently the only built-in provider, but that should not make its current Home-Scope behavior the permanent limit of Channel. A provider should be able to say that a conversation belongs to an external project, at which point Channel resolves or creates the corresponding Synergy Project Scope before routing the message.

This is useful beyond Clarus without inventing a new abstraction. A future Feishu provider could optionally map a group, topic, or configured workspace to a Project Scope. A WeChat-like provider or another conversational platform could do the same. Providers that do not declare project context continue to use Home Scope exactly as they do today.

In other words, the scalable contract is:

```text
Channel account
└─ zero or more externally identified Project Scopes
   └─ one or more endpoint Sessions
```

This keeps external communication integrations under one runtime and prevents every provider from adding its own startup wiring, config domain, session router, navigation model, storage paths, reconnect state, and frontend event source across the repository.

I do not think an additional top-level `Integration` abstraction is necessary for this product model. If supported external systems are conversational Channels, Channel can remain the single extension boundary.

This does not require every provider to declare a separate capability table. Project-aware routing can be expressed by optional project context on normalized messages or task events: when project context is present, Channel resolves the corresponding Project Scope; when it is absent, Channel uses Home Scope. Clarus-specific task behavior can likewise be represented by an optional task implementation on the provider rather than a parallel boolean flag.

Feishu therefore continues to emit ordinary message contexts and uses Home Scope by default. Clarus emits project-scoped contexts and implements its task operations. A future conversational provider can use the same optional context and interfaces without changing Scope or creating another global runtime.

### Why the current Home-Scope model is costly

`getOrCreateTaskSession()` currently forces every assignment into `Scope.home()` and gives each task a dedicated Clarus workspace. The Clarus Project itself is therefore only a navigation/persistence container, not the owner of its sessions, files, events, project configuration, file services, LSP/VCS state, or runtime lifecycle.

The absence of a stable Project conversation Session also forces project messages through a custom fanout path: one project message is persisted in the Clarus activity store and copied into every non-terminal Task Session. That requires project-level deduplication, per-target fanout progress, deterministic inbox IDs, crash recovery for partial fanout, and terminal-task filtering. Much of this complexity is compensating for the missing project-scoped conversation owner.

This creates parallel product/runtime surfaces that overlap existing systems:

- custom Clarus project navigation beside standard Project Scope navigation;
- custom project message/activity storage beside Channel-backed Sessions;
- custom composer and outbound message paths beside Channel outbound delivery;
- direct Clarus startup/shutdown wiring in `GlobalRuntime` beside Channel lifecycle;
- a Clarus-specific Session endpoint/category for both messaging and task ownership;
- a separate Clarus configuration domain rather than a Channel provider configuration;
- Clarus-specific frontend invalidation and hierarchy ownership outside the normal Scope/session projection.

The result is not merely more files. It makes each future external communication provider likely to repeat the same cross-cutting changes in `GlobalRuntime`, Session, Nav, config, server routes, generated SDK, App state, and Sidebar.

### Extend Channel scope resolution; do not change Scope

Scope itself does not need a Clarus-specific schema, origin union, or provider metadata. The existing contract is sufficient:

1. Channel receives a normalized message or task event with an optional external project identity.
2. Channel deterministically resolves a managed directory from `(channelType, accountId, projectId)`.
3. Channel creates that directory if necessary.
4. Channel calls the existing `Scope.fromDirectory(directory)`.
5. Channel enters `ScopeContext` with the returned Project Scope before creating or delivering to a Session.

Conceptually:

```ts
type ChannelProjectContext = {
  projectId: string
  projectName?: string
}

type MessageContext = {
  // existing fields
  project?: ChannelProjectContext
}

async function resolveMessageScope(ctx: MessageContext): Promise<Scope> {
  if (!ctx.project) return Scope.home()

  const directory = resolveManagedProjectDirectory({
    channelType: ctx.channelType,
    accountId: ctx.accountId,
    projectId: ctx.project.projectId,
  })

  await ensureManagedDirectory(directory)
  const { scope } = await Scope.fromDirectory(directory)

  if (ctx.project.projectName) {
    await Scope.updatePersisted({ scopeID: scope.id, name: ctx.project.projectName })
  }
  return scope
}
```

The managed directory must be derived from a bounded hash or safe encoding of the full composite identity, never from an untrusted remote display name. `projectId` alone is insufficient because IDs can collide across accounts.

This mapping can remain stateless. Under the invariant that one external project always owns one managed directory, there is no need for an additional persisted `ChannelProjectBinding` mapping:

```text
(channelType, accountId, projectId)
→ deterministic managed directory
→ Scope.fromDirectory()
→ stable Scope
```

A persisted mapping would only be required if the product later supports rebinding to arbitrary existing directories, relocating managed projects, many-to-one mappings, or manual unlink/relink behavior. Those capabilities do not need to be introduced here.

The current Scope ID is derived by `Scope.fromDirectory()` rather than supplied by the caller. That is fine: the deterministic directory is the conversion layer, and Scope remains the authority for its own stable ID. No Scope API or schema change is required.

### Give every Channel provider an isolated durable store

Provider protocol state should not be placed in Scope metadata. Instead, Channel should expose a provider/account-namespaced storage facade:

```text
channel/providers/{channelType}/accounts/{accountId}/...
```

For Clarus, the account store can contain project-scoped state such as:

- remote lifecycle and cached remote display metadata;
- desired and confirmed subscription state;
- reconciliation cursors and errors;
- task assignment state;
- result outbox records;
- protocol-specific deduplication and snapshot recovery state.

This is not a persisted Project-to-Scope mapping. The project state remains keyed by the external project identity, while Scope is resolved deterministically when needed.

The ownership boundaries remain clear:

```text
Scope
  local project boundary, directory, worktree, name, icon, timestamps, sandboxes

Channel configuration
  enabled state, managed project root, endpoint options, transport reference

Channel provider storage
  remote lifecycle, subscription, cursor, reconciliation, task and outbox state

Credential store
  agent secrets and tokens

Session storage
  conversation and task transcripts
```

The Channel storage facade should centralize namespace isolation, path-segment encoding, size/read bounds, atomic writes, and locking. Each provider owns its schemas and versioned migrations, registered through the central migration runner.

Disabling or disconnecting a Channel must preserve its provider storage, Project Scopes, directories, and Sessions. A remote project archive/delete should stop remote activity as appropriate but must not silently delete local project files or automatically collapse the Scope into Home.

### Clarus as a project-aware Channel provider

Clarus should register through the existing Channel provider registry. Its provider may attach to the existing Holos tunnel, but it must not open a second WebSocket, duplicate Holos credentials, or compete with Holos for reconnect ownership.

The Channel provider contract may need a small lifecycle distinction for providers that borrow a shared transport, but this should remain inside Channel rather than requiring separate direct initialization in `GlobalRuntime`. `GlobalRuntime` should continue to start and stop Channel once; Channel starts/stops its registered providers.

Clarus configuration should consequently live under the Channel configuration domain while referencing the active Holos identity/transport. For example, the provider can own settings such as the managed project root and Clarus API origin without copying credentials from Holos auth.

This makes configuration, status, reload, and lifecycle behavior consistent across Feishu, Clarus, and future Channels, rather than requiring every new provider to add another top-level config domain and another block in `GlobalRuntime.start()`/`stop()`.

### Session ownership

For each Clarus Project Scope:

- project messages enter one stable Channel conversation Session in that Scope;
- each Clarus Task maps idempotently to one standard Task Session in the same Scope;
- all sessions use the Scope directory by default;
- a Task Session may explicitly use another workspace or managed worktree without changing Scope ownership;
- assignment, deadline extension, result submission, and ambiguous-result handling remain Clarus-specific provider behavior;
- project messages are not copied into every active Task Session by default. Task Sessions receive explicit task context or refer to the project conversation/history through the normal scoped model.

This preserves the good parts of the PR: standard Sessions, explicit result submission, the three outcome classes (`not_dispatched`, `rejected`, `ambiguous`), bounded REST snapshot hydration, reconciliation safety, and durable provider state. It removes the need for Clarus to recreate project ownership, messaging navigation, runtime lifecycle, and configuration beside Channel and Scope.

### Suggested implementation sequence

1. Extend Channel routing so each inbound context resolves either Home or a deterministic managed Project Scope.
2. Add provider/account-namespaced Channel storage with bounded, schema-validated access.
3. Move Clarus configuration into the Channel configuration domain while keeping Holos credentials and transport ownership separate.
4. Register Clarus as a Channel provider that attaches to the existing Holos transport.
5. Resolve every Clarus Project through its deterministic managed directory and `Scope.fromDirectory()`.
6. Create one stable project conversation Session and create Task Sessions in that same Scope.
7. Remove the Home-Scope Clarus task default and project-message fanout path.
8. Remove or consolidate redundant Clarus navigation, composer, event-source, and lifecycle surfaces where Channel/Scope already owns the canonical behavior.
9. Keep reconciliation, subscription, assignment ingestion, and result delivery as focused Clarus provider services, with the result outbox authoritative for delivery state.
10. Reorganize the `phaseN` tests around public invariants: project scope resolution, Channel Session reuse, assignment ingestion, reconciliation, subscription confirmation, result delivery, provider storage isolation, and migration/recovery.

### Required architectural tests

At minimum, I would expect coverage proving:

- the same `(channelType, accountId, projectId)` always resolves to the same Project Scope;
- different accounts with the same `projectId` resolve to different Scopes;
- Clarus project and Task Sessions are owned by that Project Scope, never Home;
- all sessions in the project use the managed project directory unless explicitly bound to another workspace;
- duplicate project events reuse the existing Scope and conversation Session;
- disabling or disconnecting Clarus preserves Scope, directory, Sessions, and provider state;
- remote archive/delete does not silently delete local project files;
- Feishu and other project-unaware Channels continue using Home Scope unchanged;
- a future project-aware Channel can provision a Scope through the same Channel API without modifying Scope, Session, Nav, or `GlobalRuntime`;
- Channel provider state is isolated by provider and account;
- Holos remains the single owner of its authenticated tunnel and credentials.

If the intended product contract is instead that a Clarus Project has no stable local directory and is only a remote task grouping, the current Home-Scope model should be stated explicitly and the absence of a Project Scope defended as an invariant. But if a Clarus Project is the durable working context described above, I think this boundary should be corrected before merging because the current persistence, API, navigation, config, and migration surfaces will make a later move from Home Scope to a project-aware Channel substantially more expensive.

marzich:
Thanks — this matches the intended product model, and we agree that the boundary should be corrected before merging rather than preserving the current Home-Scope Clarus container.

We plan to revise the PR around the following model:

```text
Channel account
├─ managed Home Project Scope
└─ zero or more externally identified Project Scopes
   └─ standard Synergy Sessions
```

For Clarus specifically:

```text
Clarus account: active Holos agentId
├─ Home Project Scope
├─ Clarus Project A Scope
│  ├─ ordinary user-created Sessions
│  ├─ Clarus Task Session A
│  └─ Clarus Task Session B
└─ Clarus Project B Scope
   └─ Clarus Task Session C
```

The concrete ownership decisions are:

- Clarus registers as a real project-aware Channel provider.
- Holos remains the sole owner of the authenticated Agent Tunnel, credentials, heartbeat, request correlation, and reconnect. The Clarus provider borrows that transport; it never opens or reconnects another socket.
- Every configured Channel account owns a managed Home Project Scope. Account enablement creates it immediately, so users can start ordinary Sessions before any inbound message or task.
- Project-unaware Channel traffic, including Feishu traffic, routes to the owning account’s Home Project Scope rather than global Synergy Home.
- A remote Clarus Project is represented directly by a normal Synergy Project Scope.
- `(channelType, accountId, projectId)` deterministically resolves a managed directory, and `Scope.fromDirectory()` remains the sole authority for Scope identity. We will not add a persisted project-to-Scope mapping.
- Complete discovery provisions or reuses active Clarus Project Scopes. Incomplete discovery cannot create negative state or archive projects absent from an incomplete result.
- Each Clarus task ID maps idempotently to one dedicated Task Session in that Project Scope. Users may also create ordinary Sessions in the same Scope through the standard Project UI.
- There is no additional canonical project-conversation Session. `clarus.runtime.task.assigned` creates and wakes its dedicated Task Session. Other project-level events update bounded provider activity/protocol state only; they are not copied into arbitrary Sessions and do not wake the model.
- Assignment/runtime payload is delivered without LLM rewriting as the task prompt. Separate agent-facing instructions explain the Clarus participation skill/tools and result/deadline interaction.
- Managed Channel Projects appear only under the owning account section in the sidebar; the generic Projects section remains for ordinary local Projects. This avoids duplicate navigation while preserving normal Scope and Session behavior.
- Clarus Projects and Task Sessions use Clarus-specific semantic icons/status presentation while remaining standard Project Scopes and project-owned Sessions.
- Active externally managed Projects cannot be locally archived through the standard Scope action; this should be enforced through a generic Channel project-removal guard rather than Clarus metadata in Scope.
- Managed directories use the existing Git initialization behavior. Clarus exposes an account-level `initializeGit` setting, defaulting to true and affecting only newly provisioned Projects.
- Clarus is enabled independently per Holos account. Disabled, Holos-disconnected, and connected-with-zero-active-project states have no ongoing timer, periodic discovery, provider-store scan, or competing reconnect loop. A zero-project account is refreshed only by Holos events, Holos reconnect, or an explicit user-triggered one-shot Refresh Projects action.
- Refresh Projects is an account-level Channel operation. It starts one bounded background reconciliation, reports progress through Channel state/events, and does not reconnect the Holos transport.
- The existing independent Clarus hierarchy, Home-Scope task routing, project-message fanout, direct `GlobalRuntime` lifecycle wiring, and parallel navigation ownership will be removed or folded into Channel/Scope/Session ownership.
- PR #686 is still pre-release. We plan a clean replacement of the old Clarus persistence model rather than a generic cross-Scope Session migration. Old Clarus development state can remain retained but unread; ordinary user Sessions, Projects, files, and configuration are not touched.

We also plan a generic Channel account Activity surface for diagnosis, replacing Clarus-specific project activity/navigation as the inspection boundary. It will show bounded received, normalized, and processing records so users can distinguish transport, schema/routing, agent-execution, and remote-platform failures. Activity is not a Session transcript and is never implicit LLM context.

Initial bounds:

- 100 records per UI page;
- retention capped by both 7 days and 10,000 records per account;
- 256 KiB maximum retained payload per record, with explicit truncation/redaction metadata;
- NDJSON export of the retained filtered window.

The Activity action will live on the Channel account row and open a dedicated inspection surface rather than embedding logs in the sidebar or writing them into a Project repository.

This generalizes the proposal beyond the minimum Clarus correction in one important way: existing Feishu Sessions would move from global Home Scope to the Feishu account’s managed Home Project Scope. We think this yields the coherent long-term product hierarchy—the Channel account is the sidebar work domain, its Home Project owns unscoped conversations, and externally identified Projects sit beside it—but we want explicit maintainer alignment before implementing that cross-provider change.

Could you confirm the following?

1. Do you agree that every Channel account should own a managed Home Project Scope, including existing project-unaware providers such as Feishu, rather than continuing to route those Sessions into global Synergy Home?
2. Do you agree that Channel-owned Home and external Project Scopes should appear only under the owning account’s first-class sidebar section, while the generic Projects section remains for ordinary local Projects?
3. Is a generic account-level Channel Activity inspection/export surface the appropriate replacement for the current Clarus-specific project activity and diagnostic UI?
4. For active externally managed Project Scopes, should this PR include a local per-project disconnect/opt-out action, or should remote lifecycle and account disable remain the only ways to stop participation for now?

Once these ownership questions are aligned, we will reorganize the existing correctness work around Channel account → Project Scope → Session, keep the Holos server/protocol unchanged, and preserve the established result-delivery semantics (`not_dispatched`, `rejected`, and `ambiguous`).

yzxoi:
Thanks for the discussion. After going through the current implementation, the two review rounds, and the intended Project/session behavior in #582/#585, I agree with Eric's core direction.

I think we should converge on four ownership layers and avoid introducing either a parallel Clarus runtime or a larger `Integration` abstraction:

```text
Holos
└─ exclusively owns authentication, credentials, tunnel connection,
   heartbeat, request correlation, and reconnect

Channel
└─ owns provider/account lifecycle and status, external-project-to-Scope
   resolution, Session routing, and the account/project navigation projection

Clarus Channel provider
├─ borrows the Holos tunnel; it never connects or reconnects it
├─ owns discovery, subscription, cursor, dedup, task binding,
│  reconciliation, and result outbox
└─ declares optional project-conversation and external-task capabilities

Synergy Project Scope
├─ directly represents one Clarus Project
├─ owns its managed directory, files, VCS, LSP, sandbox, and events
├─ may contain one stable Project conversation Session
└─ contains the Clarus Task Sessions for that Project
```

The important principle is: **Clarus tells Channel what happened remotely; Channel decides which canonical Scope and Session it belongs to; Holos only transports it; Scope and Session perform the local work.**

## 1. No additional Integration layer

I do not think we need a top-level `Integration` object, `ClarusProjectRuntime`, or another generic framework. The existing Channel provider boundary is the correct extension point; it only needs a small project/task-aware extension.

Likewise, the Clarus provider should not directly call `Scope.fromDirectory()`, create Sessions, invoke the model, own frontend navigation, or attach itself directly to `GlobalRuntime`. Those operations belong to Channel, Scope, and Session respectively.

Channel core also should not absorb Clarus protocol state. Cursor, subscription ACKs, dedup, task-binding transitions, request lineage, and result outbox remain private to the Clarus provider.

## 2. Borrowed Holos transport

Holos remains the sole owner of the physical connection. The Clarus provider receives a restricted borrowed transport that can observe readiness/events and issue correlated requests, but cannot connect, disconnect, change credentials, maintain heartbeats, or reconnect.

Conceptually:

```ts
interface HolosBorrowedTransport {
  status(): HolosTransportStatus
  onStatus(listener: (status: HolosTransportStatus) => void): Unsubscribe
  observe(listener: (event: NativeEvent) => Promise<boolean>): Unsubscribe
  request(input: NativeRequest): Promise<NativeResponse>
}
```

Startup should therefore be:

```text
GlobalRuntime
├─ initialize Holos
├─ register Channel providers, including Clarus
└─ initialize Channel accounts
```

If Holos is unavailable, the Clarus Channel account becomes `waiting_for_transport` / `reconnecting`; it does not throw into a second Channel reconnect loop. When Holos reconnects, the provider re-runs discovery/subscription reconciliation and flushes its outbox. Already-running Project Sessions continue locally.

Shutdown is the reverse: Channel stops the Clarus provider and unregisters its listeners before Holos closes the tunnel.

## 3. Minimal Channel extension

I would keep the generic contract deliberately small. We do not need a large capability taxonomy or a universal integration event bus.

The provider's `connect()` can receive an internal Channel-owned host with a few narrow operations:

```ts
interface ChannelHost {
  status: {
    update(status: ChannelProviderStatus): void
  }

  projects: {
    reconcile(input: {
      projects: ExternalProjectRef[]
      complete: boolean
    }): Promise<ManagedProjectHandle[]>

    ensure(project: ExternalProjectRef): Promise<ManagedProjectHandle>
  }

  conversations: {
    receive(input: ProjectConversationInput): Promise<DeliveryResult>
  }

  tasks: {
    dispatch(input: ExternalTaskDispatch): Promise<TaskDispatchResult>
    cancel(input: ExternalTaskIdentity): Promise<TaskCancelResult>
  }
}
```

This is not a fifth architectural layer; it is simply Channel's provider-host API, analogous to extending today's `onMessage` callback with project/task-aware routing.

Provider method groups can be optional, and their presence is sufficient to express capability:

- existing messaging methods for Feishu;
- optional managed-project support;
- optional Project conversation outbound;
- optional external-task operations.

Feishu does not implement project/task support and keeps its current Home-Scope and endpoint behavior unchanged. This refactor should not require a universal per-account Home Project or a Feishu session migration.

## 4. Clarus Project becomes the canonical Project Scope

The external identity is:

```text
(channelType, accountId, externalProjectId)
```

Channel derives a deterministic managed directory from that identity, creates the directory, and then calls `Scope.fromDirectory()`.

```text
Clarus discovery
→ ChannelHost.projects.reconcile()
→ derive deterministic managed directory
→ create directory
→ Scope.fromDirectory(directory)
→ persist managed ownership
```

A missing or invalid directory must never silently fall back to Home. Channel should explicitly verify that resolution produced a Project Scope and return a local `not_dispatched` outcome otherwise.

The directory can initially be non-Git. If it later becomes a repository, the existing worktree-based Scope reuse should preserve the original Scope identity.

Although the directory derivation is deterministic, Channel still needs a small durable bidirectional ownership projection:

```ts
interface ManagedProjectOwnership {
  channelType: string
  accountId: string
  externalProjectId: string
  scopeID: string
  directory: string
  remoteState: "active" | "inactive" | "stale"
  participation: "active" | "paused"
  createdAt: number
  lastSeenAt: number
}
```

This is not another Project model. It exists so Channel can resolve both directions, avoid duplicate sidebar entries, protect active managed Projects from local archive, preserve navigation after account disable, and apply remote lifecycle changes without putting Clarus-specific fields in Scope.

## 5. Session model

Clarus does not need a separate `SessionEndpoint` kind. It can remain a normal Channel endpoint with a generic target:

```ts
type ChannelTarget =
  | { kind: "chat"; chatId: string }
  | { kind: "project"; externalProjectId: string }
  | {
      kind: "task"
      externalProjectId: string
      externalTaskId: string
    }
```

Existing Feishu endpoint keys remain unchanged when no external target is present.

For Clarus:

- A Project conversation Session is optional and capability-gated. If supported, there is at most one stable Session per `(channelType, accountId, externalProjectId)`, created lazily in the Project Scope.
- Each external task gets one stable Task Session per `(channelType, accountId, externalProjectId, externalTaskId)`, also in the Project Scope.
- A normal Project message goes only to the stable Project Session. It must not fan out to all active Task Sessions.
- Duplicate task delivery reuses the same Task Session instead of creating another one.
- Endpoint identity must not contain `scopeID`, directory, display name, cursor, or result request ID.

## 6. Routing responsibilities

The expected routing is:

| Remote event               | Local behavior                                   | Wake model         |
| -------------------------- | ------------------------------------------------ | ------------------ |
| ---                        | ---                                              | ---:               |
| Project discovered/updated | Reconcile canonical Project Scope                | No                 |
| Project removed/archived   | Apply remote lifecycle to managed Scope          | No                 |
| Subscription ACK           | Update Clarus provider state/cursor              | No                 |
| Normal Project message     | Stable Project Session, if supported             | Yes                |
| Task assigned              | Project-owned Task Session                       | Yes                |
| Task extension/message     | Corresponding Task Session inbox                 | Protocol-dependent |
| Task cancelled             | Cancel corresponding Task Session                | No                 |
| File/system/notary event   | Managed files/provider state/diagnostics         | No                 |
| Result submission          | Clarus binding and result outbox                 | No                 |
| Holos disconnected         | Mark account/projects stale; preserve local work | No                 |

## 7. Preserve the correctness invariants already fixed in this PR

The architectural rewrite should move, not discard, the important state-machine fixes from the first review:

- use `(account/agent, project, task)` composite identity;
- only perform negative project archival after a complete discovery snapshot;
- mark subscription confirmed only after remote ACK;
- isolate poison messages without blocking later messages or other Projects;
- preserve `not_dispatched`, `rejected`, and `ambiguous` as distinct states;
- retry `not_dispatched` with a new request ID while preserving lineage;
- never automatically retry `ambiguous` delivery;
- preserve cursor, dedup, binding, and outbox state across restart;
- clean up listeners/timers on Channel shutdown;
- represent stale/reconnecting state honestly in navigation.

Inbound cursor progression and result delivery should be separate pipelines. The cursor advances after an event has been durably classified or handed to a Session, not after a potentially long model execution and not after result delivery. Result delivery is handled independently by the durable outbox.

The outbox must persist before send, use stable idempotency/correlation, flush after reconnect, retain request lineage, and never silently drop unresolved results after a retry count.

## 8. Navigation

The frontend should project canonical objects rather than maintain a second Clarus hierarchy:

```text
Channels
└─ Clarus account
   ├─ Project A  → canonical Scope
   │  ├─ Project conversation Session
   │  ├─ Task Session 1
   │  └─ Task Session 2
   └─ Project B  → canonical Scope
```

The account Project nodes reference canonical `scopeID`s, and their Sessions come from the normal Scope/session indexes. Externally managed Scopes may be hidden from the generic Projects section to avoid duplication, but they must remain in the canonical Scope snapshot so global navigation and direct session lookup continue to work.

This should replace the dedicated Clarus navigation route, navigation store/context, and parallel Project/task DTO hierarchy.

## 9. Lifecycle and local pause

A remotely active managed Project should not be locally archived while participation is active. The account UI should instead expose a minimal `Pause participation` action:

- persist `participation = paused`;
- unsubscribe through the Clarus provider;
- stop accepting new tasks;
- keep Scope, files, Sessions, and history;
- do not automatically re-enable on discovery;
- allow an explicit Resume.

Remote archival/removal should only be applied from an authoritative lifecycle event or a complete discovery snapshot. It should preserve local files, Sessions, and outbox history rather than delete them.

## 10. Suggested implementation sequence

I would organize the implementation in two reviewable stages, even if they remain in one PR:

### Stage A: generic Channel foundation

1. Extract the existing Channel session-routing path without changing Feishu behavior.
2. Add generic Project/task Channel endpoint targets with backward-compatible keying.
3. Add managed-project ownership, deterministic directory creation, and Scope resolution.
4. Add Project/Task Session routing and cancellation.
5. Add account → canonical Scope → canonical Session navigation projection.
6. Add active/archive/pause lifecycle rules.

### Stage B: Clarus provider migration

1. Move Clarus protocol code under `channel/provider/clarus/`.
2. Borrow the Holos native transport; remove independent connect/reconnect behavior.
3. Move discovery, subscription, cursor, dedup, binding, reconciliation, and outbox into the provider.
4. Route Project/task events through ChannelHost.
5. Move Task Sessions from Home to the corresponding Project Scope.
6. Retarget result submission to the Clarus provider.
7. Remove direct `ClarusRuntime` wiring, the parallel workspace owner, Project-message fanout, and the dedicated Clarus navigation stack.
8. Preserve and adapt the existing behavioral tests around reconciliation and delivery semantics.

I would keep the following out of this refactor:

- universal per-Channel-account Home Projects;
- migration of existing Feishu Sessions;
- a generic Channel Activity subsystem;
- arbitrary rebinding of a managed Project to a user-selected directory;
- automatic deletion of managed Project files.

This gives us one clear ownership chain without a new integration layer:

```text
Holos transports
→ Clarus provider interprets and reconciles
→ Channel resolves Scope and Session
→ Scope/Session perform the local work
```

marzich:
Thanks everyone for the architecture discussion and for the implementation work in marzich/synergy#1. We have now finished the product and ownership discussion and plan to use that branch as the implementation foundation, but not merge it unchanged. We will harden and correct the contracts on top of it, then update this PR with a clean, reviewable history.

## Final product scope for this PR

Native Clarus is a **task-execution Channel provider** in this iteration:

```text
Holos transport
→ Clarus Channel provider
→ Channel-managed Project Scope
→ one stable Task Session per external task ID
→ durable result delivery
```

- Holos remains the exclusive owner of authentication, credentials, the Agent Tunnel, heartbeat, request correlation, epoch/generation, and reconnect.
- Clarus borrows that transport. It never creates another socket, stores Holos credentials, or runs a competing reconnect loop.
- A remote Clarus Project directly owns one canonical Synergy Project Scope with a deterministic managed directory.
- Channel owns external-project-to-Scope ownership, task Session routing, lifecycle projection, navigation, and local archive protection.
- Clarus provider owns only Clarus protocol state: discovery, subscription ACKs, cursor, dedup, remote assignment state, request lineage, and result outbox.
- Each `(accountId, projectId, taskId)` maps to one stable dedicated Task Session in the Project Scope. A new task ID creates a new Session; a new run/attempt for the same task ID reuses the Session.
- Ordinary user-created Sessions may coexist in the same Project Scope.

This PR will **not enable Project conversation behavior**. Discovery must not create a Project conversation Session, and `project.message.created` must not invoke a model or be delivered to an arbitrary Session. Only `runtime.task.assigned` creates/wakes a Task Session. Other project-level events update bounded provider state/diagnostics without waking the model.

## Channel contracts to complete

We will complete the generic foundation started in marzich/synergy#1:

- a narrow Channel-owned host contract for status, managed Projects, task dispatch/update/cancel, and diagnostics;
- generic `ChannelTarget` identities for `chat`, `project`, and `task`, preserving existing Feishu chat endpoint keys;
- a small durable managed-project ownership projection containing external identity, canonical `scopeID`/directory, remote lifecycle, and timestamps;
- one canonical navigation projection:

```text
Channels
└─ Clarus account
   ├─ Project A → canonical Scope
   │  ├─ ordinary Sessions
   │  └─ Clarus Task Sessions
   └─ Project B → canonical Scope
```

Channel-managed Projects are not duplicated in the generic Projects section. A remote-archived Project remains visible under the Clarus account with an archived status while its local Scope, files, and Sessions remain usable. It disappears only when the user archives the local Project through the normal Project workflow.

We will not add a separate local “Pause participation” feature. Remote pause state does not interrupt already accepted local tasks, deadline reminders, result delivery, or deadline extension. If a user wants to stop a local task, they open its Task Session and use the existing Abort action.

## Lifecycle and correctness hardening

The revised implementation will include:

- Holos login creates a Clarus account entry disabled by default; the user explicitly enables it.
- Disabled Clarus performs no attach, discovery, provider-state scan, timer, or Scope provisioning.
- If Holos is unavailable, Clarus enters `waiting_for_transport` and waits for Holos readiness events; it does not use the ordinary Channel reconnect timer.
- Reconnect performs event-driven reattach, complete discovery/subscription reconciliation, and outbox flush.
- Zero active Projects causes no periodic discovery or storage scan; new Projects are found through Holos events, reconnect, or an explicit account-level **Refresh Projects** action.
- Inbound events are accepted only for the current agent, epoch, and generation.
- Negative Project reconciliation happens only after a complete discovery snapshot.
- Subscription is confirmed only after the correlated remote ACK.
- Poison events are durably classified without blocking later events or other Projects.
- Result delivery remains durable-before-send and preserves `not_dispatched`, `rejected`, `ambiguous`, and `acknowledged`. Only `not_dispatched` is automatically retried with a new request ID and preserved lineage; `ambiguous` is never automatically retried.

## Task delivery

The Task Session prompt will include every task-related field supplied by Clarus—project/run/task/subtask IDs, phase, attempt, attempt mode, retry lineage, deadline, goal, instructions, input, context, and task input—using deterministic bounded serialization without summarization or LLM preprocessing. Separate agent guidance will load the Clarus participation skill and explain result, progress, and deadline-extension tools.

Deadline safety will use the existing Agenda scheduler rather than a second Clarus timer system. A reminder is scheduled dynamically before the deadline and injected as hidden agent guidance into the same Task Session. It is cancelled once the result is durably submitted or the task becomes terminal. Remote Project pause does not suppress local reminders or delivery for already accepted tasks.

## Diagnostics and explicit non-goals

This PR will add bounded structured diagnostics plus an account-level NDJSON log download. The retained records include bounded/redacted task prompts and result bodies, never credentials or auth headers, with explicit truncation/redaction metadata. Initial bounds are 7 days, 10,000 records per account, and 256 KiB per record.

The following are intentionally deferred:

- universal per-Channel-account Home Projects;
- Feishu Session migration;
- Project conversation capability;
- a generic Channel Activity UI/panel;
- arbitrary managed-directory rebinding;
- Clarus Project-management-agent behavior.

Because the previous Clarus persistence model has not shipped, we will use a pre-release clean replacement rather than migrate Home-Scope Clarus Sessions. Old development state remains retained but unread; ordinary user Projects, Sessions, files, and configuration are untouched.

## Integration plan

We will use marzich/synergy#1 as the code foundation, preserve its Channel/Scope refactor, and add the contract and correctness work above. The final #686 history will be reorganized into reviewable stages rather than merging the current stacked branch unchanged:

1. generic Channel project/task foundation;
2. Clarus provider migration and lifecycle hardening;
3. restored behavior-first tests for reconnect, stale generation, discovery completeness, subscription ACKs, poison-event isolation, assignment identity/retry lineage, deadline reminders, outbox failure windows, shutdown cleanup, and zero-pressure states.

We are preparing the implementation Blueprint from this agreed scope and will update #686 once the revised branch passes the focused and repository quality gates.

yzxoi:
#852 superseded

PR #695, opened 2026-07-18
fix(agent): expose deferred tool discovery to subagents

EricSanchezok:
## Summary

- allow native classic and max subagents to search for and expand deferred tools already permitted by their agent profile
- keep denied deferred tools unavailable after expansion and cover the boundary with catalog and resolver-level tests
- keep model-only internal agents isolated from subagent discovery tools
- make the compaction agent's built-in permission layer deny all tools while retaining the invocation's independent empty-tool-set guarantee
- synchronize Cortex, execution-boundary, compaction, and built-in agent development documentation

## Verification

- `bun test test/agent/max-subagents.test.ts test/agent/agent.test.ts test/tool/exposure.test.ts` (75 pass)
- `bun run quality:quick`
- `bun run deadcode`
- `SYNERGY_LINK_HOME=/tmp/synergy-link-ci-test bun test --timeout 30000` from `packages/synergy` (4,641 pass, 1 skip)


yzxoi:
Thanks for the review. We found two regressions in the current implementation and plan to address both in a follow-up commit:

1. **Deferred expansion currently activates permissions that are too broad for some subagent profiles.** `search_tools` / `expand_tools` are added to the common native-subagent ruleset, but that common ruleset also contains latent allows for stateful Notes tools and cross-session history tools. Once the groups become expandable, a nominally read-only agent such as `code-cartographer` can expose `note_write` and `note_edit`, and ordinary agents can expose Session-history tools that should be owned by the dedicated `sessionHistory` profile. The same concern applies to deferred MCP groups because the common profile currently allows `mcp__*` wholesale.

1. **Tool discovery does not apply the Cortex/root-message tool allowlist.** `expand_tools` builds its catalog from agent and session permissions only, while the next model turn also applies the invocation's `tools` allowlist. With an allowlist such as `{ "*": false, "expand_tools": true, "note_read": true }`, expansion incorrectly reports `note_write`/`note_edit` as available and persists the group, even though those tools disappear again on the next turn.

We reproduced both cases with focused behavioral tests. The existing CI tests cover a permission-denied Browser tool remaining hidden, but do not cover stateful tools accidentally allowed by the common profile or invocation-level tool allowlists.

Planned fix:

- narrow the common deferred permissions and move Notes, Session, Memory, and MCP grants to the profiles that actually own those capabilities;
- add explicit negative profile tests, including ensuring a read-only subagent can expand Notes for reading without gaining `note_write`/`note_edit`;
- pass the effective invocation tool allowlist into `ToolDiscovery`, so `search_tools`, `expand_tools`, and the next `ToolResolver` turn use the same availability calculation;
- add regression coverage for Cortex tasks with a restrictive `tools` map, then run the focused agent/tool suites and `quality:quick`.

The compaction change itself looks sound: it keeps the independent empty-tool invocation and strengthens the built-in permission layer to deny all tools.

PR #696, opened 2026-07-18
fix: stabilize subsession status bar popover

EricSanchezok:
## Summary

- prevent the StatusBar subsession trigger from opening and closing during the same click
- keep loaded child-session results stable while refreshing the current session
- discard stale child-session responses after navigating to another session
- retry the exact failed search or pagination request instead of reconstructing it from stale displayed state
- keep the trigger width fixed and move the loaded total to the panel

## Verification

- `bun test src/components/status-bar/subsession-controller.test.ts src/components/status-bar/subsession.test.ts src/components/status-bar/runtime.test.ts` (from `packages/app`)
- `bun run quality:quick`
- `bun run deadcode`


yzxoi:
Thanks for the stabilization work here. During review I reproduced two regressions caused by treating a new search query like a same-query refresh:

- when the debounced search changes, the previous query's child-session rows remain visible and clickable while the new request is pending;
- if the new root search fails, the previous query's `pageIndex`/cursor state remains active, so the footer can send the new query with a cursor from the old result set.

Both come from the same state-identity issue: stale-while-revalidate is valid for a refresh of the same `(sessionID, query)`, but a different query is a different result set and must invalidate its rows and pagination cursors.

I'm working on a focused follow-up in this PR now. The fix will make query identity explicit in the controller, preserve rows only for same-query refreshes/retries, invalidate list and pagination state on query changes, and add regression coverage for pending and failed query transitions. I'll also update the frontend product contract so this behavior remains explicit.

yzxoi:
The query-identity repair is now complete.

What changed:

- the subsession controller identifies the displayed result set by `(sessionID, normalized query)` and normalizes the query before both identity comparison and requests;
- same-query refresh/retry keeps the current rows visible;
- changing the owner or query invalidates rows, totals, page index, and pagination cursors before the new request runs;
- regression coverage checks normalized same-query refresh, pending query transitions, failed root searches, exact retry behavior, A→B→A query round trips, concurrent stale query responses, stale session responses, and successful recovery;
- `packages/app/PRODUCT.md` records the paginated result identity rule.

Verification completed:

- focused StatusBar tests: 14 passed;
- app test suite: 897 passed, 5 skipped, 0 failed;
- app typecheck: passed;
- app production build: passed;
- repository `quality:quick`: passed.

yzxoi:
Fixed the child-session Popover width mismatch reported in review.

The Popover surface remains responsively scoped to 24rem, while its direct body content now fills the available padded width instead of adding a nested 18rem (`w-72`) shell. This removes the empty right column without narrowing the intended three-column session rows.

Coverage includes the compiled production CSS contract and actual Chromium geometry checks at both 800px and 375px viewports, using the production `border-box` model. The checks compare the child content width/right edge with the Popover body's available content box.

PR #722, opened 2026-07-20
fix(app): polish raw message inspector

EricSanchezok:
## Summary

- make the raw-message list denser while using available width and preserving each identifier's stable tail
- add an explicit line-wrap toggle and reliable horizontal overflow for long JSON
- simplify the Context entry, improve sensitive-data guidance, and update English, Chinese, and pseudo catalogs

## Verification

- `bun run quality:quick`
- visual QA in light and dark themes at desktop, 760px, and 375px widths


yzxoi:
Thanks for the polish pass. During review we found three regressions that should be addressed before merge:

1. On narrow layouts, the new **Wrap lines** text button remains the shared small-button height (22px), while adjacent preview controls are expanded to the required 44px touch target.
2. The long-line width wrapper and `@pierre/diffs` both own horizontal overflow. For CJK content or lines beyond the 1000ch cap, this can produce nested horizontal scrolling before the true line end is reachable.
3. The new single-row list layout can squeeze the only `Hidden` / `Excluded` status indicators out of view around the 768px breakpoint because role and timestamp columns are non-shrinking and the fixed ID tail consumes most of the remaining metadata width.

We’re working on fixes now and will push a follow-up commit to this PR shortly, including focused coverage for the affected responsive and overflow behavior.

yzxoi (approved):

PR #723, opened 2026-07-20
fix: prevent repeated background task polling

EricSanchezok:
## Summary

- align primary agent, Cortex reminder, and task tool guidance around automatic completion notifications
- remove contradictory progress and tail polling examples from background task results
- add regression coverage for the shared one-shot diagnostic contract

## Testing

- `bun test test/agent/task-polling-prompt.test.ts`
- `bun run format:check`
- `git diff --check`

## Not Run

- `bun run typecheck` could not complete because `tsgo` is unavailable in the current environment


yzxoi:
Thanks for the anti-polling cleanup. I traced the complete Cortex notification/result lifecycle and found a few protocol gaps that are not covered by the current prompt-string test:

1. The automatic completion Inbox message currently reports terminal status, but does not deliver the task result. The updated prompts describe it as result delivery and remove the canonical `task_output(..., mode="full")` consumption step.
2. `TaskOutputTool.afterPersist` currently acknowledges every terminal read, so a diagnostic `progress`/`tail`/`summary` read can remove the completion notification without persisting the final result into the parent conversation.
3. `block: true` combined with a diagnostic mode can suppress the completion notification while still not returning the result.
4. Terminal tasks are evicted from the in-memory Cortex registry after cleanup, while the completion notification is durable, so delayed `mode="full"` retrieval needs a durable child-session fallback.

I’m going to update this PR with a two-phase contract:

- the completion notification stays lightweight and tells the parent to retrieve the final result once with `task_output(task_id="...", mode="full")`;
- only a successfully persisted terminal `full`/default read acknowledges the completion notification;
- diagnostic modes never consume the notification;
- `block: true` is limited to `full`/default mode;
- delayed result reads recover from the durable child session;
- behavior tests cover terminal diagnostics, full-result consumption, notification content, wait races, restart/eviction recovery, and multi-task isolation.

I’ll also align the remaining agent/tool guidance and the Cortex architecture document before pushing the fix to this branch.

yzxoi (approved):

PR #731, opened 2026-07-21
fix: bound search output and capture memory incidents

EricSanchezok:
## Root cause

Bun allocation failures can occur while process memory is distributed across the JavaScript heap, external allocations, and ArrayBuffers even when host memory is available. Concurrent long LLM turns, retained message projections, raw tool input, and fully buffered search subprocess output can amplify transient allocation pressure. Existing telemetry did not preserve enough category or turn-level evidence to attribute an incident.

## Changes

- Stream and bound ripgrep and ast-grep output, stop producers after the requested result window, and preserve pagination and truncation semantics.
- Add MessageCache footprint, entry, hit/miss, eviction, and protected-overbudget metrics.
- Record per-turn projection, request, tool-schema, output, raw-tool, active-stream, and lifecycle memory deltas.
- Add heap-used, external, and ArrayBuffer data to Performance summary, charts, and Analyze input.
- Emit bounded OOM incidents with nearby resources, running spans, cache and turn sizes, and GC decisions.
- Capture retained heap snapshots at soft pressure in development before allocation headroom is exhausted.

## Verification

- bun test with 126 focused search, LLM memory, incident, cache, observability, and Performance tests
- bun run test:changed in packages/synergy
- bun run test:ci in packages/synergy
- bun run quality:quick
- isolated source server health check and broad search replay with early subprocess termination


yzxoi:
## Review findings for PR #731

Thanks for the bounded search output and OOM telemetry work. I reviewed the design and code on this branch and landed follow-up fixes in `38d21466`.

### Design / code issues found

#### P0

1. **Soft-pressure heap snapshot was not necessary and was risky.**  
Auto-capturing a Chrome heap snapshot at soft pressure (local default) allocates on the same path that is already near failure. It is not required for search bounding, MessageCache metrics, turn-size telemetry, or OOM incident attribution.  
**Fix landed:** removed `SessionMemoryProfile` and the soft-path snapshot hook; docs updated accordingly.

1. **OOM detector was too narrow.**  
Matching only `/out of memory/i` misses common Bun/V8 messages such as `Array buffer allocation failed`, heap-limit errors, `Cannot allocate memory`, and nested `cause`/`ENOMEM`.  
**Fix landed:** broader multi-message / cause-aware detector.

1. **OOM capture was too heavy.**  
Capture called `maybeCollect` (GC / previously snapshot), scanned spans since process start, and did extra work after allocation failure.  
**Fix landed:** light capture only — current memory sample, recent resource rows, last-5-minute running spans, cache/turn summaries; no GC/snapshot.

#### P1

1. **`Ripgrep.files` was still unbounded.**  
Content matches used `ProcessOutput.lines`, but file listing still used an unbounded reader.  
**Fix landed:** stream `files()` through the same bounded line reader (20MB default total).

1. **Workspace content search lost pagination after limit/timeout.**  
Truncation could set `truncated=true` with empty `nextCursor` when fewer than a full page was collected.  
**Fix landed:** resume from `offset + page.length` whenever early-stop left items.

1. **ast-grep mapped every `LimitError` to `max_output_bytes`.**  
Oversized single records were mislabeled.  
**Fix landed:** preserve `error.reason` and format `max_record_bytes` distinctly.

1. **MessageCache hit/miss counters were polluted by internal `get()` in upsert paths.**  
**Fix landed:** internal reads skip hit/miss accounting.

1. **`ProcessOutput.terminate` / `drainText` could hang.**  
Terminate returned after timeout even if the process survived; drain had no abort.  
**Fix landed:** terminate returns success boolean; drainText accepts `signal` and cancels.

#### P2

1. **Active LLM stream always rendered as warning in the Performance panel.**  
**Fix landed:** default tone for stream counts.

1. **Telemetry density / process-level memory deltas can still be misread as pure per-turn cost.**  
Not fully redesigned here; left as residual risk. Prefer reading deltas as coincident process samples.

1. **arrayBuffers soft/critical thresholds remain asymmetric (1 GiB / 8 GiB).**  
Residual design note only.

### What this follow-up commit changes

- Remove soft-pressure heap snapshot path entirely
- Harden OOM recognition + lighten incident capture
- Bound `Ripgrep.files`
- Fix content-search pagination after early stop
- Preserve ast-grep limit reasons
- Fix MessageCache stats
- Abortable drain + terminate status
- Soften LLM stream warning tone
- Update `docs/operations/performance-observability.md`

### Verification

Focused tests run in this worktree:

- `packages/synergy/test/process/output.test.ts` pass
- `packages/synergy/test/session/memory-pressure.test.ts` pass
- `packages/synergy/test/session/llm-memory.test.ts` pass

Blocked in this isolated worktree by missing workspace package links for `@ericsanchezok/synergy-util` when importing session message types:

- `packages/synergy/test/session/message-cache.test.ts`
- `packages/synergy/test/session/memory-incident.test.ts`

Please re-run the full package suite / CI on the PR branch.

### Residual risks not fully closed here

- Hot-path `LLMTurnMemory.estimateBytes` still walks large graphs every step
- LLM retention bounds themselves (raw tool input / concurrent streams) remain mostly observational rather than hard caps
- Process-level memory deltas are still not true per-turn attribution

Happy to iterate further if you want hard retention caps next.

yzxoi (dismissed):

yzxoi:
P3 clarification: automatic soft-pressure heap snapshots have been removed from this PR.

The final implementation intentionally keeps memory-pressure and OOM diagnostics lightweight: bounded resource samples, bounded MessageCache summaries, and LLM-turn size telemetry. It does not trigger heap snapshots or explicit GC on the allocation-failure path, avoiding additional allocation and pause risk when the process is already under pressure.

The follow-up fixes also ensure that output-safety truncation does not advertise unusable pagination cursors, exact-byte-limit output is accepted, abort is checked before records are yielded, cache incident summaries use bounded top-N selection, and failed incident capture can be retried.

yzxoi (approved):

Issue #742, opened 2026-07-21
[P1][area/tools] session_search/session_read bypass bounded history and retain multi-GB ArrayBuffers on latest dev


marzich:
## Summary

On the exact latest `dev` HEAD (`7e76f2c75d1f052ac7833d71d2156683e09033dc`, 2026-07-21 19:34 +08), the bounded model-history and message-page fixes do not cover `session_search`, `session_read`, or the legacy full-message API.

A small `session-historian` child repeatedly used those tools and caused hundreds of thousands of storage operations, multi-GB file/slab cache growth, and multi-GB Bun/JSC `ArrayBuffer` retention. After all observed session/tool/LLM spans ended, the main process remained about 2.5 GiB RSS above its clean post-restart baseline.

This run had no local memory patches. The only worktree difference from `origin/dev` is an unrelated sandbox environment-variable allowlist entry.

## Environment

- Ubuntu laptop, 16 GiB RAM
- Bun 1.3.14
- Synergy: exact `origin/dev` commit `7e76f2c7`
- `MemoryHigh=11G`, `MemoryMax=13G`, `MemorySwapMax=0`
- LSP disabled
- One Synergy Bun process in the service cgroup
- Service remained HTTP 200; no OOM or restart occurred in this sample

## Memory trajectory after clean restart

Service restart: 2026-07-21 19:38:44 CST.

| Time (CST) | cgroup | main RSS | anon | file | slab |
|---|---:|---:|---:|---:|---:|
| 19:40 | 1,198 MiB | 439 MiB | 391 MiB | 736 MiB | 67 MiB |
| 19:50 | 4,283 MiB | 1,504 MiB | 1,455 MiB | 2,404 MiB | 417 MiB |
| 19:56 | 7,862 MiB | 4,044 MiB | 4,003 MiB | 3,222 MiB | 640 MiB |
| 20:03 peak | 8,254 MiB | 4,350 MiB | - | - | - |
| 20:18, all spans closed | about 3,069 MiB | 2,967 MiB | - | - | - |

Runtime instrumentation:

- peak ArrayBuffers: **3,510 MiB**
- peak external: **3,679 MiB**
- after all spans closed:
  - RSS: **2,966.5 MiB**
  - external: **2,307.0 MiB**
  - ArrayBuffers: **2,180.9 MiB**
- cumulative application reads: **6,516 MiB / 569,335 read operations**
- no open session, tool, or LLM spans at the final sample
[...]

PR #743, opened 2026-07-21
fix(session): release history tool buffers

marzich:
## Summary
- preserve unrestricted `session_search` traversal while periodically reclaiming released history buffers
- make `session_read` hydrate only the requested message page instead of the full transcript
- preserve legacy delivery chronology without eagerly hydrating every message part
- request a full GC after history tools release their local references

## Changes
| File | Change |
|------|--------|
| `session/message-v2.ts` | Sort lightweight legacy message info, then stream parts one message at a time |
| `tool/session-read.ts` | Select offset/anchor windows before hydrating parts |
| `tool/session-search.ts` | Keep full search coverage, add abort checks and periodic/final buffer collection |
| `session/memory-pressure.ts` | Support interval-limited and explicitly forced full collection |
| focused tests | Cover bounded hydration, lazy legacy streaming, and full-GC requests |

## Verification
- 17 focused tests passed
- `bun run typecheck` passed across 12 packages
- `bun run format:check` passed
- `git diff --check` passed

Related to #742. This patch intentionally does not cap candidate sessions or messages searched.


yzxoi:
Thanks for the memory-pressure work here. I pulled the PR head into an isolated worktree and verified the main optimization direction, but found two issues that need fixing before merge:

1. `session_read` lost the previous per-message error isolation. The new page hydrator awaits `MessageV2.parts()` without a per-message `try/catch`, so one corrupt or non-externalizable part now fails the entire page. I reproduced this with one healthy message plus one attachment using an invalid data URL: `session_read` throws `Invalid data URL` instead of returning the healthy history and skipping the unreadable message.

1. The history tools call `forceFull: true` from `finally`. In Bun, the boolean passed to `Bun.gc(force)` means synchronous/forced collection, not “full collection”. This makes every `session_read` / `session_search` completion—including errors, cancellation, empty results, and invalid regex input—eligible to bypass the interval and synchronously stop the shared event loop. Concurrent completions are not coalesced, and `session_search` can also run a periodic collection immediately before the forced completion collection.

The ownership boundary should be:

- the history tool only signals that large history references have been released;
- `SessionMemoryPressure` (or one unified GC scheduler) owns whether collection is asynchronous or synchronous;
- the scheduler owns interval throttling and concurrent-request coalescing;
- synchronous GC remains reserved for actual critical memory pressure.

I’m preparing a focused fix with behavioral coverage for corrupt-message isolation, release-signal throttling/coalescing, and the corrected Bun GC semantics. I’ll push the update to this PR branch after the checks pass.

yzxoi:
Pushed the focused fix in `1af5c5b5`.

Changes:

- history tools now emit release signals instead of selecting `forceFull` / synchronous GC policy;
- `SessionMemoryPressure` coalesces release signals, keeps the existing minimum interval, uses asynchronous GC for normal collection, and reserves synchronous GC for measured critical pressure;
- `session_read` again skips an individually unreadable message instead of failing the entire page;
- added regression coverage for coalescing, interval throttling, sync-vs-async GC semantics, tool policy separation, and corrupt history reads.

Validation:

- focused history/memory tests: 18 passed;
- `packages/synergy` typecheck passed;
- repository `quality:quick` passed;
- `test:changed`: 3489 passed, 2 skipped, with one unrelated telemetry sampling failure that passed immediately in isolation;
- independent maintainability and performance reviews found no blockers.

yzxoi (approved):

PR #749, opened 2026-07-21
feat(plugin): add interaction hooks and agent calls

EricSanchezok:
## Summary

- add a single in-memory Composer document boundary with settled draft observers, revision-gated completions/decorations/edits, and serial pre-submit interception
- add App-wide safe text selection and host-owned text action infrastructure, plus canonical message slots and an asynchronous persisted user-message observer
- expose the Core abilities through Plugin API 3 contributions and capability-gated trusted UI/Host Services
- introduce bounded sessionless `AgentCall.text()`, migrate Experience Encoder, and expose constrained plugin-owned agent calls
- update plugin, frontend, LLM architecture, product, localization, and repository workflow documentation

## Core interaction contract

- `ComposerDocumentController` owns UTF-16 snapshots, atomic edits, revision invalidation, 700 ms IME-aware settled dispatch, ghost completion, decorations, undo-preserving DOM application, and abortable serial before-submit hooks
- ordinary Web Composer messages complete preflight before Session creation, history, optimistic state, draft clearing, or `session.input()`; shell, slash/backend commands, abort, and workflow launches keep their existing paths
- `TextSelectionController` distributes only `{ text }` after 250 ms, rejects over-limit and sensitive selections, and arbitrates DOM, Composer, Note, Monaco, and Terminal providers
- host-owned text action UI preserves native editing actions and provides keyboard/focus/error/pending behavior
- `session.user-message.after` runs after canonical persisted materialization across input, inbox, channel/agenda-backed inbox, and shell/backend command paths; `MessageV2` semantics exclude internal messages and observer failure never blocks the LLM loop
- canonical message shells expose additive `message.before`, `message.after`, and `message.actions` slots without replacing native rendering

## Plugin API and security

- add `composer.read`, `composer.write`, `composer.intercept`, `selection.read`, and `agent.call` capabilities to manifests, approval hashes, consent summaries, and risk classification
- add flat `ui.composerExtension`, `ui.selectionExtension`, `ui.textAction`, and `ui.messageSlot` contributions with descriptor/build validation
- trusted UI adapters are generation-, Scope-, Session-, Composer-, and message-aware and dispose stale registrations before remount
- text actions invoke existing UI-exposed command operations; no second transport or business result protocol is introduced
- Hook metadata enforces required capabilities at build/install and runtime; message content is redacted from lifecycle error logs
- plugin agent calls use the existing Host Service RPC envelope, require contribution-level approval, default to generation-owned hidden agents, enforce allowlists and hard input/output/runtime bounds, and expose no provider credentials or raw SDK access

## AgentCall

`AgentCall.text()` centralizes Agent/model resolution, explicit fallback handling, ephemeral invocation identity, tool-free `LLM.stream()`, combined cancellation/timeout/output limits, bounded text collection, stream disposal, and structured errors without persisting Session/message state or creating Cortex work.

Experience Encoder is the first migrated caller. Session title/summary, Smart Allow, GitHub Classifier, Agent generator, and other sessionless streams remain intentionally out of this PR and will be tracked after CI is green.

## Verification

Passed locally:

- focused Composer, selection, submit intent, AgentCall, Experience Encoder, plugin Host Service/capability/hook, and user-message materialization tests
- plugin and plugin-kit tests/build/validate/pack coverage
- App, UI, Synergy, plugin, and plugin-kit typechecks/builds
- SDK generation with no generated contract diff
- Lingui extraction, strict catalog compilation, and localization source validation
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run skill:check`
- `bun run package-guide:check`
- `bun run monorepo:check`
- `bun run package:check`

The Windows Core CI shard ran 1,440 tests (1,436 pass) and reproduced four pre-existing platform/timing failures: Cortex notification timing, temporary-directory `EBUSY`, attachment `FileTime`, and SessionSummary timeout. No new interaction/plugin/AgentCall/materialization test failed. Linux GitHub CI is the merge authority.


yzxoi:
Thanks for the substantial work here. The overall direction is good: the host-owned composer model, split capabilities, revision-checked edits, bounded `agent.call`, and unified sessionless Agent invocation are all useful additions.

I found five issues that should be addressed before merge:

1. **`AgentCall` timeout keeps the process alive** (`packages/synergy/src/agent/call.ts`). The old Experience Encoder timeout called `timer.unref?.()`, but the new shared `interruption()` timer does not. Because encoding is launched fire-and-forget, an in-flight encoder call can keep CLI/server shutdown waiting until the timeout expires.

1. **`session.user-message.after` observes internal child-session prompts** (`packages/synergy/src/session/user-message-materialization.ts`). The filter relies on `message.info.origin`, but several internal `SessionInvoke` callers do not set a canonical internal origin and therefore default to `origin: user`. This includes Cortex initial/repair prompts, Chronicler prompts containing conversation transcripts, and LookAt prompts. A plugin with `session.read` can consequently receive internal prompts despite the documented “does not run for synthetic/internal messages” contract.

1. **The selection context menu can use stale text** (`packages/app/src/context/text-selection.ts`, `packages/app/src/plugin/bridge.tsx`). `update()` waits 250 ms before replacing `#current`, while the synchronous `contextmenu` handler immediately reads `current()`. Right-clicking shortly after a new selection either hides the menu or invokes an action with the previous selection.

1. **Inline completion is visually misaligned after file pills** (`packages/app/src/components/prompt-input/prompt-input.tsx`). The ghost-text mirror uses `editor.documentText()`, which excludes file pills, while the real contenteditable DOM includes their visible width. A completion after a pill is therefore rendered left of the caret.

1. **Composer preflight edits are not transactional** (`packages/app/src/components/prompt-input/submit.ts`). Preflight handlers may apply edits directly to the real draft. If a later handler fails, times out, or the user cancels with Escape, the catch path returns without restoring the preflight snapshot, leaving partial plugin edits in the unsent draft.

Suggested coverage:

- real internal Cortex/Chronicler/LookAt paths do not emit `session.user-message.after`;
- a pending `AgentCall` timeout does not keep a subprocess alive;
- context-menu selection is current before the settle delay;
- completion positioning with a preceding file pill;
- preflight rollback when a later handler fails or the host cancels.

I ran the focused App/Core/Plugin/Plugin Kit tests, package typechecks, App and UI suites, format, and lint. Those checks pass. `test:changed` reached 2567 passing tests with one unrelated anonymous cleanup-hook timeout; the adjacent snapshot-isolation suite passes 8/8 when rerun alone.

yzxoi (approved):

Issue #752, opened 2026-07-22
Support private-network HTTP with unified browser capability fallbacks


marzich:
## Summary

Synergy Web should treat ordinary HTTP on a private overlay network as a supported deployment mode, not as an accidental browser configuration.

For example:

```text
http://100.x.x.x:18080
```

This is not a browser Secure Context. The immediate regression is an unconditional `globalThis.crypto.randomUUID()` call that crashes the SPA during module initialization, but the larger issue is that browser capability assumptions and fallbacks are currently distributed across individual components.

This issue tracks a unified compatibility approach. PR #753 is intentionally only the smallest first fix and should not close this issue.

## Immediate regression

`packages/ui/src/context/marked-math.ts` initializes a marker at module scope:

```ts
const generatedKatexMarker = globalThis.crypto.randomUUID()
```

`packages/app/src/app.tsx` imports the Markdown provider at module scope. In Chromium on private-network HTTP, `globalThis.crypto` and `crypto.getRandomValues()` are available while `crypto.randomUUID` is undefined. The resulting exception occurs before the App ErrorBoundary mounts and prevents the entire SPA from starting.

The unconditional call was introduced by commit `d93a657e` (`fix(ui): preserve Markdown math context and copy trust`). A second unconditional browser call exists in `packages/app/src/components/workspace/browser/browser-webrtc.ts`; that one fails when the browser client is constructed rather than during initial boot.

Observed error:

```text
TypeError: globalThis.crypto.randomUUID is not a function
```

## Broader design gap

The failure exposes several related problems:

1. Optional browser APIs are accessed directly from feature code, sometimes at module scope before error boundaries exist.
2. Non-security correlation IDs and security-sensitive nonces are both expressed as UUID generation, despite requiring different failure behavior.
3. Some call sites already implement local fallbacks while others assume Secure Context, producing inconsistent behavior.
4. Unsupported optional capabilities can crash the whole application instead of disabling only the affected action or panel.
5. Current browser validation does not exercise a real non-loopback HTTP origin, so localhost testing does not reveal Secure Context regressions.

### 1. Centralize capability-safe browser primitives

Provide small shared helpers for browser APIs used by App/UI rather than duplicating ad hoc checks. Keep the helpers focused; this does not require a large browser abstraction layer.

For ordinary UI identifiers:

1. Prefer `globalThis.crypto.randomUUID()` when available.
2. Otherwise generate UUID-v4-shaped output with `crypto.getRandomValues()`.
[...]

PR #753, opened 2026-07-22
fix(web): support UUID fallback over HTTP

marzich:
## Summary

- prevent the web app from crashing during module initialization when `crypto.randomUUID` is unavailable
- add a dependency-free browser UUID helper with native Web Crypto, `getRandomValues`, and lightweight fallbacks
- use the same helper for browser WebRTC connection IDs

Related to #752. This PR intentionally does not close the issue so the private-network HTTP behavior can continue to be tracked after the patch lands.

## Root cause

`marked-math.ts` generated a module-level marker with an unconditional `globalThis.crypto.randomUUID()` call. Browsers do not expose `randomUUID` to non-secure origins such as a private-network HTTP address, so importing the Markdown provider could fail before the application error boundary mounted.

## Scope

This is intentionally limited to the two unguarded browser calls. Server and desktop UUID generation is unchanged, and existing browser call sites that already have fallbacks are not refactored.

## Verification

- `bun test test/uuid.test.ts` in `packages/util` (4 passed)
- `bun test test/marked-math.test.ts` in `packages/ui` (9 passed)
- `bun test src/components/workspace/browser/browser-webrtc.test.ts` in `packages/app` (4 passed)
- util, UI, and app typechecks passed
- repository pre-push formatting, lint, and all-package typecheck hooks passed
- production frontend build passed
- built assets contain no exact unguarded `crypto.randomUUID()` call
- verified in Chromium over non-secure private-network HTTP with `isSecureContext === false`, no `randomUUID`, and no page errors


yzxoi (approved):
## Review Result: No Findings

Zero blocking defects introduced by this PR. The change is scoped correctly as a minimal Phase 1 fix, mechanically safe, and well-tested.

### What this PR does right

- **Layered fallback strategy** — `randomUUID()` → `getRandomValues()` v4 → `Math.random()` fallback covers the private-network HTTP scenario without over-engineering.
- **Dependency injection** — `UUIDEnvironment` makes every code path testable without global mocking.
- **Correct UUID v4 bit-setting** — RFC 4122 version and variant bits are properly applied in the `getRandomValues` path.
- **Consumer substitutions are safe** — `connectionId` is a ticket-authenticated signaling discriminator, not a security boundary. KaTeX marker only controls clipboard copy trust. Both tolerate weak randomness.
- **No package or export regressions** — CI Package Validation confirms `dist/uuid.js` + `dist/uuid.d.ts` build correctly, pass `publint` and `arethetypeswrong`.

### Non-blocking observations

1. **Helper name contracts UUID format** — the final fallback returns a non-UUID shaped string. Both current consumers are fine, but a future caller expecting RFC UUID format would break silently in exotic runtimes.
2. **Two cheap test gaps** — no parameterless `generateUUID()` test exercising `globalThis.crypto`, and no double-throw test for the fallthrough path.
3. **Pre-existing regex gap** (not in this PR): `stripGeneratedKatexMarker` at `marked-math.ts:13-15` uses a `\s` anchor that a `\"data-...` adjacent attribute bypasses. Worth a separate hardening PR.

### Overall

This is a clean, well-scoped fix that addresses the reported startup crash and WebRTC construction failure over HTTP. The architecture is consistent with the Phase 1 plan in #752.

Thank you @marzich for the contribution — well-researched, carefully scoped, and cleanly implemented.

PR #756, opened 2026-07-22
fix: polish Settings interactions

EricSanchezok:
## Summary

- hide the entire Usage attention section when no provider accounts require credential recovery
- align shared numeric input typography and clarify model quick-switch copy
- show complete model-role consumer details in an interactive, scrollable, visually balanced popover
- replace the permanent Synergy Link target form with a compact responsive dialog

## Validation

- `bun run quality` — 14/14 tasks passed; core suite 5071 passed, 2 skipped, 0 failed
- focused Settings regression tests
- isolated Web visual verification at desktop and narrow widths
- independent maintainability review


yzxoi:
Review findings before the follow-up fix:

1. **The add-target dialog can trap the entire app if the create request hangs.** While the request is pending, Escape/outside-click are blocked and both Close and Cancel are disabled. The SDK call has no cancellation signal or reliable hard timeout, so a stalled request leaves no way to exit the modal without reloading or closing the app.
2. **The add-target dialog body gets double padding.** The new Settings override has lower specificity than the shared Dialog body rule, so the shared `0 24px 24px` padding remains in addition to the fields/footer padding.
3. **The model-role detail popover can clip its fallback/resolution content on short viewports.** The outer surface has a viewport-constrained max height plus `overflow: hidden`, while only the agent subsection scrolls; content below that subsection can become unreachable.

I’m adding focused regression coverage and will push fixes to this PR branch.

yzxoi (approved):

Issue #760, opened 2026-07-22
feat: enhance Performance dashboard with cgroup v2 memory metrics and layered memory attribution


yzxoi:
## User report

[...]

PR #763, opened 2026-07-22
feat(performance): add Linux memory attribution and release controls

marzich:
## Summary
- make Linux cgroup v2 memory the authoritative service-memory view while preserving the existing process-RSS behavior on macOS, Windows, and non-cgroup hosts
- expose Linux cgroup working-set/reclaimable layers, PSI and memory.events deltas, plus sampled JSC/mimalloc ownership data
- attribute Linux bash and external-agent child processes back to session/message/tool owners without retaining prompts, command text, or output
- collect at Linux session release boundaries under pressure and provide an explicit opt-in cgroup cache reclaim path

## Safety boundaries
- no prompt compression or concurrency reduction
- no service-limit or restart-policy changes
- Linux-specific sampling and owner tracking are no-ops on other platforms
- synchronous collection is limited to Linux release boundaries under soft pressure; active streams remain asynchronous
- cgroup `memory.reclaim` is disabled by default and requires `SYNERGY_LINUX_MEMORY_RECLAIM_ENABLED=true`

## Changes
| Area | Change |
|------|--------|
| Linux cgroup sampler | Reads service total, anon/file/slab layers, working-set estimate, reclaimable estimate, limits, peak, swap, pressure and event deltas |
| Runtime attribution | Samples JSC heap/extra memory, mimalloc committed/reserved memory, top object types and positive type deltas |
| Child ownership | Associates Linux bash/Codex/Claude Code/OpenClaw processes with session, message, call, tool and trace identifiers |
| Performance UI/API | Shows service-wide memory, reclaimable layers, pressure signals, runtime owners and owned child processes |
| Release handling | Measures GC effectiveness; at Linux release boundaries uses synchronous GC only under pressure and offers bounded opt-in cache reclaim |
| Tests/docs | Covers Linux behavior, non-Linux preservation, migrations, aggregation, tracking, PSI decisions and reclaim gating |

Addresses #760

## Verification
- `bun test test/session/memory-pressure.test.ts` (12 passed)
- 67 focused Synergy observability, process, external-agent and memory-pressure tests
- 13 focused Performance dashboard tests
- `bun run format:check`
- `bun run lint`
- `bun run typecheck` (11 packages)
- `bun run monorepo:check`
- `bun run --cwd packages/app i18n:check`
- `bun run --cwd packages/sdk/js build`
- one-shot Linux cgroup/JSC sampler smoke check

The local `secrets:check` command could not run because `gitleaks` is not installed on this host; GitHub CI remains the authoritative secret scan.


yzxoi (changes requested):
Thanks for working on improving service-level memory visibility. I agree that cgroup v2 can be a useful source of telemetry, but I’m requesting changes because the current sampling and ownership model can both waste significant resources and report misleading data.

## 1. The five-second sampler synchronously walks the cgroup tree and every process

`ObservabilityResources.snapshot()` calls `ServiceMemory.sample()` on the default five-second interval. That call synchronously:

- traverses the current cgroup and all descendant cgroup directories;
- reads `cgroup.procs` from each directory;
- reads `/proc/<pid>/status`, `/proc/<pid>/comm`, and `/proc/<pid>/smaps_rollup` for every discovered PID.

All of this uses synchronous filesystem calls on the server’s main event loop. `smaps_rollup` is especially expensive because the kernel must aggregate the process’s mappings to calculate PSS. As the number and memory size of processes grow, this creates periodic event-loop stalls and can directly affect HTTP, SSE, LLM streaming, and tool/event processing. The observer can therefore perturb the same event-loop-lag telemetry it is collecting.

This becomes much worse if the process is in a broad cgroup: every five seconds Synergy may traverse a large part of the host hierarchy and inspect many unrelated processes.

PSS attribution should not run synchronously in the five-second main-loop sampler. It needs a lower-frequency or asynchronous/worker-based path, plus explicit bounds for process count, traversal size, and elapsed sampling time. If a previous sample is still running, the next one should be skipped rather than accumulated.

## 2. The current cgroup is not necessarily the Synergy service boundary

The implementation reads `/proc/self/cgroup` and treats the resulting cgroup as “the Synergy service.” That assumption is valid for the systemd daemon, where `synergy.service` receives a dedicated cgroup, but it is not generally valid for foreground or ephemeral launch modes such as `bun dev server`, a foreground `synergy server`, or `synergy send`.

Those processes may share a terminal/session cgroup, user slice, container root cgroup, or another supervisor-owned boundary with unrelated processes. In that case:

- `memory.current` is not Synergy’s service memory;
- pressure/OOM counters are not Synergy-specific;
- unrelated PIDs are persisted and displayed as `service-child` processes;
- child counts and RSS/PSS totals become misleading;
- a root/shared cgroup can turn the sampler into a system-wide process scan.

The implementation needs to prove that the cgroup is Synergy-owned before presenting it as a service boundary. For launch modes without a known dedicated cgroup, it should fall back to the existing main-process plus registered-child accounting rather than promote the current cgroup to service scope.

## Is this PR useful in its current form?

The feature is useful if Synergy has a reliable service ownership boundary and the attribution sampler is low-overhead. Under those conditions, `memory.current`, cgroup pressure/OOM events, and occasional PSS attribution would provide valuable operational visibility that process RSS cannot provide.

In the current form, however, the headline number can be systematically wrong in common launch modes, while the five-second synchronous PSS scan can introduce latency into the service being observed. That makes the telemetry difficult to trust and potentially more expensive than the value it provides.

I suggest first defining which runtime modes have a Synergy-owned cgroup, using `process-sum` elsewhere, and moving bounded PSS attribution off the five-second event-loop path. With those constraints, the rest of the schema/API/dashboard work in this PR would have a meaningful and reliable foundation.

Issue #764, opened 2026-07-22
P0: streamed tool-call transition can spin the Bun event loop and leave HTTP unresponsive


marzich:
## Summary

Synergy can enter a permanent main-thread CPU spin while handling a streamed tool call. The TCP listener remains open, but HTTP requests connect and receive no response. This is not an OOM or Linux memory-pressure event: the Bun main thread consumes one full CPU core and cannot process `SIGTERM`, so systemd must send the final kill after `TimeoutStopSec`.

## Incident evidence

- Incident: 2026-07-22 14:55:47 CST
- Session: `ses_0a428be0affembGg1aWZGDQEgu`
- Assistant message: `msg_f889b88a3001IFSjyaEmxtwJtz`
- Tool call: `call_DEt2Hmptkgs4Beq4ccxW5nQ` (`bash`)
- Last durable state: tool part remained `generating`
- Last backend log: `tool.stream.tool_call.received`
- Missing next log: `tool.stream.tool_call.part_running`
- No child process was spawned for the bash call

The preceding four `view_file` calls completed normally in the same response. The failing bash input was small and ordinary:

- raw input: 514 characters / 514 bytes
- parsed top-level keys: `command`, `workdir`, `description`
- standalone parse: <0.01 ms
- standalone stringify: <0.01 ms
- reconstructed `MessageV2.Part.parse`: about 1.6 ms

This narrows the stall to the synchronous path between `tool.stream.tool_call.received` and the durable `running` part update, including provider/pending metadata merge, input normalization/measurement, schema validation, and `Session.updatePart` serialization.

## Why this is not OOM

At the time the service was unresponsive:

- cgroup memory: about 3.19 GB
- main-process RSS: about 2.58 GB
- `memory.high`: 0 events
- `memory.max`: 0 events
- `oom`: 0
- `oom_kill`: 0
- swap: 0
- memory PSI `some`/`full` avg10: 0.00
- kernel journal: no OOM, hung-task, GPU-reset, or thermal event

In contrast, the Bun main thread remained at approximately 95-101% of one CPU core. Internal resource telemetry stopped at 14:55:42 because the event loop stopped progressing, while an out-of-process cgroup watcher continued sampling normally. HTTP requests connected immediately but timed out with zero response bytes, and receive queues/CLOSE-WAIT sockets accumulated.

[...]

PR #765, opened 2026-07-22
fix(session): bound streamed tool input finalization

marzich:
## Summary
- prefer already bounded streamed JSON when finalizing tool-call input
- avoid re-traversing the provider final input when a complete streamed object is available
- add stage breadcrumbs that identify whether a future stall occurs in input normalization, metadata merge, or persistence without logging tool content

## Incident context
A Linux production incident left the Bun main thread at roughly one full CPU core while HTTP accepted connections but could not serve requests. The last durable transition was `tool.stream.tool_call.received`; the corresponding tool never reached `part_running`, no child process spawned, memory was below cgroup limits, and PSI/OOM counters were clean.

This patch addresses the bounded-input traversal risk and improves localization. It does not claim that every possible JSC synchronous-spin source is resolved.

## Changes
| File | Change |
|---|---|
| `packages/synergy/src/session/processor.ts` | Reuse bounded streamed JSON and emit content-free transition breadcrumbs |
| `packages/synergy/src/session/tool-input.ts` | Parse completed streamed JSON into a plain record |
| `packages/synergy/test/session/processor.test.ts` | Verify a hostile provider input is not traversed when bounded streamed JSON exists |
| `packages/synergy/test/session/tool-input.test.ts` | Cover valid and fallback parsing paths |

## Verification
- `bun test test/session/tool-input.test.ts test/session/processor.test.ts` (46 passed)
- pre-push Prettier, oxlint, monorepo typecheck, and sherif checks passed

Relates to #764.


yzxoi:
## Review follow-up

The bounded streamed-input direction is useful: raw tool argument deltas already have an incremental byte limit, so they are appropriate for progress display, memory accounting, and diagnostics. The new stage breadcrumbs are also valuable for continuing the #764 investigation.

The important ownership distinction is:

- `tool-input-delta` / the accumulated raw string is transport and progress data;
- the AI SDK's final `tool-call.input` is the canonical input for that call and is the value that continues into permission checks, enforcement, and execution.

Using `JSON.parse(accumulatedRaw)` as persisted `state.input` can make the running part, loop guards, and replay history describe a different input from the one that actually executes whenever an adapter's raw stream and final tool-call event differ. That invariant should not depend on provider-specific stream behavior.

### Updated design

1. Keep the AI SDK final `tool-call.input` canonical for the final serialized-input bound, persisted tool part, doom-loop checks, permission evaluation, and execution.
2. Keep streamed raw arguments only for incremental byte limits, memory accounting, progress state, and diagnostic byte counts.
3. Preserve the final canonical-input bound for final-only providers and for cases where the final input is larger than the streamed representation.
4. Test an intentionally different raw/final pair and assert that the running part persists the AI SDK final input.
5. Test that a small raw stream cannot hide an oversized final AI SDK input.

I updated this PR accordingly:

- removed `SessionToolInput.fromStream()` as a canonicalization path;
- restored AI SDK final input as the persisted and loop-guard input;
- retained incremental raw-stream bounds and raw memory accounting;
- final-bounded the canonical AI SDK input and exposed both canonical and streamed byte counts in the breadcrumb;
- added the two behavior tests above;
- documented the ownership invariant in `docs/architecture/llm-loop.md` and the `integrate-llm` development workflow.

### Verification

- `bun test test/session/processor.test.ts test/session/tool-input.test.ts` — 45 passed
- `bun run typecheck` — passed (11 packages)
- `bun run skill:check` — passed

The remaining #764 work is separate from canonical input ownership: provider metadata still needs an explicit bounded-copy contract, and deeper schema/storage/event breadcrumbs belong inside the persistence owner if the incident needs further localization.

yzxoi (approved):

PR #772, opened 2026-07-22
feat(plugin): complete API3 MCP host contracts

EricSanchezok:
## Summary

- add shared strict MCP declarations to Plugin API 3 and atomically register namespaced plugin MCP servers through the host lifecycle
- add bounded task, asset, shell, CLI, hook, and trusted UI settings contracts required by real plugin migrations
- harden OAuth ownership against stale generations and preserve plugin contribution lifecycle, approval, and execution boundaries

## Verification

- `bun test test/mcp/auth.test.ts test/mcp/oauth.test.ts test/mcp/supervisor.test.ts test/plugin/cli-command.test.ts test/plugin/contribution-registry.test.ts test/plugin/shell-host-service.test.ts test/plugin/asset-create-host-service.test.ts test/plugin/system-transform-hook.test.ts test/plugin/task-run-host-service.test.ts` from `packages/synergy` (117 passed)
- `bun test test/mcp.test.ts test/host-contract.test.ts` from `packages/plugin` (20 passed)
- `bun test src/plugin/surface-settings.test.ts` from `packages/app` (3 passed)
- native Synergy and App typechecks
- `bun run quality:quick`
- push hook: formatting, lint, and monorepo typecheck


yzxoi:
Quick follow-up review found two issues that should be addressed before merging:

1. **P1 — `shell.run` bypasses the normal approval and sandbox pipeline.** `runPluginShell()` only rejects `deny`, so a `guarded` profile decision of `ask` proceeds without approval. It also prepares the process with `sandboxMode: "none"`, bypassing the configured workspace sandbox, protected paths, and external-write containment. This should fail closed for non-`allow` decisions and execute with the policy returned by `EnforcementGate`.

1. **P2 — plugin settings can retain values across Scope switches.** `SettingsSectionContent` sends the correct `scopeID`, but its `createResource` source is only `pluginId`. When the same plugin is active in another Scope, the resource does not refetch; stale values can be displayed and then saved into the new Scope. The resource key should include both `pluginId` and `scopeId`, with a regression test for switching Scopes.

The MCP/OAuth generation ownership, atomic plugin MCP replacement, CLI namespace, bounded task/asset services, and system-transform work otherwise look well structured. I’m preparing a focused follow-up commit for the two issues above.

yzxoi (dismissed):

yzxoi (approved):

PR #773, opened 2026-07-22
feat: converge blueprint implementation routes

EricSanchezok:
## Summary

- converge materially different Blueprint routes from evidence and user-owned decisions before execution
- require the shared eight-section Blueprint contract in Plan and Lattice while preserving distinct synergy and synergy-max guidance
- make Blueprint execution and independent review enforce trajectory conformance in addition to outcome completion
- add behavioral coverage for authoring, execution, and audit prompt contracts

## Testing

- `bun test test/session/plan-prompt-contract.test.ts test/session/workflow-user-wrapper.test.ts test/server/blueprint-route.test.ts test/session/blueprint-continuation.test.ts`
- `bun run quality:quick`


yzxoi:
Thanks for pushing the Blueprint route-convergence direction. After reviewing the full flow, I think the current PR still mixes three different lifecycle concerns: planning-time clarification, execution-time contract enforcement, and audit-time rejection.

The main issue is not only what questions the prompts ask, but **when questioning must end and when the Blueprint becomes a frozen execution contract**.

## Problems to address

### 1. Planning currently encourages iterative questioning

The updated prompts mostly move toward resolving implementation choices from evidence, but the coding Plan prompt still explicitly allows a repeated `investigate → ask → continue → ask again` cycle.

For an autonomously executable Blueprint, clarification should instead have one explicit checkpoint:

1. investigate the codebase, requirements, and existing constraints;
2. collect every remaining user-owned material fork;
3. ask them together in one `question` call;
4. incorporate the answers and freeze the Blueprint;
5. do not leave later material questions to the executor.

The `question` tool already supports multiple questions in one call, so the prompts and workflow policy should make that batching behavior explicit.

### 2. Lattice auto must not ask after its initial clarification checkpoint

`firstBlueprintStarted` currently disables `question` after the first BlueprintLoop starts, but this still permits repeated questions before that boundary and gives later Lattice steps no valid outcome when they discover a new user-owned material fork.

For auto mode, questioning should only be available during one initial clarification checkpoint. After it closes, later steps must not ask. If a later step cannot produce a decision-complete Blueprint from the frozen decisions and available evidence, the Lattice run should pause with the current step blocked rather than guessing, emitting an incomplete Blueprint, or consuming continuation/model-call budget indefinitely.

Collaborative mode can still expose a clarification/review checkpoint before a step Blueprint is executed, but questions should not leak into ordinary Blueprint execution.

### 3. Blueprint execution needs an explicit block operation

The PR tells the executor to “surface the blocking conflict” when the selected route is missing, ambiguous, conflicting, or infeasible, but there is currently no runtime action that implements this instruction. A normal assistant reply leaves the loop running, so continuation can wake it again.

We need an executor-only `blueprint_loop_block` action that:

- persists a structured conflict and required resolution;
- transitions the loop out of `running` and stops automatic continuation;
- surfaces the conflict in the UI;
- pauses the owning Lattice run and marks its current step blocked when applicable;
- does not start an audit.

We can reuse the existing non-terminal `waiting` lifecycle status and attach structured block data, rather than adding another parallel BlueprintLoop status. The existing bare `wait` / `resume` endpoints are not sufficient because they store no reason or resolution and the Web composer currently submits a waiting loop through `loop.start` again.

### 4. Block, reject, and cancel must have distinct meanings

- **Block**: the frozen contract cannot currently be executed; external input, environment recovery, or Blueprint revision is required.
- **Reject**: the Blueprint is valid, but the implementation is incomplete, incorrect, unverified, or route-nonconformant in a way the executor can fix under the same contract.
- **Cancel**: the run is abandoned or superseded and will not continue.

If resolving a block only supplies a credential, restores an external dependency, or clarifies a non-material detail, the same loop may resume.

If the resolution changes a material owner, architecture, data flow, public contract, migration, compatibility strategy, artifact shape, or user-visible behavior, the old loop must not simply resume. It should be cancelled as superseded, the Blueprint should be revised, and a new loop should start from the new frozen contract.

### 5. Execution and audit need an immutable Blueprint baseline

The loop stores `noteVersion`, but execution and audit still read the current Note. If the Blueprint is edited while the run is blocked, the executor may have worked against one version while the reviewer audits another.

The loop therefore needs an immutable contract snapshot or an equivalent versioned-read mechanism. Both execution and audit should use the contract bound when that loop was created. A materially revised Blueprint should create a new loop and audit baseline.

The optional start instruction should also be limited to incidental execution context. It must not silently become a second material contract that can override or contradict the frozen Blueprint route.

## Planned revision

We are going to revise this before merging by:

1. replacing the iterative-question language with a single batched clarification checkpoint;
2. closing the question capability after that checkpoint, with stricter behavior for Lattice auto;
3. adding a structured `blueprint_loop_block` path that reuses the existing waiting lifecycle and stops continuation;
4. wiring blocked BlueprintLoops into Lattice step blocking and run pausing;
5. adding an explicit block-resolution path in the API/UI instead of restarting a waiting loop;
6. separating fixable audit rejection from contract-level blocking;
7. distinguishing same-contract resume from material Blueprint revision and loop supersession;
8. binding execution and audit to an immutable Blueprint contract snapshot;
9. adding behavioral tests for batched clarification, auto-mode question gating, block/resume/cancel, Lattice pause, audit decisions, and contract-version consistency.

I’ll update the implementation and tests in a follow-up revision. No code changes are being made as part of this review comment yet.

yzxoi:
The latest prompt revision addresses the clarification issue I raised: Plan now completes investigation first, batches the remaining user-owned material forks into one `question` call, and closes the clarification checkpoint afterward. The Lattice prompts also make the intended auto/collaborative boundaries much clearer.

One scope note before merging: the current implementation no longer adds execution-time or audit-time trajectory-conformance enforcement—the latest commit reverted those BlueprintLoop executor/reviewer prompt changes and their tests back to outcome completeness. The PR body still says that execution and independent review enforce trajectory conformance and that the PR adds execution/audit prompt-contract coverage, so that description is now broader than the code being merged.

That is acceptable for this revision. I’m planning to merge the current authoring/clarification changes as-is and handle structured Blueprint blocking, immutable execution/audit contract baselines, and route-conformance enforcement in a follow-up change. We should update the PR summary accordingly so the merged scope is explicit.

yzxoi (approved):

Issue #777, opened 2026-07-22
[P1][Linux][area/session] Bound native/external memory for high-concurrency workflows without reducing concurrency


marzich:
## Summary

Synergy on Linux can accumulate native/external memory across many settled LLM/tool turns and approach the service memory limit even when the number of visible primary sessions is small.

The production requirement is unchanged:

- support at least 10 concurrent subagents;
- support at least 5 concurrent worktrees;
- preserve full prompt/context behavior;
- do not solve the problem by lowering concurrency;
- keep the service out of swap.

This issue is the Linux-specific optimization umbrella. Cross-platform ownership fixes remain separate in #774, #775, and #776. Linux-only GC, allocator, cgroup, and process-recycling behavior should be implemented here without changing tuned macOS/Windows behavior.

## Environment and current controls

Observed host:

- Ubuntu, 16 GiB RAM;
- Bun 1.3.14;
- `MemoryHigh=11G`;
- `MemoryMax=13G`;
- `MemorySwapMax=0`;
- service cgroup v2;
- LSP disabled for the measured Synergy process.

Bun 1.3.14 is currently the latest release, so a simple runtime upgrade is not presently available. Bun also has an open report of Linux RSS growth behaving worse than macOS under equivalent long-running child workloads:

https://github.com/oven-sh/bun/issues/21560

That report supports a platform-specific investigation, but it is not proof that all Synergy retention is a Bun bug.

## Current Linux evidence

A representative sample showed:

- cgroup current: approximately 2.6-2.9 GB;
- main Bun PSS/RSS: approximately 1.5 GB;
- cgroup working set: approximately 1.5 GB;
- reclaimable file cache + slab: approximately 1.2 GB;
- process external memory: approximately 783 MB;
- process ArrayBuffers: approximately 638 MB;
- cgroup peak for the process lifetime: approximately 5.24 GB;
- swap: 0;
- no cgroup high/OOM events in that sample.

Therefore the headline cgroup value mixes two different problems:

1. real process anonymous/native memory;
2. reclaimable Linux page cache/slab charged to the service.

The cache component affects headroom but does not explain the live process ArrayBuffers.

The UI appeared to have one active primary session, but the backend simultaneously had an OpenAI Codex parent stream and a DeepSeek child/subagent stream. Visible session count is not an execution-load metric.

[...]

Issue #776, opened 2026-07-22
[P2][area/session] Add a deterministic cross-platform memory attribution harness for provider streams and session turns


marzich:
## Summary

Memory regressions are currently diagnosed from production workflows, where provider latency, prompt size, tool activity, background jobs, and OS accounting overlap. This repeatedly narrows the problem to `processor.process` or native `ArrayBuffer` memory but cannot identify the exact ownership layer.

Add a deterministic, offline, cross-platform attribution harness that replays the same recorded SSE fixture through progressively larger portions of the stack.

This is separate from the Linux cgroup dashboard work in #760. The harness should work on every supported platform and answer which software layer retains memory.

## Why session count is not a useful benchmark unit

Ten visible sessions can be cheaper than one long workflow if most sessions are idle. Memory must be correlated with:

- active provider streams;
- completed LLM/tool turns;
- request and raw response bytes;
- background `LoopJob` overlap;
- tool child processes;
- post-turn retained footprint.

A repeatable fixture is needed to compare macOS, Linux, and Windows under equivalent work.

## Proposed layered fixture

Use one sanitized, recorded provider response and run a small fixed number of sequential iterations through four layers:

### Layer A: runtime transport

```text
Bun fetch / local HTTP fixture -> drain bytes -> terminal cleanup
```

This isolates Bun fetch, ReadableStream, decoder, and allocator behavior.

### Layer B: provider adapter

```text
transport -> provider SDK normalization -> drain provider events
```

This identifies provider-specific SSE parsing, chunk conversion, and response ownership.

### Layer C: AI SDK

```text
provider adapter -> streamText/fullStream -> consume -> dispose
```

This identifies `tee()`, step/result recording, request-body retention, and AI SDK transforms.

### Layer D: Synergy turn

```text
AI SDK -> SessionProcessor -> persistence/tool event handling -> turn release
```

This identifies Synergy message parts, metadata, processor maps, observability, and background-job references.

No live provider call, Codex task, broad build, or stress load is needed.

## Required measurements

At each layer and iteration, capture:

- `Bun.unsafe.memoryFootprint()` with RSS fallback;
- `process.memoryUsage()`: RSS, heap, external, ArrayBuffers;
- JSC heap/extra/object counts when available;
- mimalloc committed/reserved/abandoned pages when available;
- active stream/reader/job counts;
- input bytes, output bytes, and chunk count;
- timestamps at start, terminal settlement, +1s, +5s, +30s, and after an explicit full GC.

[...]

Issue #775, opened 2026-07-22
[P1][area/session] ArrayBuffer retention persists after fullStream residual-branch cancellation and disposal


marzich:
## Summary

The current `dev` stream-ownership fix correctly cancels AI SDK's residual `baseStream` branch and awaits that cancellation from `SessionProcessor`. Despite that fix, settled turns still show a positive native/external `ArrayBuffer` slope.

This is a follow-up to #613 and #615, not a report that their fix is missing. The remaining work is to define and enforce a complete cross-platform ownership contract across provider response, AI SDK result, stream branches, step/request metadata, and processor cleanup.

Inspected at `origin/dev@cdb9b2babffadcc4465c05a7abb4534e462731a7`. Measurements were collected on Linux with additional telemetry only.

## Existing cleanup is present

`packages/synergy/src/session/llm.ts`:

- reads AI SDK's retained `baseStream`;
- calls `cancel()` immediately;
- returns an owned stream wrapper;
- exposes `dispose()` that waits for cancellation.

`packages/synergy/src/session/processor.ts` calls `await ownedStream.dispose()` in `finally`.

That is necessary and should remain. The new evidence shows it is not sufficient to return native/external memory to a stable post-turn baseline.

## Current evidence

Over a two-hour observation window, considering only turns whose maximum observed global active-stream count was one:

| Provider | Settled turns | Median ArrayBuffer delta at `turn.released` | P95 | Median request size |
| --- | ---: | ---: | ---: | ---: |
| `openai-codex` | 65 | +5.26 MiB | +40.96 MiB | 1.10 MiB |
| `deepseek-lamas` | 14 | +27.76 MiB | +56.48 MiB | 0.09 MiB |

The DeepSeek path retained more memory despite much smaller requests. This makes provider transport/decoder behavior, raw chunking, AI SDK normalization, and terminal stream cleanup stronger suspects than prompt size alone.

A forced-GC incident snapshot reported approximately:

- process ArrayBuffers: 844 MiB;
- process external memory: 933 MiB;
- ArrayBuffer self-size visible in the JS heap snapshot: about 33.6 MiB, including one 32 MiB WebAssembly memory.

The large difference means the final owner is mostly in native/external backing stores or runtime accounting that a normal JS dominator graph cannot attribute.

## Relevant upstream behavior

Synergy currently uses:

- `ai@5.0.212`;
- `@ai-sdk/openai@2.0.111`;
- `@ai-sdk/openai-compatible@1.0.42`.

[...]

Issue #774, opened 2026-07-22
[P1][area/session] Non-blocking LoopJob jobs retain full session histories and lack lifecycle accounting


marzich:
## Summary

On current `dev`, non-blocking `LoopJob` work can keep an entire session history reachable after the foreground turn has moved on.

The API already has a `JobInstance` type, but `LoopJob.execute()` uses an instance only to resolve the job type and then calls `job.execute(ctx)`. Every detached job therefore receives the full `LoopJob.Context`, including `messages: MessageV2.WithParts[]`.

This is a cross-platform object-lifetime problem. It is not a request to reduce context, compress prompts, or lower concurrency.

Inspected at `origin/dev@cdb9b2babffadcc4465c05a7abb4534e462731a7`.

### Detached jobs retain the full Context

`packages/synergy/src/session/loop-job.ts`:

- `Context.messages` contains the complete loop working set.
- Non-blocking jobs run fire-and-forget with `job.execute(ctx).catch(...)`.
- There is no central active-job registry, per-session deduplication, retained-byte accounting, or completion barrier.
- Fields collected in `JobInstance` are not passed to `execute`, so jobs cannot naturally carry a small immutable payload.

### Summary keeps the history alive

`packages/synergy/src/session/summary.ts` registers a non-blocking post job and passes:

```ts
messages: ctx.messages.slice()
```

The shallow copy still keeps every message, part, tool result, metadata object, and referenced string reachable while summary work runs. The first active summary can overlap subsequent loop turns.

### Chronicler creates additional full-history representations

`packages/synergy/src/library/chronicler.ts`:

1. receives the full `ctx.messages`;
2. converts it to model messages;
3. joins it into one `conversationText` string;
4. starts another child session;
5. attaches an abort listener that is not explicitly removed after normal completion.

A compaction boundary can therefore temporarily retain the original history, converted messages, a large joined string, and a child invocation at the same time.

## Impact

For long tool workflows, the issue is driven by number of turns and history size rather than visible top-level session count. A background job that overlaps the next turn extends the lifetime of the previous turn's full object graph and can amplify stream/native-buffer retention.

This is a strong retention amplifier, but it is not claimed to be the sole owner of previously observed multi-GiB native ArrayBuffers.

## Proposed design

1. Change job execution to receive the collected instance:

```ts
execute(instance: JobInstance, ctx: LightweightContext)
```

[...]

PR #780, opened 2026-07-22
test(performance): add stable cross-platform memory workloads

marzich:
## Status

This is a draft benchmark/experiment PR. It adds stable workloads and measurement output; it does not change production Session behavior and does not claim that the memory issues are resolved.

Related to #774, #775, #776, and #777. This PR intentionally does not close any of them. Runtime lifetime changes remain separate in #781.

## Summary

- Keep the bounded process-level history-projection and tool-stream microbenchmarks.
- Replay an anonymized completed Synergy trajectory through a fresh temporary server, Session API, deterministic local provider, tools, subagents, persistence, and deletion.
- Add fixed `trajectory`, `parallel`, and `sequential` runtime scenarios.
- Measure server RSS, heap, external memory, ArrayBuffers, Synergy child metrics, and complete process-tree RSS on Linux, macOS, and Windows.
- Record Git revision, dirty state, workload contract version, and a workload-semantic SHA-256 fingerprint.
- Emit stable phase peaks, post-deletion checkpoints, and 5/30/120-second release samples as raw JSON.

## Stable workloads

| Scenario | Standard workload | Purpose |
| --- | --- | --- |
| `trajectory` | 1 primary Session and 4 parallel background subagents; 97 source messages and 220 tool calls | Real parent/subagent lifecycle |
| `parallel` | 5 concurrent primary Sessions, 5 workers; 60 persisted messages and 225 tool calls | Concurrent primary Session cost |
| `sequential` | 5 complete primary Session lifecycles, deleting each before the next | Cross-turn/session retention slope |

The primary scenarios replay the real trajectory's 10-round heavy primary turn. The source turn ended at its step limit, so each standalone replica receives one fixed minimal terminal response. That response is included in the workload descriptor and fingerprint. Historical `task` calls are mapped to the byte-preserving payload tool in these scenarios to isolate primary concurrency from subagent fan-out.

`--scenario all` runs the three scenarios in separate temporary Synergy processes so one workload cannot contaminate the next. Smoke uses smaller fixed replica counts only to validate the harness.

## Fixture and safety

The structural fixture preserves roles, ordering, parentage, relative timing, terminal status, tool type/status, and payload sizes. Prompts, outputs, paths, runtime IDs, provider/model names, and credentials are excluded and replaced with deterministic byte-preserving data.

Every run creates and removes a fresh temporary Synergy home and workspace. It uses only a loopback server, deterministic local OpenAI-compatible provider, and loopback payload tool. No real provider, external Agent, credential, production Session, or external network request is used. This is bounded measurement, not a stress test.

## Cross-platform and cross-version contract

The workload fingerprint is based on fixture identity, scenario, preset, replica count, aggregate messages/tools/bytes, worker settings, timing, and adapter semantics. It does not include the tested source commit. Results are comparable only when `workload.contractVersion` and `workload.fingerprint` are identical.

Each result records the tested revision and whether the worktree was dirty. Internal implementation changes remain comparable while the represented LLM/Session/tool interaction contract stays unchanged. If workload semantics change, the fingerprint changes and the old and new result must not be treated as the same benchmark.

Process-tree RSS uses Linux procfs, macOS `ps`, or Windows PowerShell/CIM. It is a sum and can double-count shared pages, so it is an attribution metric rather than physical working set. Linux cgroup/service-memory fields remain supplemental platform-specific evidence.

## Linux verification

- `bun run typecheck`: 11/11 tasks passed
- `bun run lint`: passed
- repository Prettier check: passed
- `bun run perf:tools:test`: 4 tests passed, 39 assertions
- smoke matrix: all three isolated scenarios passed and removed their temporary runtimes
- standard trajectory: full 97-message/220-tool replay passed
- standard parallel: 5/5 replicas passed; each verified 12 messages and 45 tools
- standard sequential: 5/5 lifecycle checkpoints passed

Clean standard parallel result at `7a9ff644`:

- workload fingerprint: `e5ba8e55ccef4ded17378be9deccd6f9dd48c42a3d1742fbb1a19cdc7b8c62b2`
- 60 messages, 225 tool calls, 55 provider rounds
- peak process-tree RSS: 3,612,360,704 bytes
- temporary runtime removed successfully

The detailed unified collection protocol is in the PR comment. macOS and Windows maintainers are requested to attach raw JSON from the same commit.


marzich:
## Cross-platform memory test protocol

This comment is the collection protocol for the experiment in this draft PR. Please add every platform result as a new PR comment so the raw evidence and follow-up analysis remain attached to the candidate changes.

Tracking issues: #776 and #777. The PR and these measurements do not close those issues.

### Current status

- The harness is implemented and passes a Linux smoke run.
- The existing Linux numbers are only single-run development samples.
- Repeated Linux results will be posted here after the protocol below is complete.
- macOS and Windows maintainers are requested to run the same protocol and post their results here.
- The PR should remain a draft while platform evidence is incomplete.

### Test invariants

Use the same conditions on every platform:

1. Check out this PR at commit `c43a4626` or a later explicitly noted PR head.
2. Use Bun 1.3.14 unless all platforms jointly move to another pinned version.
3. Run the commands sequentially, never concurrently.
4. Do not run a production Synergy service, browser workload, build, or another memory benchmark in the same process tree.
5. Record OS version, architecture, CPU, total RAM, Bun version, and exact commit.
6. Do not include credentials, prompts, local paths, provider configuration, or session identifiers in the report.

The benchmark does not contact a provider, invoke an external agent, or modify Synergy session state.

### Phase A: bounded microbenchmark

Install the pinned dependencies, then run five independent samples of each preset:

```bash
bun --version
git rev-parse HEAD

bun run perf:memory --preset smoke > memory-smoke-1.json
bun run perf:memory --preset smoke > memory-smoke-2.json
bun run perf:memory --preset smoke > memory-smoke-3.json
bun run perf:memory --preset smoke > memory-smoke-4.json
bun run perf:memory --preset smoke > memory-smoke-5.json

bun run perf:memory --preset standard > memory-standard-1.json
bun run perf:memory --preset standard > memory-standard-2.json
bun run perf:memory --preset standard > memory-standard-3.json
bun run perf:memory --preset standard > memory-standard-4.json
bun run perf:memory --preset standard > memory-standard-5.json
```

`smoke` is a quick validation. `standard` remains bounded: it uses 32 history turns and a 32 MiB tool-stream fixture; it is not a concurrency or stress test.

Each scenario runs in a fresh child process and emits:

- `baseline`: memory after an explicit full GC before the fixture;
- `peak`: largest sampled value while the fixture is live;
- `afterGC`: memory after references are released and full GC runs twice;
- `peakDelta`: allocation above baseline;
- `retainedDelta`: memory still above baseline after release and GC;
- `reclaimedAfterPeak`: memory reclaimed between peak and post-GC measurement.

Memory fields are RSS, heap used/total, external, ArrayBuffers, and runtime footprint. `footprint: null` means the active Bun build does not expose `Bun.unsafe.memoryFootprint()`; it must not be interpreted as zero.

### Phase B: result interpretation

For each platform and scenario, calculate the median and range across the five runs.

- A retained external/ArrayBuffer delta indicates that native-backed data remains accounted after GC.
- External and ArrayBuffers returning to baseline while RSS/footprint remains high indicates allocator-retained pages or delayed OS reclamation.
- A repeated retained delta that grows between independent runs suggests a harness/process isolation defect and must be investigated before product conclusions are drawn.
- Compare runtime footprint across platforms when available.
- Compare RSS deltas within the same platform. Do not compare absolute Darwin RSS directly with Linux cgroup totals or Windows private usage.

Please summarize with this table and include the raw JSON outputs in code blocks or as attachments:

| Preset | Scenario | Median peak RSS delta | Median retained RSS delta | Median retained external | Median retained ArrayBuffers | Median retained footprint | Range / notes |
| --- | --- | ---: | ---: | ---: | ---: | ---: | --- |
| smoke | history-projection | | | | | | |
| smoke | tool-stream | | | | | | |
| standard | history-projection | | | | | | |
| standard | tool-stream | | | | | | |

Report metadata:

```text
Platform / OS version:
Architecture:
CPU:
RAM:
Bun version:
PR commit:
Runtime footprint available: yes/no
Background workload during test: none / describe
```

### Maintainer requests

- **Linux:** I will post five-run smoke and standard results after collection finishes.
- **macOS:** Please test at least Apple Silicon; Intel results are also useful. Post the raw JSON and summary table in this PR.
- **Windows:** Please test Windows x64; Windows on ARM is useful where available. Post the raw JSON and summary table in this PR.

If a command fails on a platform, post the command, Bun/OS metadata, and a redacted stack trace rather than modifying the fixture locally. Platform fixes should preserve the same preset sizes and output schema so the comparison remains valid.

### Follow-up scope

This PR currently covers history projection and decoded tool-input processing. #776 still calls for additional deterministic layers covering local transport, provider adaptation, AI SDK streaming, cancellation, malformed streams, and full Synergy processor settlement. Those layers should be added only after this initial harness is proven portable and the first cross-platform baseline is recorded.

Runtime lifetime changes and isolated Synergy lifecycle verification are intentionally reviewed in #781 rather than this benchmark-only PR.

marzich:
## Protocol amendment: fresh isolated Synergy is required

This amendment applies only to the benchmark experiment in this PR. It supersedes the sentence in the earlier protocol that assigned isolated-runtime verification to #781. PR #781 is a functional-change review with focused automated tests and CI; maintainers are not being asked to run this manual protocol for #781.

Every result reported for this PR must come from a fresh, isolated test environment at the exact tested PR commit. Do not reuse a production Synergy home, server, session, cache, state database, lock, log, or port.

### 0. Create the isolated environment

1. Use a clean checkout/worktree at the exact PR commit and the pinned Bun version.
2. Create a new temporary `SYNERGY_HOME` and an unused loopback port.
3. Copy only the configuration directory needed to expose the same provider/model settings. Do not copy `data`, `state`, `sessions`, `logs`, `cache`, daemon files, or locks.
4. If the runtime check needs authentication, seed exactly one test provider credential into the isolated credential store. Never include that value in commands, reports, logs, or PR comments.
5. Do not stop, restart, signal, or reuse the production Synergy service. It may remain running independently, but it must not share the test home, port, or process tree.

Example setup:

```bash
TEST_HOME="$(mktemp -d -t synergy-memory-test-XXXXXX)"
mkdir -p "$TEST_HOME/.synergy"
cp -R ~/.synergy/config "$TEST_HOME/.synergy/config"

# Pick and verify an unused loopback port before starting the test server.
SYNERGY_HOME="$TEST_HOME" bun dev server --port "$TEST_PORT"
```

Verify `/global/health` on the loopback test port before creating any test session.

### 1. Run the bounded microbenchmark

Run the five smoke and five standard samples from the isolated checkout with `SYNERGY_HOME` set to the new test home. The temporary server must not be running during these process-level samples, so its RSS cannot skew the fixture processes.

The commands, fixture sizes, JSON schema, aggregation rules, and metadata requirements remain exactly as specified in the earlier protocol comment. Run every sample sequentially. Do not run builds, browsers, production workloads, or another benchmark in the same process tree.

### 2. Run one isolated Synergy lifecycle check

After the microbenchmark samples finish, start one temporary Synergy server from the same checkout and test home. Use one fresh empty Scope and one Session only:

1. Record an idle baseline after the server is healthy and resource sampling has started.
2. Send one request whose final response is a short fixed string.
3. Create one unpredictable text fixture no larger than 4 KiB and send one request that must read it through read-only tools before returning a short fixed answer.
4. Do not enable write tools, external agents, background agents, parallel tasks, browser automation, encoding, builds, or stress/concurrency workloads.
5. Verify both turns reach a terminal state, their final messages are visible, no tool remains running, and the Session has no pending reply.
6. Record server RSS, heap used, external memory, and ArrayBuffers at idle baseline, after each turn, and 5, 30, and 120 seconds after the final turn. Use Synergy's resource telemetry as the cross-platform record. Linux may additionally report cgroup/service memory, but it must not substitute that value for process RSS in the cross-platform table.
7. Stop only the temporary server PID, verify its port is closed, securely remove the isolated credential, and remove the temporary home. Leave the production service untouched.

The lifecycle check is a bounded sanity comparison, not a provider benchmark and not a claim that two requests reproduce production memory pressure. Report it separately from the microbenchmark table.

### 3. Report the isolation evidence

Alongside the existing five-run result table, report:

```text
Fresh temporary home: yes
Production home/state reused: no
Dedicated loopback port: yes
Temporary server health check: pass/fail
Fixed response turn: pass/fail
Read-only tool turn: pass/fail
Terminal/pending state verification: pass/fail
Idle / post-turn / +5s / +30s / +120s memory summary:
Temporary process and credential cleanup: pass/fail
```

Do not publish the temporary path, port, Scope/Session/message identifiers, provider credentials, prompts containing private data, or machine-internal configuration. macOS and Windows maintainers should follow the same isolation and workload rules so their results can be compared with Linux without contamination from historical Synergy state.

marzich:
## Superseding cross-platform collection protocol

This automated protocol supersedes the earlier conceptual/manual runtime amendment. Use commit `ce41c602` so every platform runs the same fixture and harness.

### What is now standardized

`perf:memory:runtime` creates its own temporary Synergy home, workspace, loopback server, deterministic provider, and deterministic payload tool. It replays one anonymized completed Synergy trajectory with 1 root Session, 4 parallel background subagents, 97 source messages, and 220 tool calls. It never imports the operator's config, credentials, Sessions, caches, or logs, and it removes the temporary runtime after the result is written.

There is no manual Session setup and no real provider credential. Do not point the harness at an existing Synergy server.

### Exact commands

From the repository root at commit `ce41c602`:

```bash
bun install --frozen-lockfile

# Gate only; these results are not the cross-platform comparison.
bun run perf:memory --preset smoke > memory-process-smoke.json
bun run perf:memory:runtime --preset smoke > memory-runtime-smoke.json

# Comparable measurements.
bun run perf:memory --preset standard > memory-process-standard.json
bun run perf:memory:runtime --preset standard > memory-runtime-standard.json
```

Run each command once, sequentially, on an otherwise idle machine. Do not run them in parallel. The runtime standard command intentionally waits for the 5, 30, and 120 second post-deletion samples.

### Required result checks

`memory-runtime-standard.json` must report:

```text
preset = standard
cleanup.temporaryRuntimeRemoved = true
checks.verifiedTrajectoryMessages = 97
checks.verifiedChildSessions = 4
checks.verifiedToolCalls = 220
checks.verifiedCompletedTools = 219
checks.verifiedErrorTools = 1
checks.pendingReplyCleared = true
checks.noRunningTools = true
```

The current dev runtime also persists one empty assistant step-limit boundary. It is expected as `checks.verifiedRuntimeBoundaryMessages = 1` and is excluded from the 97 source trajectory messages.

### What to post in this thread

Please attach both `memory-process-standard.json` and `memory-runtime-standard.json`, then include:

```text
OS and version:
Architecture:
Bun version:
CPU:
Physical memory:

Runtime baseline: RSS / heap / external / ArrayBuffers
Runtime peak: RSS / heap / external / ArrayBuffers
Release +5s: RSS / heap / external / ArrayBuffers
Release +30s: RSS / heap / external / ArrayBuffers
Release +120s: RSS / heap / external / ArrayBuffers
```

Use process RSS, heap, external, and ArrayBuffers for cross-platform comparisons. `serviceMemory` is supplemental: on Linux an ad hoc benchmark process may share a cgroup with unrelated processes, while macOS and Windows use different availability/fallback semantics.

### Linux reference result

Linux x64, Bun 1.3.14:

| Sample | RSS | Heap used | External | ArrayBuffers |
| --- | ---: | ---: | ---: | ---: |
| Baseline | 419.03 MiB | 167.12 MiB | 51.21 MiB | 2.51 MiB |
| Peak | 819.03 MiB | 362.53 MiB | 110.31 MiB | 4.60 MiB |
| Release +5s | 751.13 MiB | 325.74 MiB | 108.55 MiB | 4.15 MiB |
| Release +30s | 716.28 MiB | 293.41 MiB | 82.31 MiB | 2.41 MiB |
| Release +120s | 719.32 MiB | 294.18 MiB | 82.43 MiB | 2.41 MiB |

At +120s the replay had zero Session runtimes and zero message-cache bytes. ArrayBuffers were close to baseline, while RSS remained 300.30 MiB above baseline. macOS and Windows results are needed to determine how much of that residual is platform allocator behavior versus cross-platform retained heap.

marzich:
## Linux standard trajectory run

The runtime fixture completed successfully on Linux/Bun 1.3.14 in two fresh temporary runtimes. I compared benchmark-only current `dev` with an integration worktree containing candidate lifecycle/observability changes from #819, #822, #823, #825, #826, and #827. Production Synergy was not used or restarted.

At release +120s:

- dev: RSS 717.7 MiB, heap 295.3 MiB, external 79.9 MiB, ArrayBuffers 1.91 MiB
- candidate stack: RSS 698.6 MiB, heap 222.0 MiB, external 51.2 MiB, ArrayBuffers 1.29 MiB

Relative to idle, retained RSS was 288.2 MiB on dev and 262.3 MiB with the candidate stack. Retained heap was 124.3 MiB versus 50.8 MiB. Retained external was 28.2 MiB versus -0.4 MiB.

The harness validated health, full HTTP Session flow, provider streaming, four background subagents, Cortex notifications, task output retrieval, terminal state, pendingReply cleanup, no running tools, and temporary-runtime removal.

This paired result demonstrates that the harness can expose lifecycle improvements separately from allocator RSS retention. It does not establish statistical significance. Maintainers running macOS or Windows should use the exact standard command below from a clean temporary checkout and post the complete JSON plus platform, architecture, and Bun version:

```bash
bun install
bun run perf:memory:runtime --preset standard > session-memory-runtime-standard.json
```

Do not point the benchmark at a production Synergy home; the harness already creates and removes its own isolated runtime.

marzich:
## Unified benchmark protocol

Please run this in a disposable checkout or worktree of this PR. The harness itself starts a brand-new temporary Synergy server/home/workspace for each scenario and removes it afterward; do not point it at a production Synergy instance or copy user configuration into it.

### 1. Environment record

Record:

```bash
git rev-parse HEAD
git status --short
bun --version
uname -a  # Windows: include OS/build and architecture instead
```

Use the same Bun version when comparing source revisions. A clean worktree is preferred because every result records `revision.dirty`.

### 2. Harness validation

```bash
bun install
bun run perf:tools:test
bun run perf:memory:runtime:matrix --preset smoke > memory-smoke-matrix.json
```

The JSON must contain `trajectory`, `parallel`, and `sequential`, with `schemaVersion: 4` per scenario and `cleanup.temporaryRuntimeRemoved: true`.

### 3. Standard real-lifecycle workloads

Run each scenario separately so failures and raw artifacts remain attributable:

```bash
bun run perf:memory:runtime --preset standard --scenario trajectory > memory-trajectory-standard.json
bun run perf:memory:runtime --preset standard --scenario parallel > memory-parallel-standard.json
bun run perf:memory:runtime --preset standard --scenario sequential > memory-sequential-standard.json
```

These are fixed workloads, not arbitrary load generation:

- `trajectory`: one primary plus four parallel background subagents, native `task`/`task_output`/Cortex completion lifecycle.
- `parallel`: five concurrent primaries and five Agent workers, 225 aggregate tool calls.
- `sequential`: five heavy primary lifecycles with an after-delete checkpoint for every replica.

The standard release window intentionally lasts 120 seconds. Do not stop a run after the workload becomes idle; the release samples are part of the result.

### 4. Repetition and version comparison

For a publishable comparison, run each standard scenario three times per source revision on an otherwise idle machine and compare medians. Keep every raw JSON file.

Only compare two files when both of these match exactly:

```text
workload.contractVersion
workload.fingerprint
```

Also keep platform, architecture, Bun version, preset, scenario, replica count, and worker settings identical. The source revision is expected to differ between a baseline and candidate; it is not part of the workload fingerprint.

Compare the stable fields directly from the raw JSON:

- phase peak server RSS, process-tree RSS, descendant RSS, heap, external, and ArrayBuffers;
- sequential `*-after-delete` checkpoints and their slope;
- `release-plus-5s`, `release-plus-30s`, and `release-plus-120s` absolute values;
- the corresponding idle-relative deltas;
- runtime/session/message-cache counts at terminal and release.

This PR intentionally does not add a comparison utility. The raw schema and workload identity are the contract, so downstream analysis can use the same artifacts without coupling benchmark validity to one reporting tool.

### 5. Platform requests

- **macOS maintainers:** please confirm the `ps` process-tree collector reports the server plus descendants and attach all raw standard JSON files.
- **Windows maintainers:** please confirm the PowerShell/CIM process-tree collector works without elevation and attach all raw standard JSON files.
- **Linux maintainers:** attach process-tree metrics and any available service-memory/cgroup fields, but keep those platform-specific values separate from cross-platform process metrics.

Do not include credentials, production logs, user Sessions, or private paths beyond the automatically generated temporary path. The fixture contains no source prompts or outputs.

marzich:
## Linux local result update

This comment records the local Linux runs performed while completing the benchmark. It is evidence for harness behavior, not a claim that the underlying memory issue is fixed.

Environment: Linux x64, Bun 1.3.14.

### Final contract result

The final `parallel` run was executed from a clean worktree at commit `7a9ff6442d132f9f88c01a07c1c956d9a5be18e0` with runtime result schema v4.

- workload fingerprint: `e5ba8e55ccef4ded17378be9deccd6f9dd48c42a3d1742fbb1a19cdc7b8c62b2`
- 5 concurrent primary Sessions / 5 Agent workers
- 60 persisted messages / 225 tool calls / 55 provider rounds
- all 5 replicas verified 12 messages and 45 completed tool calls
- clean terminal state and temporary runtime removal passed

| Sample | Server RSS | Process-tree RSS | External | ArrayBuffers |
| --- | ---: | ---: | ---: | ---: |
| Idle | 454,672,384 | 1,127,677,952 | 54,763,032 | 2,629,994 |
| Peak | 854,958,080 | 3,612,360,704 | 100,838,389 | 3,050,656 |
| Release +5s | 665,235,456 | 3,208,847,360 | 60,764,798 | 1,021,859 |
| Release +30s | 709,898,240 | 3,219,968,000 | 69,513,760 | 1,547,499 |
| Release +120s | 755,421,184 | 3,281,707,008 | 82,359,122 | 2,138,331 |

The result shows why both server and full process-tree metrics are needed: the server retained about 301 MB above idle at +120s, while descendants still accounted for about 2.51 GB. Summed process-tree RSS can double-count shared pages, so this is attribution evidence rather than physical working set.

### Supporting pre-contract runs

The full `trajectory` and `sequential` standard scenarios were run immediately before the schema-v4 workload contract/fingerprint metadata was finalized. Their runtime workload implementation is the same, but these files are schema v3 and have no final fingerprint. They demonstrate scenario completion and release sampling; they must not be presented as final same-fingerprint cross-version results.

| Scenario | Workload | Idle server RSS | Peak server RSS | +120s server RSS | Peak process-tree RSS | Result |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| `trajectory` | 1 primary + 4 subagents; 97 messages / 220 tools | 448,606,208 | 837,009,408 | 728,379,392 | 3,143,585,792 | passed |
| `sequential` | 5 heavy primary lifecycles; 225 tools | 450,527,232 | 825,298,944 | 661,827,584 | 2,776,969,216 | passed |

Sequential after-delete server RSS checkpoints were:

```text
run-1  656,986,112
run-2  696,315,904
run-3  741,122,048
run-4  745,283,584
run-5  736,915,456
```

This supporting run showed an upward server-RSS slope across the first four completed/deleted lifecycles while descendant RSS stayed comparatively stable. Repeated final schema-v4 runs on each platform are still required before drawing a regression or platform conclusion.

Raw local files were retained outside the repository under `/tmp`; they were not committed because benchmark output is machine evidence rather than source. Future platform submissions should attach or otherwise preserve the raw JSON generated by the unified protocol above.

yzxoi:
The updated benchmark is now available in #838. It retains the cross-platform trajectory workload while fingerprinting the complete elastic Agent worker settings from current dev, so comparisons describe the runtime actually under test.

PR #781, opened 2026-07-22
fix(session): release detached background and turn payloads

marzich:
## Summary

- Detach non-blocking `LoopJob` execution from the full live session context.
- Coalesce duplicate background work per job/session key with one active run and one latest pending payload.
- Add bounded weak-reference ownership probes for history, prompt, provider stream, processor, and tool-call payloads.
- Clear turn-local history and tool-definition containers after the processor has settled.

Related to #774 and #775, with Linux validation tracked in #777. This PR does not automatically close those issues.

## Changes

| Area | Change |
| --- | --- |
| `LoopJob` contract | Non-blocking jobs capture a minimal serializable payload instead of retaining `LoopJob.Context`. |
| Background lifecycle | Duplicate work is latest-wins coalesced; active count, duration, payload size, and coalescing are observable. |
| History users | Title, compaction, summary, and chronicler jobs reload the required history only while executing. |
| Turn ownership | WeakRef/FinalizationRegistry probes report target bytes, collection latency, and 30s/120s liveness by owner. |
| Terminal cleanup | Turn-local arrays/maps are cleared only after normal processor settlement; timed-out processors keep their live inputs intact. |

## Safety

- No prompt compression, context truncation, or concurrency reduction.
- No platform-specific GC or allocator policy change.
- The retention probe is bounded to 256 groups and 32 targets per group.
- The probe does not hold strong references and can be disabled with `SYNERGY_RETENTION_PROBE_ENABLED=0`.
- Failed background jobs remain isolated from the foreground turn.

## Verification

- `bun run format:check`
- `bun run lint`
- `bun run typecheck` (11/11 packages)
- `bun run monorepo:check`
- 77 focused session tests passed across LoopJob, summary, processor, memory pressure, LLM memory, and retention-probe suites.
- An isolated source-server workflow completed message persistence, read-only tool calls, background title work, and session cleanup without a session/tool error.

The cross-platform memory benchmark and maintainer data-collection protocol are intentionally separated into draft PR #780.


yzxoi:
## Review findings

Thanks for working on this — the `capture(ctx) -> execute(payload)` split is the right direction and substantially reduces accidental retention of the live loop context. I found one merge-blocking correctness regression plus several lifecycle/observability issues that should be fixed before merge:

1. **P1: summarize loses intermediate terminal turns.** `LoopJob` defaults every background job to one active + one latest pending payload per `job.type/sessionID`. The summarize job has no custom key, so rapid revisions A1/A2/A3 execute as A1/A3. This bypasses `SessionSummary`'s existing FIFO and permanently skips A2's message-level summary. Please preserve per-revision identity or let `SessionSummary` own the queue.
2. **P2: background jobs have no cooperative timeout contract.** A hung execution keeps the scheduler entry forever and prevents all later work for the same key from running.
3. **P2: detached history reloads use `Session.messages()`.** Title, prune, summarize, and chronicler can rehydrate the full transcript instead of an independent compact-aware model working set, increasing peak memory/I/O on long compacted sessions.
4. **P2: `RetentionProbe` collection metrics are lossy.** Active sweeps delete dead targets/groups before the `FinalizationRegistry` callback, so `llm.turn.retention.collected` is usually skipped; pruned groups also leave registrations until target collection.
5. **P2: the default-enabled probe mutates observed objects.** The non-enumerable marker still changes `Reflect.ownKeys()`/object shape. A `WeakMap` marker avoids changing provider/tool payloads.
6. **P2: tool input is serialized twice** in the tool-call hot path; reuse the already computed `toolInputBytes` for the size guard.

The focused suites and current CI are green, but the new generic coalescing test explicitly expects `[1, 3]` and there is no summarize integration test covering three rapid terminal turns, so the P1 regression is not detected.

I am preparing a follow-up patch on this PR branch with regression coverage and the focused fixes above.

yzxoi (approved):

PR #782, opened 2026-07-22
fix(app): hide note editor tags

EricSanchezok:
## Summary

- remove the tag chips and tag input from the Note editor
- preserve existing tag data, synchronization, and backend behavior
- remove frontend state and handlers used only by the hidden controls

## Verification

- independent maintainability review — passed
- Note sync test could not run because `@happy-dom/global-registrator` is unavailable in this worktree
- App typecheck could not run because `tsgo` is unavailable in this worktree


yzxoi:
实际目标为：`note-bubble-menu` 的定位逻辑错误导致 notes 页面下工具浮标和 tags 重叠，将会在新的 PR 修复。

PR #792, opened 2026-07-23
fix: exit terminal light loops

EricSanchezok:
## Summary

- clear ordinary Light Loop workflow bindings when approval or another terminal outcome completes the loop
- retain plugin-owned terminal records for reliable `lightloop.after` delivery while treating those records as inactive everywhere else
- migrate existing terminal ordinary Light Loop sessions and prevent stale composer mode state
- leave Lattice behavior unchanged

## Testing

- `bun run quality`
- `bun run deadcode`
- `SYNERGY_LINK_HOME=/tmp/synergy-link-lightloop-ci bun run test:ci` from `packages/synergy`
- focused Light Loop runtime, continuation, review-tool, migration, working-state, wrapper, and App control tests


yzxoi:
## Design issue: terminal plugin Light Loops remain equipped

This change correctly introduces an active-vs-terminal distinction for retained plugin-owned Light Loop records, but that distinction is only applied to some consumers.

After a plugin-owned Light Loop reaches a terminal status, `session.workflow` is retained so `lightloop.after` can be retried. At that point the UI, continuation policy, working state, and `loop_stop` resolver treat the workflow as inactive, while other paths still treat any `workflow.kind === "lightloop"` as equipped:

- `SessionInvoke` continues injecting the Light Loop system prompt.
- Agenda wake-up still chooses the Light Loop agent for a terminal record.
- `SessionWorkflowService` still rejects enabling Plan, Lattice, or a new Light Loop because the workflow slot is non-empty.
- The prompt submit path skips enabling a newly armed Light Loop because it sees the retained terminal `kind`.

This creates a split state: the composer shows that Light Loop has exited, but the execution context and workflow slot can still behave as if it is active. In particular, the approval mail wakes the execution session after terminalization, so the next model call can receive instructions to continue the Light Loop while `loop_stop` has already been hidden.

The intended semantic should be terminal = automatically unequipped. Plugin lifecycle delivery must remain durable, but it should not keep occupying the interactive workflow slot. I’ll follow up with a fix that separates retained terminal hook-delivery state from the active workflow binding and adds regression coverage for post-terminal invocation and re-equipping workflows.

yzxoi (approved):

PR #796, opened 2026-07-23
feat: unify conversation attachment previews

EricSanchezok:
## Summary

Unify sent conversation attachment opening across Web and Desktop while preserving the image-specific preview experience:

- image attachments continue to open in Image Preview
- located non-image attachments open as resource tabs in the right Side Workspace
- Image Preview and Attachment Workspace can reveal an eligible source file through the existing Files workbench
- every preview continues to read the immutable conversation asset; opening the source is an explicit secondary action

## Opening contract

| Situation | Primary click result | Secondary actions / fallback |
| --- | --- | --- |
| Sent image (`image/*`) | Open the existing Image Preview modal, including the message-local gallery | Show **View source in Files** only when the recorded source path normalizes inside the active Scope; retain download/open-image actions |
| Sent image whose inline thumbnail failed | Fall back to the file-shaped card; clicking still opens Image Preview, which owns its existing load-error state | Download remains available |
| Sent located non-image (`sessionID + messageID + attachmentID`) | Open a lazy Attachment resource tab in the right Side Workspace | Reopening the same locator focuses the same tab; another locator opens a distinct tab |
| Draft composer attachment | Keep the existing draft compatibility behavior | No durable Attachment tab is created because there is no message locator |
| Legacy non-image without a complete locator | Keep the existing path/URL compatibility behavior | Prefer a scoped Files path when present, otherwise use the resolved asset/remote URL |
| Hidden attachment presentation | Render no conversation card | No click target is introduced |
| Missing historical attachment | Keep the resource tab and show an unavailable state | The locator remains durable; bytes are not persisted in layout state |

## Format behavior

| Format | Attachment Workspace behavior |
| --- | --- |
| PDF | Lazy-load PDF.js; render pages with previous/next, fit-width, and zoom controls; preview up to 50 MiB |
| Markdown | Preview with the shared Markdown renderer or switch to read-only Source; preview up to 4 MiB |
| HTML | Switch between read-only Source and sanitized Preview; remove scripts, event handlers, forms, frames, objects, and external media sources; render through the shared no-script/no-network/no-form sandbox; preview up to 4 MiB |
| Text, JSON, CSV, XML, YAML, and common code files | Open read-only Source through the existing source viewer; preview up to 4 MiB |
| Audio / video | Keep native inline playback controls in the conversation and add an explicit open action; the Attachment tab uses native media controls |
| Office files, archives, and unknown binary formats | Show filename, MIME/type metadata, and Download without attempting an unsafe or misleading inline conversion |
| Oversized or failed/CORS-blocked preview | Show a clear unable-to-preview or size-limit state while keeping Download available |

## Source-file policy

The source action uses this explicit precedence:

1. `localPath`
2. `source.type === "file"` and `source.path`
3. `metadata.attachment.sourcePath`

The action is exposed only after the path normalizes inside the current server Scope. Image Preview closes after handing off to Files. Attachment Workspace stays open as its own resource tab while Files becomes active. Missing/deleted sources are delegated to the existing Files workbench recovery state.

The `attach` tool and OpenAI image tools now record the normalized output/source path. The attachment asset URL remains the canonical preview and download target, so a later source edit cannot silently change conversation history.

## Architecture and state

- one shared resource-open policy is used by the Web and Desktop renderer
- Attachment tabs persist only a versioned `{ sessionID, messageID, attachmentID }` locator
- attachment resolution checks the current frontend data store first, including completed tool-result attachments, then uses the generated `session.message` SDK method when history is not resident
- the programmatic Attachment panel is excluded from the user launcher
- PDF.js, the Attachment panel, and its CSS stay behind the workbench lazy boundary; the production manifest test verifies the dynamic entry and separate PDF worker
- no server route, OpenAPI schema, database migration, or generated SDK contract changes are required

## Implemented plan

- [x] centralize the image / located non-image / compatibility routing decision
- [x] add stable Attachment resource identity and same-resource tab reuse
- [x] implement per-format preview capability and explicit byte limits
- [x] preserve native media controls while adding a non-conflicting open affordance
- [x] add source-file handoff to both Image Preview and Attachment Workspace
- [x] sanitize HTML and retain the existing sandbox boundary
- [x] resolve evicted top-level and nested tool attachments through existing data/SDK paths
- [x] keep Attachment and PDF resources out of the initial bundle
- [x] add English, Simplified Chinese, and pseudo-locale strings
- [x] document the durable product behavior in `packages/app/PRODUCT.md`

## Verification

- `bun test --cwd packages/app test/app-build-css-contract.test.ts`
- `bun run --cwd packages/desktop desktop:build`
- `bun turbo test --filter='!synergy'`
- `bun run release:test`
- `bun run format:check`
- `bun run lint`
- `bun run test-layout:check`
- `bun run typecheck`
- `bun run monorepo:check`
- `bun run package:check`
- `bun run deadcode`

All applicable local gates pass. The local full-history secret scan reports an existing JWT-shaped test fixture from earlier `dev` commits; this change introduces no secret finding, and the PR's shallow-checkout Secret Scan remains the publication gate.


yzxoi:
I found five issues that should be fixed before merge:

1. **PDF renders can complete out of order.** Page/zoom/width changes start overlapping `getPage()` calls without invalidating the older request, so a stale request can cancel the newer render and paint the wrong page or scale. A rejected `getPage()` is also currently unhandled.
2. **Attachment preview reads outlive the active tab.** The payload fetch does not receive an `AbortSignal`; closing or switching the keyed workbench tab can leave a read of up to 50 MiB running, and repeated opens can stack work and retained buffers.
3. **The published `ResourceOpenController` contract became source- and runtime-incompatible.** The two new workspace methods are required, and call sites only optional-chain the controller rather than the methods. Existing external providers therefore fail type checking or throw at runtime under mixed versions.
4. **The attachment toolbar clips actions in a narrow side panel.** Its responsive rule follows viewport width, while the workspace panel can independently shrink to 300 px on a wide desktop viewport. Because the surface hides overflow, Source/View source/Download can become unreachable.
5. **Filename-less tabs persist an English `Attachment` title.** This bypasses the localized workbench fallback, so a Chinese UI can show an English tab title.

I’ll add regression coverage for these cases, apply the fixes, and push the follow-up commit to this PR branch.

yzxoi (approved):

PR #801, opened 2026-07-23
fix(cortex): resume workflows after silent task completion

EricSanchezok:
## Summary

- resume parent-session continuation arbitration when a Cortex task completes without a parent notification
- preserve silent-task behavior: no Cortex completion Inbox item is created
- add an end-to-end regression covering an active LightLoop blocked by hidden Cortex work

## Bug

The continuation gate correctly defers workflow proposals while a Cortex child is queued or running. When the parent loop released during that wait, its one drive request observed the active child and stopped without creating a continuation.

For ordinary Cortex tasks, terminal completion persisted an Inbox notification and requested `SessionDrive` for the parent. Silent tasks with `notifyParentOnComplete: false` intentionally skipped that notification path, but the parent drive request lived inside the same path. As a result, the final wait disappeared without another continuation arbitration. An active LightLoop could therefore remain idle indefinitely even though it had no stop request and no remaining continuation waits.

## Fix

After publishing terminal Cortex state, the silent-completion branch now requests `SessionDrive` for the parent session. This keeps the two contracts distinct:

- `notifyParentOnComplete` controls whether the parent receives a Cortex result notification.
- every terminal transition that clears a continuation wait gives the parent session another opportunity to arbitrate pending workflow continuation.

If the parent is still running, the drive safely declines and the normal release drive remains authoritative. If the parent is idle, the continuation kernel immediately reevaluates the now-cleared wait set. Drive failures are logged without changing the already-published task result.

## Regression coverage

The new behavioral test creates a terminal parent session with an active LightLoop, launches a hidden Cortex task with parent notification disabled, and verifies that:

1. continuation is blocked while the child is active;
2. child completion automatically produces the system `Continue light loop` steer;
3. the parent session is woken;
4. no Cortex completion notification is added.

## Verification

- `bun test test/session/continuation-kernel.test.ts`
- `bun test test/cortex/manager.test.ts`
- `bun test test/session/continuation-kernel.test.ts test/session/light-loop-continuation.test.ts test/session/manager.test.ts`
- `bun run test-layout:check`
- `bun run typecheck`
- `bun run monorepo:check`
- `bun run package:check`
- push hook: format, lint, typecheck, and monorepo checks


yzxoi:
I reviewed this against the latest `dev` architecture. The live-path fix is necessary and the placement is directionally correct: a silent Cortex terminal transition can release the final `ContinuationWait`, and `SessionDrive` is the right owner for re-running continuation arbitration without publishing a Cortex completion message.

There is one durability gap I want to fix before merging. The new drive only runs inside the live `updateTaskStatus()` path. On restart, `reconcileInterruptedCortexDelegations()` can turn the last queued/running silent child into `interrupted`, which releases the wait, but `reconcileParentNotifications()` skips hidden/`notifyParentOnComplete: false` delegations and the ordinary workflow recovery path does not drive active LightLoop/BlueprintLoop sessions without a stop request. The parent can therefore remain permanently idle after restart—the same user-visible failure as the original bug.

I’m going to add startup recovery coverage for this case and update the recovery path so releasing a silent Cortex wait durably re-drives the parent through `SessionDrive`. I’ll also keep the live-path regression and add adjacent coverage where useful.

yzxoi (approved):

PR #810, opened 2026-07-24
fix(runtime): bound hot-path observability logs

### LINKED ISSUE #809: perf(runtime): hot-path tool and permission telemetry can saturate the server event loop
## Summary

A single tool-heavy workflow can drive the Synergy main Bun thread to a sustained full core even when the service is not OOMing. The dominant local amplification is INFO-level telemetry emitted from two hot paths:

- every `PermissionNext.evaluate()` call;
- every tool initialization in `ToolRegistry.tools()`.

Each INFO record is not only formatted and written to the process log; it is also mirrored into observability/event replay. A normal tool turn therefore multiplies into hundreds of log/SQLite/event operations.

## Production evidence (Linux)

During one active workflow:

- main Bun thread: approximately 100% of one CPU core for hours;
- 11,539 `log.record` events in a 15-minute window;
- `tool.registry`: 4,462 INFO records;
- `permission`: 4,094 INFO records;
- event-loop lag peaked around 3-4.6 seconds;
- backend requests that are normally milliseconds reached 1-2.7 seconds;
- cgroup PSI was zero at the sampled point and no OOM occurred, so memory pressure did not explain the persistent spin.

A local minimal change that demoted permission evaluation telemetry to DEBUG and removed per-tool initialization INFO timers reduced the restarted idle service to roughly 1-6% CPU and restored root HTTP responses to roughly 6-9 ms. This comparison is not a formal same-load benchmark, but it confirms that the hot-path telemetry has material runtime cost.

## Relevant code

- `packages/synergy/src/permission/next.ts`
  - `log.info("evaluate", ...)` runs for every permission check.
- `packages/synergy/src/tool/registry.ts`
  - `using _ = log.time(t.id)` emits an INFO timing record for each initialized tool.

The resolver already has aggregate diagnostics and explicit warning/error paths, so per-tool success timings do not need to be emitted at INFO level.

## Proposed minimal fix

1. Change per-check permission evaluation logging from INFO to DEBUG.
2. Remove the per-tool initialization INFO timer while retaining initialization failures and aggregate resolver diagnostics.
3. Keep permission behavior, tool availability, schemas, concurrency, and security decisions unchanged.
4. Add a focused test that prevents per-check permission records from returning at INFO level.

## Expected result

Tool-heavy workflows should no longer generate thousands of low-value INFO records per few minutes, reducing event-loop work, observability writes, replay traffic, log growth, and HTTP latency without reducing functionality.

marzich:
## Summary

- demote per-permission-evaluation telemetry from INFO to DEBUG
- remove per-tool successful initialization timers from the hot resolver path
- keep tool initialization failures, aggregate diagnostics, permission behavior, and tool behavior unchanged
- add a regression test that prevents per-check permission events at INFO level

## Why

In a Linux production incident, one tool-heavy workflow generated 11,539 log records in 15 minutes. `tool.registry` accounted for 4,462 records and `permission` for 4,094. The main Bun thread stayed at approximately one full CPU core, event-loop lag reached 3-4.6 seconds, and normally fast HTTP requests reached 1-2.7 seconds.

These success-path INFO records are also mirrored into observability/event replay, amplifying a single tool turn into many formatting, SQLite, and event operations. The changed records remain available at DEBUG where applicable; warnings and failures remain visible.

Fixes #809.

## Validation

- `bun test test/permission --timeout 30000` (229 passed)
- `bun test test/permission/telemetry.test.ts test/tool/schema.test.ts --timeout 30000` (2 passed)
- `bunx prettier --check packages/synergy/src/permission/next.ts packages/synergy/src/tool/registry.ts packages/synergy/test/permission/telemetry.test.ts`
- repository pre-push formatting, lint, workspace typecheck, and dependency checks passed


yzxoi:
I reviewed this against the current Synergy execution-boundary and observability architecture and reproduced the relevant log behavior.

The two production-code changes are directionally correct and necessary:

- `PermissionNext.evaluate()` is auxiliary permission/tool-visibility filtering, not the centralized `EnforcementGate` authorization path, so moving its success-path telemetry below INFO does not weaken execution enforcement.
- Removing the per-tool `log.time(t.id)` is appropriate. It emits one INFO/SQLite event per initialized tool, while aggregate `toolResolver.resolve` timing and per-tool initialization warnings remain available.

I found three follow-ups before merging:

1. **Restore the DEBUG telemetry contract test.** The current test now verifies only that INFO emits no event, but deletes the existing assertions that DEBUG telemetry is bounded and does not contain raw `pattern`, `ruleset`, or sensitive values. Since DEBUG remains a supported level and still persists one `log.record` per evaluation, we should keep both contracts: no per-check event at INFO, and bounded/redacted data at DEBUG.
2. **Add coverage for the `tool.registry` half of the regression.** This was the largest event source in the incident, but there is currently no test preventing per-tool successful initialization INFO records from returning.
3. **Update the performance-observability documentation** to state that permission evaluation emits no record at INFO and emits bounded diagnostic records at DEBUG.

I also verified that the patch applies cleanly over the current `dev`, the focused permission/tool suites and typecheck pass, and INFO produces zero matching permission events while DEBUG still produces one event per evaluation.

I am applying these test/documentation fixes to this PR branch now. The broader lack of sampling/rate limiting for high-frequency `log.record` events should remain a separate observability follow-up rather than expanding this incident fix.

yzxoi (approved):

Issue #813, opened 2026-07-24
Active-session memory amplification in the isolated runtime compared with v2.1.3


marzich:
## Summary

Synergy v2.1.3 did not show the same degree of memory growth under active session load observed in the current runtime. A tree-level comparison and Linux process attribution indicate that the newer execution-isolation architecture adds both a substantially higher per-process cost and a full-turn IPC copy path.

This issue documents the architectural amplification mechanism. It does **not** claim that every long-running memory increase is explained, nor that v2.1.3 had perfect lifecycle cleanup.

## Version comparison

- v2.1.3: `ccb39fd1608f589d8eee29eae46eb2bcd0681f7e` (2026-06-15)
- Current dev examined: `1e6187b0ee4eee8d0bccbaaaaae136c4139ea401` (2026-07-24)

The histories are not linearly comparable through a merge base, so the findings below come from source-tree comparison and live process attribution rather than commit ancestry.

### 1. v2.1.3 executed LLM turns in the main process

The current runtime routes Agent turns through an isolated `AgentWorkerPool`. On the first real turn, `ensureWorkers()` fills the configured pool to its fixed size.

Current defaults:

- workers: `min(4, available CPUs - 1)`, at least 1
- queued turns: 256
- aggregate queued payload limit: 256 MiB
- recycle after 64 turns
- recycle at 1536 MiB RSS or 1024 MiB heap per worker

There is currently no idle-worker timeout, minimum/maximum split, or automatic pool shrink. Recycling replaces a worker rather than reducing the pool size.

v2.1.3 had no AgentTurn worker pool and called the LLM stream in-process.

### 2. Live Linux PSS shows that isolated workers materially raise active-runtime memory

A production sample after the worker pool had been activated showed:

| Process | PSS |
|---|---:|
| Main Synergy process | ~668 MiB |
| Agent worker 1 | ~310 MiB |
| Agent worker 2 | ~283 MiB |
| Agent worker 3 | ~337 MiB |
| Agent worker 4 | ~218 MiB |

The four Agent workers accounted for about **1.15 GiB PSS**. Service cgroup memory was approximately **2.17 GiB** at the same time.

This confirms that the higher footprint is not only log page cache or a misleading main-process RSS reading. Multiple independent Bun/JSC runtimes account for most of the working set.

### 3. Every active turn introduces a full-request serialization and copy path

The current AgentTurn request includes the prepared system prompt, model working messages, provider plan, and tool definitions. It is then processed through this path:

[...]

PR #819, opened 2026-07-24
feat(observability): restore Linux runtime memory attribution

marzich:
## Summary
- report Linux service-cgroup working set, reclaimable memory, pressure, and OOM counters
- attribute Bun/JSC heap and allocator state without changing non-Linux behavior
- expose the new measurements through the existing Performance catalog

## Safety
- observability-only; no GC, scheduling, concurrency, prompt, or output policy changes
- non-Linux platforms retain the existing process-tree fallback
- excludes the local-only sandbox allowlist commit

## Verification
- `bun test test/process/service-memory.test.ts test/observability/linux-runtime-memory.test.ts` (6 passed)
- `bun run format:check`
- `bun run lint`
- pre-push monorepo typecheck

Related to #777 and #813. This PR does not close either investigation.

### DIFF
--- a/packages/synergy/src/observability/linux-runtime-memory.ts
+++ b/packages/synergy/src/observability/linux-runtime-memory.ts
@@ -1,0 +1,114 @@
+import { heapStats } from "bun:jsc"
+
+export namespace LinuxRuntimeMemory {
+  const DEFAULT_INTERVAL_MS = 60_000
+  const TOP_TYPE_LIMIT = 12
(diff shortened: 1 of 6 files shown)


yzxoi:
The consolidated implementation is now available in #838. It keeps Linux working-set, memory.events, PSI, JavaScriptCore growth, and allocator diagnostics, integrated through the existing ServiceMemory and observability boundaries.

Issue #816, opened 2026-07-24
[P1][area/tools] Release local process handles after stdio close without losing tail output


marzich:
## Summary

Local Bash tracks process exit as the terminal event even though stdout/stderr may close later. Process/listener/stdin references are not deterministically cleared at the stdio close boundary. This can delay reclamation and risks conflating exit with complete output delivery.

## Proposed change

- Treat exit as status and close as the resource-release boundary.
- Preserve output arriving between exit and close.
- After close, detach listeners and clear child, stdin, timer, abort, and temporary execution references.
- Preserve foreground/background, poll/log/kill, timeout, cancellation, output, and attachment behavior.

## Non-goals

- No new output truncation or storage format.
- No concurrency or process admission changes.

## Acceptance criteria

- Tail output is not lost.
- Active cancellation and process-tree termination still work.
- Closed processes no longer retain live child/stream handles.

Parent investigation: #813

PR #823, opened 2026-07-24
fix(process): release local handles after stdio close

marzich:
## Summary

- finalize local Bash processes on `close`, after stdout and stderr have drained
- remove child stream listeners and release retained child/stdin handles after completion
- remove session-shell stream and abort listeners when its process closes

## Why

`exit` can fire before inherited stdio closes. Keeping completion on `exit` risks either losing tail output or retaining child-process wrappers and listeners after the logical tool lifecycle. This preserves output correctness while releasing those strong references at the actual stdio boundary.

Related to #816 and #813. This PR does not close either issue.

## Verification

- `bun test test/tool/bash.test.ts test/process/registry.test.ts` (53 passed)
- repository Prettier, oxlint, Turbo typecheck, and sherif pre-push gates

No service restart or production runtime change was performed.

### DIFF
--- a/packages/synergy/src/session/shell.ts
+++ b/packages/synergy/src/session/shell.ts
@@ -186,7 +186,7 @@ async function shellInSession(input: ShellInput, lease: SessionManager.LoopLease
 
   let output = ""
 
-  proc.stdout?.on("data", (chunk) => {
+  const appendOutput = (chunk: Buffer) => {
     output += chunk.toString()
     if (part.state.status === "running") {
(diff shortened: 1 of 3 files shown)


yzxoi:
The architecture-aligned rewrite is now available in #838. It keeps this PRs useful child-process exit/stdio lifecycle work, but replaces the unbounded close wait with a shared bounded drain boundary that preserves tail output and releases inherited descendant pipes deterministically.

Issue #815, opened 2026-07-24
[P1][Linux][area/session] Coalesce full GC at completed tool and turn lifecycles


marzich:
## Summary

Current release signals coalesce, but they always request non-full collection and the critical thresholds are close to the service memory limit. Linux should reclaim eligible completed-turn/tool ownership before cgroup pressure makes the service unresponsive.

## Proposed change

- Trigger collection only from real tool, turn, or runtime release boundaries.
- Coalesce release signals and enforce a full-GC cooldown.
- Enable the lifecycle full-collection policy by default on Linux and make it configurable.
- Keep non-Linux defaults unchanged.
- Never use this policy for task admission, queuing, pausing, merging, cancellation, or prompt reduction.

## Acceptance criteria

- Multiple release signals produce one eligible collection.
- Active streams do not independently trigger full collection.
- Repeated completed workloads stop accumulating reclaimable main-process external/ArrayBuffer memory.
- Existing concurrency behavior is unchanged.

Parent investigations: #777, #813

PR #826, opened 2026-07-24
fix(session): collect pressured Linux turn releases

marzich:
## Summary

- coalesce Linux tool/turn completion signals instead of running a collection per session
- use full GC only at a completed release boundary under soft or critical pressure
- derive cgroup soft pressure from working set (`memory.current - inactive_file - slab_reclaimable`) rather than reclaimable file/slab cache
- track active provider streams so a release cannot trigger full GC while streaming
- enforce a separate 30 second full-GC cooldown
- preserve existing macOS and Windows behavior

Related to #815 and #813. This PR intentionally does not close either issue.

## Scope

This does not compress prompts, reduce concurrency, queue heavy tools, or change tool decisions. Normal low-pressure tool completions do not create a GC timer. Linux turn completion performs the cgroup check; other platforms retain the existing asynchronous collection path.

## Verification

- `bun test test/session/memory-pressure.test.ts test/session/llm-memory.test.ts test/session/tool-scheduler.test.ts test/session/invoke.test.ts` (60 passed)
- repository Prettier, oxlint, Turbo typecheck, and sherif pre-push gates

No production service restart was performed.

### DIFF
--- a/packages/synergy/src/session/invoke.ts
+++ b/packages/synergy/src/session/invoke.ts
@@ -1052,11 +1052,20 @@ loop_stop() does not end the Light Loop directly — a reviewer will audit your 
             postRequestedStop = postResult === "stop"
           }
         }
-        await SessionMemoryPressure.maybeCollect({
-          sessionID,
-          messageID: processor.message.id,
-          phase: "session.turn.after_post_jobs",
-        })
+        if (process.platform === "linux") {
+          SessionMemoryPressure.signalRelease({
+            sessionID,
+            messageID: processor.message.id,
+            phase: "session.turn.after_post_jobs",
+            linuxOnly: true,
+          })
+        } else {
+          await SessionMemoryPressure.maybeCollect({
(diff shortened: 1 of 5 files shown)


marzich:
Combined isolated trajectory evidence is recorded in #813: https://github.com/SII-Holos/synergy/issues/813#issuecomment-5067568097

The candidate stack materially reduced +120s retained heap/external memory, but RSS improved only modestly and +5s RSS was temporarily higher. This supports keeping this PR scoped to a Linux lifecycle collection boundary and avoiding a claim that it alone solves allocator-retained RSS. No production restart was performed.

yzxoi:
The architecture-aligned rewrite is now available in #838. It keeps coalesced lifecycle signals, active-stream protection, and cooldowns, while enforcing the missing ownership rule: service/cgroup pressure cannot trigger GC in the Control Plane for memory owned by other processes.

PR #838, opened 2026-07-24
fix(runtime): align memory recovery with process ownership

### LINKED ISSUE #815: [P1][Linux][area/session] Coalesce full GC at completed tool and turn lifecycles

### LINKED ISSUE #816: [P1][area/tools] Release local process handles after stdio close without losing tail output

yzxoi:
## Summary

- Separate Control Plane process pressure from service-wide cgroup pressure so process-local `Bun.gc()` is never used as if it could reclaim Agent worker, policy worker, browser, or plugin memory.
- Coalesce post-tool and post-turn release signals; allow asynchronous collection for soft local pressure, and gate full collection on critical local pressure, stream settlement, and cooldown.
- Introduce one bounded child-process close boundary for Bash and session shell execution: retain output between exit and stdio close, then detach listeners and handles after a one-second descendant-pipe grace period.
- Add Linux service working-set, `memory.events`, PSI, JavaScriptCore object-growth, and mimalloc diagnostics without adding an observability-triggered GC path.
- Restore the cross-platform runtime memory trajectory benchmark and fingerprint the complete elastic Agent worker lifecycle settings so results are reproducible.
- Update the LLM loop, execution-boundary, and performance-observability documentation to state the ownership and release invariants.

## Ownership model

GC is process-local. Service/cgroup pressure remains useful for admission control and diagnosis, but it cannot justify collecting the Control Plane when memory is owned by another process. Under this change:

- service-only pressure records a conservative skip instead of triggering Control Plane GC;
- soft Control Plane pressure may request asynchronous collection at a real release boundary;
- full collection requires critical Control Plane pressure, no active provider stream, a Linux lifecycle-release signal, and the full-GC cooldown;
- Agent workers retain their own post-turn measurement, recycle, and idle-retirement policies.

The branch is rebased on current `dev`, including the elastic pool lifecycle from #835 and coalesced Agent input ownership from #836.

## Consolidation

This is the rewritten, architecture-aligned follow-up for the useful parts of #823, #826, #819, and #780. It deliberately does not carry forward cgroup-driven main-process GC or unbounded waiting on inherited child stdio.

Related to #813.
Closes #815.
Closes #816.

## Validation

- Targeted runtime/process/observability suites: 77 passed, 1 platform skip
- Benchmark and documentation contract tests: 6 passed
- `bun run quality:quick`
- `bun run typecheck`
- `bun run format:check`
- `bun run lint`
- `bun run perf:memory:runtime --preset smoke` in an isolated temporary runtime; cleanup verified
- `bun run test:changed`: 3441 passed, 2 skipped; the unrelated `test/provider/proxy.test.ts` proxy-hit assertion fails locally and reproduces in its narrow untouched provider suite


yzxoi:
Self-review found six actionable follow-ups before merge:

1. Combined process/cgroup pressure still needs to gate experience re-encoding even though `decision.critical` now intentionally reports process-owned pressure only.
2. The bounded pipe-drain timeout releases stream handles, but an untracked user shell can leave detached descendants alive after the parent exits.
3. A non-release `maybeCollect()` call can still run JSC GC from cgroup-only pressure; collection ownership must remain process-local on Linux for every caller.
4. The runtime benchmark child environment currently inherits more ambient variables than the documented isolated contract permits, including credentials and memory-tuning overrides.
5. The first Linux JSC stats sample reports the entire baseline heap as object growth instead of establishing a baseline.
6. Sequential smoke worker-pool settings use 4 workers while the documented smoke contract says 2.

I am addressing all six on this branch now, with behavioral regressions first, focused implementation changes, documentation updates, and expanded verification against the current `dev` baseline.

yzxoi:
Follow-up self-review fixes are now pushed in `0de043fb`; the branch is also synchronized with the latest `dev` in `f3d3975c`.

Addressed all six findings:

- Experience re-encoding now admits work from combined process/service pressure while keeping JSC collection process-owned.
- Linux cgroup-only pressure never triggers Control Plane GC, including direct `maybeCollect()` callers.
- Pipe-drain timeout settlement now awaits owned process-tree cleanup, including untracked shell descendants, without breaking explicitly authorized detached daemons.
- Runtime benchmark children receive an explicit allowlisted environment instead of ambient credentials or memory-tuning variables.
- The first Linux JSC sample establishes a baseline and reports zero synthetic object growth.
- Sequential smoke now uses the documented two-worker ceiling.

Documentation was updated for execution boundaries, LLM lifecycle, performance observability, and knowledge re-encoding.

Validation completed:

- `bun run quality:quick`
- focused and expanded runtime/process/memory suites
- runtime memory smoke benchmark, sequential preset
- post-merge changed suite: 3,435 passed, 2 skipped; the two local environment/suite-level failures were isolated and passed 42/42 across three reruns
- GitHub CI: all 8 checks pass

The first GitHub attempt had five jobs stop in `bun install` because the registry temporarily reported no matching `@tiptap/extension-link@^3.27.2`. That dependency and lock entry are inherited unchanged from `dev`; another runner installed it successfully in the same attempt. A failed-job rerun passed without any source or lockfile change.

yzxoi:
Update: `dev` advanced with #842 while the validation above was finishing. I synchronized that baseline without a force push in `80371d6c`; it was conflict-free and did not alter the runtime fixes. The fresh CI run for `80371d6c` also passes all 8 checks.

marzich:
## Review findings for current HEAD (`80371d6c`)

I reviewed the current PR head. All reported CI checks are green, and the overall direction is sound: it separates process-local GC from Linux cgroup pressure, centralizes child-process cleanup, narrows inherited environment variables, and treats the first JSC object sample as a baseline.

However, I recommend holding the merge until the two P1 items below are addressed. The P2 items are also important for making the memory recovery and telemetry behavior reliable.

### P1: the new Linux telemetry can amplify observability load substantially

[Resource sampling](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/observability/resources.ts#L112-L116) calls `recordServiceMemoryMetrics(cgroup)` on every resource sample. With the [default 5-second interval](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/observability/config.ts#L76-L83), the implementation can write up to 35 cgroup metric rows per sample ([metric emission](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/observability/resources.ts#L244-L326)).

That is approximately 604,800 cgroup metric rows per day on Linux, before the JSC object metrics. This can cause WAL/checkpoint/retention churn, consume much of the default 250 MB observability budget, and evict more useful diagnostics. In a PR intended to diagnose memory pressure, the telemetry itself should not become a material source of pressure.

Suggested change:

- Rate-limit the detailed cgroup metrics to 30-60 seconds.
- Record static limits only when they change.
- Do not write zero event/stall deltas.
- Keep the detailed breakdown in the existing resource sample where possible.
- Add a test that bounds metric writes over a simulated hour.

### P1: Windows process-tree termination can return before termination completes

The Windows path starts `taskkill /f /t`, but after 200 ms it kills the `taskkill` process and resolves without checking its exit status or verifying that the target process tree exited ([implementation](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/util/shell.ts#L14-L27)).

On a busy Windows host, this can leave descendants alive while the caller assumes cleanup succeeded, retaining memory and handles. The current close tests explicitly skip Windows ([test coverage](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/test/process/child-process-close.test.ts#L15-L65)).

Suggested change:

- Use a Windows-specific timeout measured in seconds, not 200 ms.
- Await and inspect `taskkill` completion.
- Verify the target PID/tree is gone before reporting success.
- Add a Windows test or a deterministic mocked process-runner test.

### P2: a critical release request can be absorbed by an unrelated in-flight collection

[`maybeCollect()`](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/session/memory-pressure.ts#L175-L184) returns the existing `collectionInFlight` promise to every caller. [`flushReleaseSignals()`](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/session/memory-pressure.ts#L288-L303) then consumes the pending release signal and treats that unrelated result as the release result.

If a normal collection is active when a critical Linux release signal arrives, the critical request may never get its own post-collection resample/full-GC decision.

Suggested change: retain the highest-priority pending request and, after the current collection settles, resample and run again when a critical/release request arrived. Add a test for “normal collection in flight, then critical Linux release.”

### P2: JSC growth metrics have unbounded label cardinality

`runtime.jsc.object_type.growth` includes the changing object count as a `total` label ([emission](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/observability/resources.ts#L365-L373)). This creates a new time series whenever the count changes. The existing object-count gauge already carries this value.

Suggested change: remove `total` from the growth metric labels and its catalog declaration.

### P2: failed JSC heap sampling bypasses the intended interval

[`LinuxRuntimeMemory.sample()`](https://github.com/SII-Holos/synergy/blob/80371d6cbeae4743f25b067188d25b35dd44a2a0/packages/synergy/src/observability/linux-runtime-memory.ts#L56-L95) updates `lastSampleAt` only after a successful `heapStats()` call. If `heapStats()` repeatedly throws, it is retried on every resource sample (normally every 5 seconds), despite the intended 60-second sampling interval.

Suggested change: track `lastAttemptAt` independently of successful samples and test repeated failures.

### Benchmark evidence requested

Because this PR changes memory recovery policy and process lifetime behavior, the smoke benchmark alone is not enough to establish safety. Please run the standard trajectory in at least these modes:

- single-session trajectory;
- parallel-session trajectory;
- sequential repeated trajectory.

For baseline and candidate, run each scenario at least three times and compare medians for:

- retained RSS at workload end and after 120 seconds;
- complete process-tree RSS;
- `external` and `arrayBuffers`;
- child-process count and surviving descendants.

The current green CI does not exercise the escalation race, Windows termination boundary, telemetry write rate, or repeated heap-sampling failure described above.

yzxoi:
Thanks for the thorough review and the concrete failure modes. I reproduced the GC coalescing race and the failed heap-sampling retry behavior, and confirmed the cgroup metric volume and the `total` label cardinality issue.

The Windows `taskkill` finding was valid for `80371d6c`. The current head, `6ffefac8`, now assigns Bash, Process, and session-shell trees to gated Windows Job Objects before execution and terminates them with `TerminateJobObject`. I will also harden the remaining shared `Shell.killTree()` fallback so LSP and non-owned process paths await and verify termination.

I am following up on this PR with:

- 5-second core cgroup `current` / `high` / `max` sampling, independently rate-limited detailed cgroup metrics, zero-delta suppression, static-limit change detection plus a low-frequency heartbeat, and a deterministic one-hour metric-write bound test;
- preservation of the highest-priority GC request that arrives during an in-flight collection, followed by a fresh sample and decision that still honors process ownership, active-stream safety, and cooldowns;
- removal of the changing JSC `total` label from growth metrics and the catalog;
- `lastAttemptAt` throttling for failed JSC heap sampling, with repeated-failure coverage;
- hardened shared Windows process-tree termination with deterministic tests;
- standard trajectory, parallel, and sequential runtime-memory benchmark runs after the focused fixes and quality gates pass.

I will push the fixes and validation results to this PR.

yzxoi:
Thanks again for the detailed review. The follow-up fixes have been pushed in commit `05e41923`.

## Fixed

- Kept 5-second resource samples for cgroup `current` / `high` / `max`, while reducing metric-table writes:
- `current` remains at the resource interval;
- `high` / `max` emit on change plus a 5-minute heartbeat;
- detailed cgroup breakdown, event totals, and PSI metrics use an independent 30–60 second interval;
- zero event/stall deltas are omitted.
- Added a deterministic one-hour emission-bound regression.
- Removed the JSC growth `total` label from both emission and the catalog.
- Added `lastAttemptAt` before `heapStats()` and covered repeated failures plus last-good snapshot retention.
- Preserved one highest-priority Linux release request during an in-flight collection, then re-sampled and re-ran the decision after the active collection. The successor still respects process ownership, active-stream safety, and both cooldowns.
- Hardened `Shell.killTree()` on Windows to wait for `taskkill`, bound it with a timeout, check exit/liveness, fall back to direct `SIGKILL`, and remain best-effort/no-reject for recovery call sites.

## Verification

- Focused regressions: **38 passed**.
- Formatting, lint, test-layout, and monorepo typecheck: passed.
- Maintainability, performance, security, and documentation reviews: no blocking findings.
- Full and changed core suites each reached thousands of passing tests but each encountered one unrelated anonymous 5-second global hook timeout in different test regions; the neighboring resolver test passed in isolation.
- `package:check` reached publint but is blocked in this review checkout because its local `origin` path is embedded as `repository.url` rather than a public Git URL.

## Standard runtime memory benchmarks

Final isolated runs passed all health/terminal/pending-reply/tool-state checks and removed their temporary runtimes:

- `trajectory` (1 root + 4 subagents): peak server RSS **1,362,690,048 B**; peak process-tree RSS **3,418,324,992 B**.
- `parallel` (5 primary sessions): peak server RSS **1,096,007,680 B**; peak process-tree RSS **2,829,828,096 B**.
- `sequential` (5 create/delete lifecycles): peak server RSS **1,179,254,784 B**; peak process-tree RSS **2,035,564,544 B**.

CI is now running on the updated PR head.

PR #844, opened 2026-07-24
fix(cli): avoid registering scopes during discovery

yzxoi:
## Summary

- resolve plugin CLI metadata against a transient Scope
- defer persistent Scope registration until a declared plugin command executes
- document and test the side-effect-free discovery invariant

## Verification

- `bun test test/cli/help.test.ts test/plugin/cli-command.test.ts test/scope/project.test.ts`
- `bun run typecheck`
- `bun run quality:quick`
- `bun run test:changed` (3179 passed; two unrelated flaky tests passed on rerun, and the remaining proxy failure reproduces on unchanged `dev`)


yzxoi:
Review follow-up: I found one P2 documentation/contract mismatch.

The implementation correctly prevents Scope registration, `.git/synergy` writes, stale-Scope archival, and Scope events during plugin CLI discovery/help. However, the new wording says discovery/help is fully “metadata-only” and “side-effect free”, while `getLoadedPlugins()` still calls `Config.current()` (which may migrate/archive legacy project config) and may install non-`file://` plugin specs into the cache.

I’ll narrow the docs and Skill wording to the actual invariant: discovery/help must not cause Scope-registration side effects. The implementation itself can remain unchanged.

PR #849, opened 2026-07-24
fix(provider): stop idle timeout from retaining streamed chunks

### LINKED ISSUE #848: Provider idle-timeout wrapper retains every streamed chunk until turn completion
## Summary

`ProviderStream.withIdleTimeout()` retains one settled `reader.read()` result for every provider response chunk for the entire stream lifetime. On long-running, highly fragmented LLM streams this grows the active Agent worker to roughly 1 GiB even when parent-visible text/tool payloads are small. The worker memory watermark then deliberately terminates the worker and the session reports:

```text
Error: Agent worker exited (SIGTERM)
```

This affects multiple providers. It has now been observed with both OpenAI Codex and a correctly labelled `text/event-stream` response from DeepSeek.

## Root cause

The idle-timeout wrapper creates one shared pending promise for the full stream:

```ts
const aborted = new Promise<never>((_, reject) => { ... })
```

Every `pull()` then runs:

```ts
await Promise.race([reader.read(), aborted])
```

When `reader.read()` wins, the reaction attached to the still-pending shared `aborted` promise cannot be removed. That reaction retains the settled race promise and its `{ done, value }` result, including the chunk buffer. Repeating this for every chunk produces linear live-memory growth with total bytes read/chunk count. GC cannot collect the chunks because the shared pending promise is still reachable from the active stream wrapper.

This path was introduced in `0e1e6f2e` while fixing streaming GC stalls.

### DeepSeek incident

A temporary isolated Synergy instance failed a session using:

```text
provider: deepseek
model: deepseek-v4-flash
```

The provider returned a correctly labelled response:

```text
HTTP 200
Content-Type: text/event-stream; charset=utf-8
```

The active worker grew as follows during one turn:

| Time (UTC) | RSS | heapUsed | ArrayBuffers |
|---|---:|---:|---:|
| 13:27:25 | 224 MiB | 79 MiB | 1.7 MiB |
| 13:27:55 | 470 MiB | 369 MiB | 252 MiB |
| 13:28:25 | 748 MiB | 637 MiB | 524 MiB |
| 13:28:55 | 1,029 MiB | 923 MiB | 796 MiB |
| 13:29:10 | 1,154 MiB | 1,036 MiB | 928 MiB |

A forced `Bun.gc(true)` left ArrayBuffers unchanged at about 928 MiB. The parent only received small IPC frames during the turn plus one approximately 35 KiB tool-input frame, so the retained memory was inside the worker/provider stream path rather than persisted session or tool payloads.

The worker was terminated by Synergy's active-worker heap watermark. This was not Linux OOM and the server process remained alive.

### Codex incident

A separate Codex turn showed the same signature: parent-visible output stopped while the worker accumulated about 994 MiB ArrayBuffers, post-collection memory did not fall, and the watermark terminated the worker. A missing Codex response `Content-Type` was initially suspected, but the DeepSeek reproduction proves that header handling is not the common root cause.

## Deterministic local reproduction

A local stream passed 512 fresh chunks of 64 KiB through `ProviderStream.withIdleTimeout()` and discarded every chunk immediately. Total data read was 32 MiB.

Before the fix, after stream completion and `Bun.gc(true)`:

```text
retained heap delta: ~32.8 MiB
WeakRef live chunks: 64 / 64
```

After replacing the shared abort promise with a per-read abort waiter that removes its listener in `finally`:

```text
retained heap delta: ~0.44 MiB
WeakRef live chunks: 0 / 64
```

The comparison uses no provider, network call, prompt, tool execution, or workload concurrency.

## Proposed fix

- Create the abort waiter inside each `readWithAbort()` call.
- Remove that read's abort listener in `finally`, regardless of read or abort outcome.
- Keep cancellation of the owned reader on errors/timeout.
- Preserve the existing pull-based stream, idle timer, downstream cancellation, and reader release behavior.

## Acceptance criteria

- Completed provider chunks are collectable while the wrapper/reader remains reachable.
- Idle timeout still aborts and cancels the upstream reader.
- Downstream cancellation and upstream errors still release the reader lock.
- SSE parser-bound behavior remains unchanged.
- Long provider streams no longer grow Agent worker memory in proportion to all bytes/chunks previously read.

marzich:
## Summary

- stop `ProviderStream.withIdleTimeout()` from retaining every settled `reader.read()` result for the lifetime of an active LLM stream
- replace the shared never-settling abort promise with a per-read abort waiter whose listener is removed in `finally`
- add a focused GC regression test proving completed chunk buffers are no longer retained

Addresses #848. The issue is intentionally left open for production validation.

## Root cause

The previous implementation reused one pending `aborted` promise in every `Promise.race([reader.read(), aborted])`. Each completed read left a reaction attached to that promise, retaining the settled race result and its chunk buffer. Long fragmented streams therefore accumulated live memory until the Agent worker hit its 1 GiB watermark and exited with SIGTERM.

This was reproduced across Codex and DeepSeek. It is not provider-specific and was not a Linux OOM.

## Changes

| File | Change |
|---|---|
| `packages/synergy/src/provider/stream.ts` | Scope the abort waiter to one read and remove its listener after the read settles |
| `packages/synergy/test/provider/sse-buffer.test.ts` | Verify completed chunks are collectable while the wrapper remains reachable |

## Measured result

A bounded local reproduction read 512 fresh 64 KiB chunks (32 MiB total), discarded them immediately, then ran `Bun.gc(true)`:

| Version | Retained heap delta | Live WeakRef chunks |
|---|---:|---:|
| before | ~32.8 MiB | 64 / 64 |
| this PR | ~0.44 MiB | 0 / 64 |

## Verification

- `bun test test/provider/sse-buffer.test.ts` - 8 passed
- `bun run typecheck` - 11 monorepo tasks passed
- `bun run format:check` - passed
- `bun run lint` - passed
- `bun run monorepo:check` - passed
- `git diff --check` - passed

No production service was restarted as part of this PR workflow.


yzxoi:
I reproduced a reliability issue in the new WeakRef regression test before changing the implementation.

The production fix looks correct, but the test currently assumes that two `Bun.gc(true)` calls must make every chunk WeakRef disappear immediately. That is not guaranteed: JSC/Bun may keep one recent local or temporary value alive across those collections even though there is no stream-lifetime retention.

On the PR head, running only this test in 30 fresh Bun processes produced 17 failures, all with:

```text
Expected length: 0
Received length: 1
```

This is materially different from the original bug, where all 64 chunks remain reachable, but the exact-zero assertion makes the regression test highly flaky.

I’m going to adjust the test so the stream-consumption frame fully unwinds and the assertion distinguishes bounded transient retention from the original O(n) retention, then stress-run it before pushing the follow-up commit.

yzxoi (approved):

PR #852, opened 2026-07-24
feat(channel): harden native Clarus task runtime

marzich:
## What does this PR do?

- integrates Clarus as a native top-level operation and event family over the existing authenticated Holos Agent Tunnel
- maps remote Clarus Projects and Tasks to Channel-owned managed Scopes and stable unattended Sessions
- adds replay-safe assignment delivery, deadline guidance, extension/result outboxes, archive semantics, bounded diagnostics, Web navigation and settings controls, generated SDK contracts, and the built-in Clarus participation tools and Skill
- hardens protocol recovery after review feedback: validates identity, attempt, and deadline semantics without truncation; discovers paused Projects; persists extension intent before projection; and fails closed on oversized/deep input references
- preserves Channel config reload cascades and independent reply/push provider capabilities across the rebase onto current `dev`
- restores `ToolScheduler` availability after graceful shutdown, fixing an order-dependent full-suite failure discovered by the PR quality gate

This PR supersedes #686 with the current native Channel architecture and hardening work.

## Why?

Clarus tasks need to participate in Synergy as first-class Channel work without introducing a second WebSocket, standalone adapter, or parallel project/session model. The runtime must preserve account identity, managed Project ownership, stable Session routing, replay convergence, result acknowledgement, deadline behavior, diagnostics redaction, and existing Feishu behavior across reconnects and restarts.

The follow-up hardening closes five confirmed review findings at their owning boundaries rather than masking them in callers:

- malformed external identifiers and assignment metadata are rejected at tunnel parsing
- active and paused Project discovery are paginated separately per the Holos API contract
- extension outbox intent is durable before assignment state changes
- input-reference count and traversal limits fail closed
- interrupted extension recovery remains idempotent

## How was it tested?

- `bun test test/channel/` from `packages/synergy` — 252 passed, 0 failed
- `bun test test/runtime/reload.test.ts test/server/config-domain-route.test.ts` from `packages/synergy` — 27 passed, 0 failed
- `bun test test/channel/outbound.test.ts` from `packages/synergy` — 6 passed, 0 failed
- `bun run quality:quick`
- pre-rebase maintainability, security, and API compatibility reviews found no blockers
- post-rebase maintainability review of conflict-resolution and integration changes found no blockers

## Checklist

- [x] Rebases cleanly onto current `origin/dev`
- [x] All five Synergy Test / PR review findings were reproduced and fixed
- [x] Relevant narrow and full Channel tests pass
- [x] New tests live under the owning package's `test/` directory
- [x] `bun run package:check` passes as part of `quality:quick`
- [x] `bun run localization:check` passes as part of `quality:quick`
- [x] Channel architecture, configuration reference, built-in Skill, and package guidance remain synchronized
- [x] No local auth files, unrelated cleanup, placeholder scripts, or redundant compatibility wrappers are included


yzxoi:
## Architecture follow-up from #686

The overall architecture in this PR is the right successor to #686. It correctly converges on the ownership chain agreed in that discussion:

```text
Holos transports
→ Clarus Channel provider interprets and reconciles
→ Channel owns managed Project Scope and Task Session routing
→ Scope and Session own the local work
```

In particular, the move to a `task_only` / `borrowed_transport` Channel provider, canonical managed Project Scopes, typed Channel task targets, Channel-owned navigation, provider-namespaced durable state, and removal of the parallel Clarus runtime/navigation model are all aligned with the final #686 direction.

However, the current implementation still violates several of the correctness invariants that were explicitly preserved in the #686 architecture discussion. I think the first two are merge blockers.

### 1. Blocker: active-only discovery is treated as a complete lifecycle snapshot

`ClarusProjectClient.listProjects()` always sends `status=active`:

- `packages/synergy/src/channel/provider/clarus/project-client.ts:42-46`

`ClarusProvider.syncProjects()` then passes the collected result to `host.projects.reconcile({ complete: true })` after pagination:

- `packages/synergy/src/channel/provider/clarus/index.ts:324-336`

Complete Channel reconciliation archives every previously owned Project that is absent from the supplied set. A Project that changes from `active` to `paused` therefore disappears from discovery and is incorrectly projected as `archived`, even though `ManagedProjectOwnership` explicitly models `paused` as a live remote state. This can block new task routing and remove the active/paused local archive guard from a still-valid remote Project.

Negative reconciliation must only use an authoritative snapshot containing every relevant lifecycle state. Please either fetch active and paused Projects as one complete snapshot, or mark an active-only result as partial so it cannot archive absent ownership.

Longer-term, this API would be safer as a discriminated discovery result rather than `projects + complete: boolean`, for example `kind: "complete" | "partial"`, with an explicit remote state instead of `isActive: boolean`.

### 2. Blocker: external identities are truncated before canonical routing and persistence

`Bounds.id()` silently truncates strings to `NATIVE_MAX_ID_LENGTH`:

- `packages/synergy/src/channel/provider/clarus/tunnel-adapter.ts:139-145`

The truncated values are then used for `projectID`, `runID`, `taskID`, `phase`, `subtaskID`, attempt mode, and retry lineage:

- `packages/synergy/src/channel/provider/clarus/tunnel-adapter.ts:215-234`

For identity-bearing fields this is unsafe. Two remote IDs that differ only after character 256 collapse onto the same managed Project ownership, Channel task endpoint, assignment hash, replay identity, or result/extension state. That can route one remote entity into another entity’s durable local state.

Transport/display bounds must not become lossy domain identity normalization. Identity fields should be validated losslessly at the adapter boundary and rejected as an invalid event when they exceed the supported contract. Truncation is appropriate for display or diagnostic text only, ideally with explicit truncation metadata.

### 3. Extension outbox does not satisfy durable-before-state crash consistency

The result pipeline persists its outbox record before marking the assignment pending. The extension pipeline currently does the reverse:

- `packages/synergy/src/channel/provider/clarus/extension-outbox.ts:175-189`

`beginExtension()` durably sets `extensionState: "pending"` before the extension outbox record is written. A crash or storage failure between those operations leaves a pending assignment with no recoverable outbox record. Recovery begins by scanning outbox records, so it cannot repair this state, while future extension attempts are rejected because the assignment remains pending.

The extension pipeline should follow the same authoritative ordering as result delivery:

```text
persist outbox intent
→ project pending state onto assignment
→ send
→ settle outbox
→ settle assignment projection
```

The outbox should remain the recovery root; assignment result/extension fields should be treated as projections of that durable intent and outcome.

### 4. Input hydration bounds silently omit declared dependencies

`collectInputRefs()` returns once the set reaches `MAX_REFS`:

- `packages/synergy/src/channel/provider/clarus/assignment-preflight.ts:35-57`

This silently drops later declared `input_refs`. Preflight then resolves only the truncated list and can continue to Session creation and model wake without noticing that required inputs were omitted. A single encountered `input_refs` array can also exceed the intended cap because the inner loop does not enforce the limit.

For execution-control data, bounds must fail closed rather than return a partial semantic value. Reaching the reference count or traversal-depth limit should reject the assignment before Session creation, assignment persistence, inbox delivery, or model wake. Silent truncation should be reserved for non-authoritative display/diagnostic data.

### 5. Assignment semantic validation occurs after local side effects

The inbound assignment schema currently accepts any finite number as `attempt` and any string as `deadline_at`:

- `packages/synergy/src/channel/provider/clarus/tunnel-adapter.ts:58-69`

As a result, values such as `attempt: 0`, negative/fractional attempts, or `deadline_at: "not-a-date"` are emitted as known assignments. The attempt may fail only when assignment persistence validates it, after Channel has already created a Session and delivered the visible task message. An invalid deadline can fail later during Agenda synchronization after the assignment has been persisted.

The protocol-to-semantic adapter is the trust boundary. Once it emits `RuntimeTaskAssignedEvent`, the event should already satisfy all invariants needed for dispatch. Please validate a positive integer attempt and a strict parseable deadline in the wire schema so malformed events are classified as invalid before any Scope, Session, Inbox, assignment, or Agenda side effect.

## Recommendation

This PR does not need another architectural rewrite: the owner boundaries are now substantially correct and aligned with #686. The remaining work is to strengthen four boundary contracts:

1. **Identity must be lossless.**
2. **Negative reconciliation requires an authoritative complete snapshot.**
3. **Outbox intent must be the durable recovery root.**
4. **Bounded semantic input must fail closed rather than truncate.**

Please fix the discovery/lifecycle and identity issues before merging. I also recommend addressing the extension write order and fail-closed assignment validation in this PR because they are localized within the current owners and otherwise leave persistent unrecoverable or partially delivered task states.

marzich:
I reproduced one more runtime lifecycle issue while testing this branch from a fresh Synergy home.

### Reproduction

1. Start Synergy with no Holos identity or Channel configuration.
2. Import a Holos Agent identity in Settings.
3. Enable the automatically provisioned Clarus account.
4. Observe that the Holos Agent Tunnel reaches `connected`, but the Clarus Channel does not connect.

The runtime log reports:

```text
unknown channel type, skipping
```

Calling the existing `POST /channel/start` endpoint immediately resolves the issue. The Clarus status then converges through:

```text
waiting_for_transport → connecting → connected
```

### Root cause

The credentials route persists the Holos identity, provisions the Clarus account, and reloads `HolosRuntime`. Updating the Channels config invokes `Channel.reload()`, but that runtime reload path does not ensure built-in Channel providers are registered. `global-runtime` startup and the explicit Channel start endpoints do call `registerProviders()`, so a fresh process works after restart or explicit start, while runtime-first setup can skip the newly configured `clarus` type.

### Expected behavior

After importing the first Holos identity and enabling Clarus, the configured Clarus account should connect without requiring a process restart or a manual Channel start call.

I am preparing a small follow-up patch with a behavioral regression test at the runtime reload boundary.

marzich:
The two reproduced first-run lifecycle fixes are now included directly in this PR branch in commit `ca9a6541`:

- runtime Channel reloads register built-in providers before reconnecting configured accounts
- non-archived `scope.updated` events trigger the existing debounced Scope-index refresh, so asynchronously discovered managed Clarus Projects appear in the Sidebar

Verification completed:

- runtime reload tests: 22 passed
- Channel tests: 253 passed
- App Layout tests: 82 passed
- Synergy and App typechecks passed
- App production build passed
- tracked-source quality gates passed
- independent maintainability review: PASS

No separate patch PR was created.

yzxoi:
@codex review

yzxoi (dismissed):

yzxoi (approved):

PR #853, opened 2026-07-24
fix(process): harden owned child cleanup

### LINKED ISSUE #850: fix(session): drain timeout skips descendant process-group cleanup after parent exit
## Summary

The bounded child-process drain handling merged in #838 does not terminate an inherited-pipe descendant in the direct session-shell path on Unix.

When the parent shell emits `exit`, `shellInSession()` sets its local `exited` flag to `true`. If a descendant still owns stdout/stderr, the close grace timer later calls `terminate()`, which passes `exited: () => exited` to `Shell.killTree()`. `Shell.killTree()` immediately returns when that callback is true, so the process group remains alive.

This contradicts the intended #838 invariant that an untracked descendant cannot outlive the direct user-shell request after the bounded drain window.

Related regression: #838

Merge commit: `43e7e1bbf4208d31784b55441509b622d1c4c626`

## Affected code

- `packages/synergy/src/session/shell.ts`: `onExit()` sets `exited = true`; `onDrainTimeout()` then calls `terminate()`.
- `packages/synergy/src/util/shell.ts`: `killTreeOnce()` returns before signaling the process group when `didExit(opts?.exited)` is true.

## Deterministic reproduction

On Linux, spawn a detached shell whose child inherits its pipes:

```ts
const child = spawn("/bin/sh", ["-c", "(sleep 30) &"], {
  detached: true,
  stdio: ["ignore", "pipe", "pipe"],
})

let exited = false
child.once("exit", () => {
  exited = true
})
await once(child, "exit")
await Shell.killTree(child, { exited: () => exited })
process.kill(-child.pid!, 0) // still succeeds
```

Observed on the merged code:

```json
{"parentExited":true,"groupAliveAfterKillTree":true}
```

## Impact

- Background descendants can survive after the session shell is reported complete.
- Surviving processes may continue consuming CPU, memory, file descriptors, and workspace resources.
- The new close boundary destroys local stream handles afterward, making the surviving process less visible to the session runtime.

## Suggested direction

Separate parent-process exit state from process-group/tree liveness. Drain-timeout cleanup must be able to signal the owned process group even after the direct parent has exited. Keep the existing early-exit optimization for ordinary cancellation, but do not apply it to descendant cleanup after an inherited-pipe timeout.

## Acceptance criteria

- A direct session shell with an inherited-pipe descendant settles within the configured drain grace period.
- The descendant process group is no longer alive when settlement completes.
- Finite tail output produced during the grace period is still retained.
- Explicitly supported detached/background execution semantics remain unchanged.
- Add an integration test through the session-shell callback path with `onExit` already observed before `onDrainTimeout`; the current lower-level test calls `Shell.killTree(child)` without the `exited` callback and therefore does not cover this regression.

### LINKED ISSUE #851: fix(windows): make Job Object termination fail-safe and kill descendants on owner exit
## Summary

The Windows Job Object ownership added in #838 is not fail-safe on termination or owner-process exit:

1. `Owner.terminate()` ignores a `false` return from `TerminateJobObject()`, closes the only Job handle, and marks the owner released. The caller receives success and cannot retry or use the shared fallback.
2. The Job is created with `CreateJobObjectW()` but is never configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. If Synergy exits unexpectedly, closing the owner handle does not guarantee that associated child processes terminate.

Related regression: #838

Merge commit: `43e7e1bbf4208d31784b55441509b622d1c4c626`

## Affected code

- `packages/synergy/src/process/windows-process-job.ts`
  - `attach()` creates and assigns the Job.
  - `Owner.terminate()` discards the `terminateJob()` result and always calls `release()`.
  - `loadRuntime()` does not expose or call `SetInformationJobObject()`.

The repository's Rust sandbox helper already demonstrates the expected Windows ownership configuration using `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.

## Deterministic unit evidence

With a runtime whose `terminateJob()` returns `false`, two calls to `owner.terminate()` produce:

```json
["close:2","terminate-failed","close:1"]
```

The Job handle is closed after the first failed termination, and the second call cannot retry.

## Impact

- A failed Job termination can leave a Bash or session-shell process tree alive while cleanup reports completion.
- A Synergy crash can leave owned Windows descendants running because Job-handle closure has no kill-on-close policy.
- Surviving descendants can retain memory, CPU, files, pipes, and workspace locks.

## Suggested direction

- Configure every owned Job with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` before assigning/releasing the gated process.
- Treat `TerminateJobObject() == false` as a real failure: retain ownership for retry or invoke a verified fallback instead of silently releasing the handle.
- Propagate enough structured status for callers and observability to distinguish successful termination from best-effort failure.

## Acceptance criteria

- Closing the last owner handle terminates associated descendants.
- A failed `TerminateJobObject()` does not silently release ownership or report success.
- Explicit termination remains bounded and can retry or fall back safely.
- Add mocked tests for `TerminateJobObject() == false`.
- Add a Windows integration test confirming descendant termination on explicit cleanup and owner-process exit.

marzich:
## Summary

- allow Unix drain-timeout cleanup to terminate an owned process group after the direct parent has exited
- configure Windows Job Objects with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` before process assignment
- retain Windows Job ownership when explicit termination fails, while preserving explicitly authorized detached-daemon behavior

Closes #850
Closes #851

## Changes

| Area | Change |
| --- | --- |
| Unix process cleanup | Added an explicit drain-cleanup mode that bypasses only the direct-parent exit shortcut and wired it through session shell and Bash process registry paths |
| Windows process ownership | Configured owned Jobs for kill-on-close, checked `TerminateJobObject` failures, and retained handles for retry |
| Detached daemons | Kept recognized, explicitly authorized detached daemon commands outside the kill-on-close Job |
| Tests | Added parent-exited drain cleanup, Windows Job configuration/failure/retry, and registry propagation coverage |
| Documentation | Updated the execution-boundary ownership guarantees |

## Verification

- `bun test test/process/child-process-close.test.ts test/process/windows-process-job.test.ts test/process/registry.test.ts test/util/shell.test.ts test/tool/bash.test.ts` (73 passed)
- `bun run typecheck` (11 packages passed)
- `bun run format:check`
- `bun run quality:quick`
- pre-push `prettier`, `oxlint`, typecheck, and `sherif` checks

## Platform notes

- The Unix descendant cleanup integration test ran on Linux.
- Windows Job behavior is covered with deterministic mocked Win32 runtime tests; native Windows integration remains appropriate for CI/maintainer verification.


yzxoi:
## Review findings and follow-up

Thanks for the focused follow-up. The underlying fixes are necessary: I reproduced the Unix parent-exited drain regression on the base commit, and the PR fixes the same-process-group case. Configuring Windows Jobs with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` before opening the execution gate is also the right ownership direction.

Before merging, I found three remaining gaps that I will address on this PR branch:

1. **Unix drain cleanup still relies on the exited parent’s numeric PID/PGID.** A descendant can create a new session/process group while retaining inherited stdout/stderr; `kill(-parentPid)` then misses it. The old PGID can also eventually be reused after the leader exits, so it is not a durable ownership token.
2. **Windows termination-failure ownership is retained only inside `Owner.terminate()`, not by production callers.** Bash and direct session shell currently catch `TerminateJobObject` failure, immediately close and clear the owner, and ignore `CloseHandle` failure. This prevents a verified retry/fallback path.
3. **Authorized detached daemons bypass only the unsandboxed TypeScript Job path.** Guarded/autonomous Windows execution normally uses the sandbox helper, whose Job has kill-on-close and an active-process limit, so the documented authorized-daemon exception does not hold under the default sandbox path.

I am adding behavioral coverage at the owning boundaries and will push a focused follow-up that keeps the intended Unix drain bound, makes Windows close/termination outcomes explicit, and aligns detached-daemon handling with the actual sandbox path.

yzxoi (approved):

PR #858, opened 2026-07-24
fix(tools): enforce hard execution deadlines

### LINKED ISSUE #857: bug(tools): file_search can remain running forever after its timeout
## Summary

A `file_search` call can remain in `running` state indefinitely even though the tool metadata records a finite timeout. This blocks the assistant turn and leaves the session apparently frozen.

## Incident evidence

- Session: `ses_0c08b048cffezNOFKbauzhUL0U`
- Tool call: `call_G2ffC6HxdqVMmf7Afr5qV7Vk`
- Input: `query=holos.main`, `limit=30`
- The call entered `tool.execute.start` and remained `running` for more than 450 seconds.
- Its metadata recorded `toolTimeoutMs=150000`.
- The tool trace continued emitting heartbeat/stalled events after the advertised deadline, with no terminal outcome.
- No `rg`, Git, or file-search child process was still present when inspected.
- Other parallel glob/file_search calls in the same turn completed in milliseconds.
- Re-running the individual file-search phases against the same workspace in an isolated process completed in about 2.6 seconds, so the query and workspace are not inherently pathological.
- The same session contains several older `file_search` parts left in `running`, showing this is a recurring lifecycle failure rather than one malformed query.

## Proven root cause

`ToolResolver.startToolTimeout()` only aborts an `AbortController`. Both builtin and MCP execution paths then directly await the tool promise. If a tool is blocked in an operation that ignores or cannot settle on abort, the resolver never regains control, never marks the slot failed, and never ends the trace. The configured tool timeout is therefore advisory rather than a hard deadline.

This differs from the session invoke path, which already races execution against a deadline because a wedged subprocess may not observe signals.

## Likely blocked owner for file_search

The path-search phase can await index refresh, filesystem node hydration, and a shared Git status cache. Those operations do not all honor the incoming abort signal. The exact inner await cannot be proven from current stage-level telemetry, but the resolver-level failure is sufficient to explain why the call remains permanently running after timeout.

## Expected behavior

- At the configured deadline, abort the cooperative work.
- Independently race the execution promise against the abort/deadline so an uncooperative tool cannot keep the slot running forever.
- Persist a terminal timeout error and allow the parent turn to continue/fail normally.
- Apply the same contract to builtin and MCP tools.

## Proposed focused fix

Introduce a small execution deadline helper used by both resolver paths. It should expose the combined session/timeout signal and a `run()` method that races tool execution against cancellation. Add a regression test with a tool promise that never settles and ignores abort.

This issue does not claim that the underlying index/Git operation should remain unbounded; follow-up stage instrumentation and cancellation should still be considered.

marzich:
## Summary

- make configured tool deadlines terminal even when an implementation ignores abort
- keep cooperative cancellation by passing the same combined abort signal to the tool
- apply the hard deadline contract to both builtin and MCP execution paths
- add a regression test with an execution promise that never settles

Fixes #857.

## Root cause

`ToolResolver` previously only aborted a signal at the configured deadline, then continued awaiting the original execution promise. A blocked `file_search` path that did not settle on abort therefore remained `running` indefinitely. The incident call exceeded its recorded 150-second timeout and continued emitting heartbeats beyond 450 seconds.

## Changes

| File | Change |
|---|---|
| `packages/synergy/src/tool/timeout.ts` | Add a disposable execution deadline that combines cancellation with a hard promise boundary |
| `packages/synergy/src/session/tool-resolver.ts` | Run builtin and MCP execution promises through that deadline |
| `packages/synergy/test/tool/timeout.test.ts` | Verify a never-settling operation rejects with `TimeoutError` |

## Verification

- `bun test test/tool/timeout.test.ts` (7 passed)
- Prettier check on changed files
- Oxlint on changed files
- `git diff --check`

## Scope

This prevents permanent session blockage without reducing search scope, permissions, or concurrency. It does not claim to identify the exact internal file-index/Git-status await from the incident; stage-level cancellation and telemetry can be improved separately.

### DIFF
--- a/packages/synergy/src/session/tool-resolver.ts
+++ b/packages/synergy/src/session/tool-resolver.ts
@@ -1,5 +1,6 @@
 import { Global } from "@/global"
 import { type Tool as AITool, tool, jsonSchema, type ToolCallOptions, type JSONSchema7 } from "ai"
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js"
 import z from "zod"
 import { Agent } from "@/agent/agent"
 import { Identifier } from "@/id/id"
@@ -611,14 +612,15 @@ export namespace ToolResolver {
 
   function startToolTimeout(ctx: Tool.Context, timeoutMs: number) {
     const timing = toolTiming(ctx)
-    const timeout = new AbortController()
-    const timer = setTimeout(() => timeout.abort(), timeoutMs)
-    if (typeof timer === "object" && "unref" in timer) timer.unref()
+    const deadline = ToolTimeout.executionDeadline({
+      signal: timing.sessionAbort,
+      timeoutMs,
+    })
     timing.toolTimeoutCleanup = () => {
-      clearTimeout(timer)
+      deadline.dispose()
       timing.toolTimeoutCleanup = undefined
     }
-    return AbortSignal.any([timing.sessionAbort, timeout.signal])
+    return deadline
   }
 
   function disposeToolTimeout(ctx: Tool.Context) {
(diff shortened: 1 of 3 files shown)


yzxoi:
Superseded by #864, which has been merged and closed #857.

The investigation found that the timeout signal was already firing, but `file_search` could remain pending inside ripgrep stream cleanup: `ProcessOutput.lines()` awaited cancellation of a Bun subprocess pipe reader, which may never resolve, and `Ripgrep.matches()` did not propagate the abort signal to its stderr drain.

#864 fixes those cooperative-cancellation gaps at the owning stream/ripgrep layer and adds regression coverage. A resolver-level hard deadline would return before the underlying operation settles and could leave zombie work behind, so this PR should be closed in favor of #864.

PR #867, opened 2026-07-25
fix(library): synthesize users for re-encode jobs

### LINKED ISSUE #866: bug(library): bulk experience re-encode passes an invalid empty user to agent workers
## Summary

Bulk experience repair can fail every selected intent/script before making an LLM request. The re-encode helpers pass `{}` as `MessageV2.User` when operating from stored experience content without hydrating the original session message. Agent worker input validation then rejects the request because `input.user.id` is missing.

## Production evidence

A repair scan selected:

- 475 intent records
- 146 script records

The intent job completed as `0 updated, 0 skipped, 475 failed` in roughly two seconds, despite estimating 475 model calls. Every sampled failure had the same schema error:

```text
path: input.user.id
Invalid input: expected string, received undefined
```

Logs show model/system/options assembly followed immediately by `library.reencode` failure. No model response was attempted.

## Root cause

`ExperienceEncoder.reencodeIntent()` uses an empty cast when the target message is absent:

```ts
userMsg: userInfo ?? ({} as MessageV2.User)
```

`reencodeScript()` always uses the same invalid placeholder:

```ts
userMsg: {} as MessageV2.User
```

This is common for maintenance jobs because `ExperienceReencode.processItem()` intentionally uses stored `experience_content.user_input/raw` and avoids loading full session history. The placeholder reaches `AgentCall.text({ user })`; since `user` is present, `AgentCall` does not synthesize its valid temporary user, and the worker protocol rejects the malformed object.

The existing tests mock `LLM.stream` below worker input validation, so they accept the invalid cast and miss the production failure.

## Expected behavior

When no original user message is available, omit `user` and provide `sessionId`. `AgentCall.text()` already synthesizes a schema-valid temporary user using the resolved agent/model. Existing real user messages should continue to be forwarded unchanged.

## Proposed focused fix

- Make the encoder agent context user optional.
- Pass `undefined` for stored-content intent/script repairs instead of `{}` casts.
- Let `AgentCall.text()` synthesize the temporary user.
- Add tests at the AgentCall boundary so missing-history intent and script repairs cannot regress.

This does not change candidate classification, prompt content, repair concurrency, or model selection.

marzich:
## Summary

- stop passing empty objects as `MessageV2.User` during stored-content experience repair
- let `AgentCall.text()` synthesize a schema-valid temporary user when history is unavailable
- cover both intent and script fast paths with downstream user-envelope assertions

Fixes #866.

## Root cause

Maintenance re-encode intentionally avoids hydrating complete session history when stored `user_input` or `raw` content is sufficient. Those paths supplied `{}` as a user message. Since `AgentCall.text()` received a truthy `user`, it skipped its normal synthesis path, and the worker protocol rejected `input.user.id` before the LLM request.

## Changes

| File | Change |
|---|---|
| `packages/synergy/src/library/experience-encoder.ts` | Make encoder user context optional and rely on AgentCall synthesis when absent |
| `packages/synergy/test/library/experience-reencode.test.ts` | Assert valid generated users for stored-content intent and script repairs |

## Verification

- `bun test test/library/experience-reencode.test.ts` (21 passed)
- Prettier on changed files
- Oxlint on changed files
- `git diff --check`

## Scope

No changes to candidate classification, prompts, repair concurrency, history-loading policy, embedding, or model selection.

### DIFF
--- a/packages/synergy/src/library/experience-encoder.ts
+++ b/packages/synergy/src/library/experience-encoder.ts
@@ -116,7 +116,7 @@ export namespace ExperienceEncoder {
     const userInfo = userMsg?.info as MessageV2.User | undefined
     const ctx: AgentContext = {
       sessionID,
-      userMsg: userInfo ?? ({} as MessageV2.User),
+      userMsg: userInfo,
       model,
       learning: effectiveLearning,
       signal,
@@ -157,7 +157,6 @@ export namespace ExperienceEncoder {
 
     const ctx: AgentContext = {
       sessionID,
-      userMsg: {} as MessageV2.User,
       model,
       learning: effectiveLearning,
       signal,
@@ -518,7 +517,7 @@ export namespace ExperienceEncoder {
 
   interface AgentContext {
     sessionID: string
-    userMsg: MessageV2.User
+    userMsg?: MessageV2.User
     model: Provider.Model | undefined
     learning: Required<Config.Learning>
     signal?: AbortSignal
@@ -535,7 +534,6 @@ export namespace ExperienceEncoder {
   }
 
   async function runAgent(agentName: string, ctx: AgentContext, content: string): Promise<string> {
-    if (!ctx.userMsg) return ""
     try {
       const result = await AgentCall.text({
         agent: agentName,
--- a/packages/synergy/test/library/experience-reencode.test.ts
+++ b/packages/synergy/test/library/experience-reencode.test.ts
@@ -456,6 +456,18 @@ describe.serial("ExperienceReencode bounded session loading", () => {
           raw: "### User\nRepair the build\n\n### Response\nThe build is repaired.",
         })
         installReencodeModelMocks()
+        let modelUser: MessageV2.User | undefined
+        ;(LLM.stream as any) = mock(async (input: { user: MessageV2.User }) => {
+          modelUser = input.user
+          const text =
+            "1. Read the stored experience\n2. Rebuild the reusable trajectory\n3. Persist the fresh embedding"
+          return {
+            textStream: (async function* () {
+              yield text
+            })(),
+            text: Promise.resolve(text),
+          }
+        })
 
         let embeddingSignal: AbortSignal | undefined
         ;(Embedding.generate as any) = mock(async (input: { id: string; signal?: AbortSignal }) => {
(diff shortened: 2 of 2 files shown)


yzxoi:
LGTM — approved.

This is a necessary and well-scoped fix for the production re-encode failure described in #866. The previous empty-object casts bypassed AgentCall.text()’s temporary-user synthesis and caused every stored-content repair to fail worker validation before reaching the model.

The change fixes the issue at the correct boundary: it removes the invalid MessageV2.User placeholders, keeps session-history loading bounded, and reuses the existing sessionless AgentCall identity path rather than duplicating synthesis logic or weakening the worker schema. The normal encoding and reward paths continue to forward real user messages unchanged.

I reviewed the full ExperienceReencode → ExperienceEncoder → AgentCall → AgentTurn flow and found no blocking design or correctness issues. The new intent and script fast-path assertions cover the regression, focused tests and typecheck pass, CI is green, and the PR merges cleanly onto the current dev.

yzxoi (approved):

marzich:
## Production verification (Linux, real stored data)

I deployed this commit locally, restarted Synergy, and ran the durable re-encode jobs against the production experience database. This validates the fix through the real agent-worker boundary rather than only the test-mode `LLM.stream` mock.

### Before the patch

- Intent job: **0 updated, 0 skipped, 475 failed** in about 5 seconds.
- Every sampled item failed before an LLM response with `input.user.id: expected string, received undefined`.
- All 475 intent candidates had stored `experience_content.user_input` and `raw`, so they used the intended no-history fast path.

### After the patch

| Job | Total | Updated | Skipped | Failed | Duration |
|---|---:|---:|---:|---:|---:|
| Intent | 476 | **476** | 0 | **0** | 829,571 ms |
| Script, first pass | 146 | **122** | 1 | 23 | 511,069 ms |
| Script, bounded retry of remaining candidates | 24 | **1** | 1 | 22 | 153,243 ms |

Post-restart journal count for the previous malformed-user validation error: **0**.

A fresh detection scan now reports:

- Intent issues: **0** (previously 475)
- Script issues: **23** (19 invalid, 4 empty)

The remaining script records are a separate data/output-quality case, not this bug: 21 first-pass failures were `script-output-too-few-steps`, 2 were `script-output-junk`, and one record had no raw content. The second bounded pass repaired one additional record; the others again could not produce a valid multi-step trajectory, commonly because the original turn was only a greeting or had no meaningful actions. I stopped rather than persist fallback/invalid scripts.

### Runtime stability

- Synergy remained responsive throughout both full jobs.
- cgroup memory peak: **3,595,730,944 bytes** (~3.35 GiB).
- cgroup memory after completion: **1,776,816,128 bytes** (~1.65 GiB).
- No service restart, OOM, or malformed-user worker failure occurred during the jobs.

This confirms that omitting the absent source user and allowing `AgentCall.text()` to synthesize a valid temporary user fixes the production failure mode.

PR #880, opened 2026-07-25
fix: reload model-backed agents after config updates

### LINKED ISSUE #828: Config domain updates can skip runtime subsystem reloads
# Config domain updates can skip runtime subsystem reloads

## Summary

Updating a canonical config domain through `PATCH /config/domains/:domain` can persist the new configuration successfully while failing to reload the runtime subsystem that owns the changed field.

This was observed with a Channel account: Settings persisted `channel.clarus.accounts.<id>.enabled = true`, but the running Channel state remained unchanged until an explicit Channel reload.

## Expected behavior

When a domain update includes a runtime-owned field such as `channel`, the server should:

1. persist the validated domain update;
2. reload the Config cache;
3. preserve the fact that `channel` changed;
4. execute the corresponding runtime cascade (`Channel.reload()` in this case).

The updated account should become available without restarting the backend.

## Actual behavior

`Config.domainUpdate()` writes the domain file and then resets Config state. The route subsequently invokes `RuntimeReload.reload({ targets: ["config"] })`. `Config.reload()` computes its diff using the already-reset state, so its old and new snapshots can both contain the updated value. It reports `changedFields: []`, and `RuntimeReload.inferConfigCascades()` therefore does not run the subsystem reload.

The configuration file is correct, but the live runtime continues using stale state.

## Reproduction

1. Start Synergy with no enabled account for a Channel provider.
2. Enable the account in Settings.
3. Confirm the canonical Channels config contains `enabled: true`.
4. Query the Channel runtime status or invoke an account action.

The account may be absent from Channel runtime state, and the action fails as not connected. An explicit Channel reload makes the persisted account visible.

## Root cause

The config route already knows the request's top-level changed fields before persistence, but that information is not carried into the runtime reload result/cascade decision. Instead, cascade inference relies only on a second diff after `Config.domainUpdate()` has invalidated Config state.

Relevant ownership:

- `packages/synergy/src/config/config.ts` — domain persistence and Config state invalidation
- `packages/synergy/src/server/config-route.ts` — request-known changed fields and reload orchestration
- `packages/synergy/src/runtime/reload.ts` — changed-field reporting and subsystem cascade inference

## Suggested direction

Allow the config reload orchestration to retain explicit request-known changed fields and union them with fields detected by `Config.reload()`. Cascades should be inferred from that union.

This should be fixed in the shared Config/runtime reload contract rather than with a Channel- or Clarus-specific workaround.

## Acceptance criteria

- A domain update containing `channel` executes the Channel runtime cascade without backend restart.
- Runtime reload reports `channel` in `changedFields` even when Config state was invalidated during persistence.
- Diff-detected fields continue to work.
- Duplicate explicit/detected fields are de-duplicated.
- Existing provider, agent, plugin, MCP, LSP, formatter, watcher, Holos, command, and tool-registry cascade behavior does not regress.

EricSanchezok:
## Summary

- preserve validated model-role config changes after the config writer invalidates the cache
- rebuild model-backed agents after the config target completes, avoiding reload prerequisite re-entry
- cover the Models config route and runtime reload cascade with behavioral regressions

## Verification

- `bun test test/server/config-domain-route.test.ts test/runtime/reload.test.ts`
- `bun run test:changed`
- `bun run quality:quick`
- `bun run test:ci`
- pre-push checks: formatting, lint, full workspace typecheck, monorepo validation


yzxoi:
Reviewed the full merge diff and reproduced the runtime behavior. This PR addresses the real root cause reported in #828: `Config.domainUpdate()` invalidates Config state before `RuntimeReload.reload({ targets: ["config"] })`, so the second diff can lose the field that changed.

Before merging, there are three correctness issues to fix:

1. **Inferred cascades are flattened into `Promise.all()`.** For a `category` update, config infers both `provider` and `agent`, while `provider` also cascades to `agent`. I reproduced both an ordering inversion (`agent` ran before `provider` completed) and two concurrent `Agent.reload()` calls from one config update. Provider→agent, MCP→command, and plugin-derived reloads need to retain dependency ordering.

1. **The supplied fields are request keys, not actual changes.** `Object.keys(body.config)` reloads subsystems for no-op PATCHes, and misses keys removed by `mode: "replace-domain"`. The config owner should compute `Config.diff(oldConfig, newConfig)` after merge/replace semantics are applied.

1. **Field names alone cannot restore transition-dependent live apply.** After the writer resets the cache, `Config.reload()` sees the new config as both old and new. As a result, `execution.agentWorkers` still does not call `AgentTurn.resize()`, and `library.autonomy` has the same old/new-value problem. Runtime reload needs the complete config transition (`oldConfig`, `config`, `changedFields`), not only field names.

The focused PR tests and CI pass, but additional runtime probes reproduced all three cases above. I am preparing a follow-up on this PR that preserves the full config transition, computes real changed fields, and executes inferred cascades in dependency order, with regressions for no-op/replace-domain updates, worker-pool live apply, and duplicate agent reloads.

Related: #828

yzxoi:
Pushed the follow-up fix for the config reload issues described above.

What changed:

- capture the real pre-write and post-write config transition at the config owner boundary;
- skip true no-op updates and detect fields removed by `replace-domain`;
- use the writer-provided old config for transition-dependent live updates;
- execute inferred reload cascades sequentially so provider reload completes before agent reload and each cascade runs once.

Verification:

- focused config route/runtime regressions: 25 passed, 0 failed;
- package typecheck passed;
- `bun run quality:quick` passed;
- `bun run test:ci` passed all isolated shards;
- maintainability review passed with no blocking findings.

`bun run test:changed` also ran 2,876 tests in one process: 2,866 passed, 1 skipped, and 9 failed from existing order/global-state instability. The six Library API failures pass when rerun in isolation; the CI-equivalent isolated shard suite is green.

Related issue: #828.

yzxoi (approved):

PR #905, opened 2026-07-26
fix(workflows): recover incomplete reviewer completion

EricSanchezok:
## Summary

- detect BlueprintLoop and Light Loop reviewers that complete without a successful terminal review transition
- reuse the exact reviewer session with the execution session ID supplied and only approve/reject tools enabled
- bound recovery attempts, fail explicitly on exhaustion, and restore pending recovery during startup reconciliation
- persist refreshed Cortex controls when reusing a child session and document the workflow invariants

## Verification

- bun test test/session/continuation-kernel.test.ts test/session/blueprint-continuation.test.ts test/session/light-loop-continuation.test.ts test/session/recovery.test.ts test/cortex/manager.test.ts
- bun run test:ci
- bun run quality:quick


yzxoi:
I reviewed the recovery design and reproduced several edge cases against the exact PR head. The underlying issue is real and the continuation-policy-based recovery is the right direction, but I found four gaps that should be addressed before merge:

1. Recovery only handles reviewer tasks that end in `completed`. If a recovery turn ends in `error` or `cancelled`, both BlueprintLoop and Light Loop remain permanently stuck because neither runtime continuation nor startup recovery advances the workflow.
2. Recovery passes both an explicit reviewer `sessionID` and `reuseInterrupted: true`. `Cortex.prepare()` currently prefers any matching interrupted sibling over the explicit session, so the new task can be bound to the wrong child and then fail the recorded-reviewer access check.
3. When an ordinary, non-plugin Light Loop exhausts the recovery budget, `setTerminalStatus(..., "failed", error)` clears the workflow but does not persist or deliver the failure reason, so the loop silently disappears instead of failing explicitly.
4. Reusing a Cortex session updates `cortex.visibility` but not `completionNotice.silent`, leaving completion visibility inconsistent when reuse changes visible/hidden mode.

I reproduced the explicit-session misbinding and both terminal-state gaps with focused tests. I’m preparing a follow-up commit that fixes these cases and adds regression coverage, then I’ll push it to this PR branch.

yzxoi (approved):

PR #927, opened 2026-07-28
fix(session): harden abort settlement

EricSanchezok:
## Summary

- bound workspace file path enrichment with caller cancellation and an independent timeout
- hard-settle built-in/plugin and MCP tool execution when the combined session/tool timeout signal aborts
- preserve terminal tool errors against late non-cooperative results without retrying side-effecting calls
- document the settlement boundaries and add regressions for stalled VCS status and ignored abort signals

## Verification

- `bun test test/workspace-file/service.test.ts`
- `bun test test/session/tool-call-execution.test.ts`
- `bun test test/session/tool-scheduler.test.ts test/tool/observability.test.ts test/tool/file-search.test.ts`
- `bun run quality:quick`


yzxoi (commented):
I found one remaining settlement boundary issue: the configured tool timeout starts before `tool.execute.before`, but the new abort race only wraps the built-in/MCP body. A stalled `tool.execute.before` or `tool.execute.after` hook can therefore keep the call pending until the hook runtime's independent 120s timeout, even after session cancellation or a much shorter configured tool timeout. The same control-flow shape exists in both built-in/plugin and MCP paths.

I'll follow up on this branch with regression coverage and a fix that propagates/races the combined abort across the full hook → execution → hook lifecycle.

PR #945, opened 2026-07-29
fix(browser): retire idle Playwright resources

### LINKED ISSUE #944: fix(browser): retire idle Playwright pages and ownerless Chromium
## Summary

Headless Playwright resources remain live after browser tool activity stops. A session-owned page has no idle lifecycle, and the shared Chromium process remains running after the last owner context is released.

This can leave an otherwise idle browser consuming substantial CPU and memory inside the Synergy service cgroup.

## Production evidence

Observed on Linux with current `dev`:

- One session invoked `browser_navigation`, `browser_console`, and `browser_network`.
- The last successful navigation completed at approximately 13:37:54.
- No later browser tool activity occurred for that session.
- More than 30 minutes later, the page was still reported as `active`.
- Chromium renderer CPU remained around 110%.
- Chromium SwiftShader GPU CPU remained around 126%.
- The Chromium process group retained roughly 700 MiB RSS.
- No other session had invoked browser tools since this Chromium instance launched.

After sending the supported Browser `close` command:

- The owner state became `empty`.
- The renderer exited.
- Service cgroup memory dropped substantially.
- The shared Chromium process and SwiftShader GPU process remained alive.
- The GPU process continued consuming approximately 126% CPU despite there being no active page/context.

Terminating that now-ownerless Chromium process removed the CPU pressure while both the main and test Synergy HTTP endpoints remained healthy.

## Root cause

There are two lifecycle gaps:

1. `BrowserRuntime` reaps session resources only when the session is archived/deleted or a Cortex session reaches a terminal state. An ordinary primary-session turn finishing does not release an inactive browser page.
2. `PlaywrightBrowserDriver.releaseOwner()` closes the owner context but deliberately leaves the shared Chromium process running. There is no idle retirement after the last context is released.

Keeping browser state across a sequence of browser tools is useful, so closing after every tool call would be incorrect. The missing behavior is bounded idle suspension and restartable shared-browser retirement.

## Proposed behavior

- Refresh an owner idle deadline after every browser command.
- After a bounded period without browser commands, dispose/suspend that owner:
  - preserve descriptor/checkpoint/storage state;
  - close page, controller, transport, listeners, and BrowserContext;
  - allow a later `resume` to restore the page.
- Do not let idle timers keep the server process alive.
- Cancel idle timers during owner disposal and runtime shutdown.
- When the last Playwright owner context is released, close the shared Chromium browser.
- Ensure a later browser command can launch a fresh Chromium instance.
- Preserve multi-owner isolation: releasing one owner must not stop Chromium while another owner remains active.

## Acceptance criteria

- A browser page remains reusable during active command sequences.
- An owner is suspended after the configured idle window.
- A later `resume` works after idle suspension.
- Explicit close releases the owner context.
- Releasing the final owner closes Chromium.
- Releasing one of multiple owners does not close Chromium.
- A new command after retirement relaunches Chromium.
- Focused lifecycle tests cover timer reset, idle disposal, final-owner shutdown, multi-owner behavior, and relaunch.

## Priority

P1. On Linux, the observed no-page SwiftShader process continuously consumed more than one CPU core and materially increased the Synergy cgroup working set.

marzich:
## Summary

- suspend session-owned Browser resources after 10 minutes without a Browser command
- preserve the existing descriptor/checkpoint/storage state so a later `resume` can restore the page
- close shared Playwright Chromium when the final owner context is released
- deduplicate concurrent Chromium relaunches after retirement

Related to #944. This PR intentionally does not close the issue until production behavior is verified.

## Root cause

Browser pages were retained across ordinary primary-session turn completion because the runtime reaper only handled archive/delete and terminal Cortex transitions. Even after an explicit page close released the final owner context, `PlaywrightBrowserDriver` left the shared Chromium process running indefinitely.

On Linux this left an ownerless SwiftShader GPU process consuming approximately 126% CPU. Before explicit close, the retained renderer and GPU processes together consumed more than two CPU cores and roughly 700 MiB RSS for over 30 minutes after the final Browser tool call.

## Changes

| File | Change |
|------|--------|
| `packages/synergy/src/browser/command-service.ts` | Maintain an unref'ed per-owner idle deadline, refresh it after Browser commands, and route expiry through the existing ordered owner disposal path. |
| `packages/synergy/src/browser/playwright-driver.ts` | Retire Chromium after the final context closes and safely relaunch one shared browser for concurrent owners. |
| `packages/synergy/test/browser/tool-navigation.test.ts` | Cover idle disposal, deadline refresh, and explicit-close behavior. |
| `packages/synergy/test/browser/playwright-driver.test.ts` | Cover final-owner retirement, multi-owner isolation, and concurrent relaunch deduplication. |

## Behavior

- Browser command sequences continue to reuse the same live page.
- Every completed Browser command refreshes the 10-minute idle deadline.
- Idle expiry uses `BrowserRuntime.disposeSession()`, which captures recoverable state and suspends the page instead of deleting it.
- Explicit `close` remains terminal for the current page and does not create an idle timer.
- Releasing one owner does not affect other active owners.
- Releasing the final owner closes Chromium; later activity launches a fresh shared instance.

## Verification

- `bun test test/browser --timeout 30000`: 83 passed
- `bun run typecheck` in `packages/synergy`: passed
- root `bun run typecheck`: 11 packages passed
- pre-push Prettier check: passed
- pre-push Oxlint: passed with zero warnings
- pre-push Sherif workspace validation: passed
- `git diff --check`: passed


yzxoi:
@codex review

yzxoi (approved):

PR #947, opened 2026-07-29
fix(tools): bound arxiv PDF downloads

### LINKED ISSUE #946: bug(tools): arxiv_download timeout does not cover response body
## Summary

`arxiv_download` can remain `running` for many minutes after arXiv has returned HTTP response headers. The advertised 60-second operation timeout is cleared before the PDF response body is consumed, so a slow or stalled body stream leaves Cortex sessions blocked until the much longer outer tool timeout.

## Production evidence

Observed on Linux with five parallel scholar Cortex sessions during a literature survey:

- several dozen arXiv downloads were started across the children;
- multiple PDFs completed, including some taking 100-250 seconds;
- three calls remained `running` for 7-9 minutes while the Synergy process still held established TLS connections to arXiv/Fastly;
- the corresponding PDF output files did not exist yet;
- affected children could not settle their current assistant turn, so the parent waited for Cortex completion;
- calls that eventually received the complete body did finish, confirming that this was not a dead Synergy process.

Example stalled paper IDs included `2602.07755` and `2607.05609`.

## Root cause

In `packages/synergy/src/tool/arxiv.ts`:

1. a 60-second abort timer is created;
2. `fetch()` is awaited;
3. the timer is cleared as soon as `fetch()` resolves response headers;
4. only then is `response.arrayBuffer()` awaited.

`fetch()` resolving does not mean the response body has been downloaded. The body therefore has no arXiv-specific deadline. The generic tool timeout defaults to two hours, so the UI can show the call as running for a long time.

The implementation also buffers each complete PDF in one `ArrayBuffer` before writing it. Parallel scholar sessions can therefore produce avoidable external-memory peaks.

Latest `dev` improves settlement when an outer abort fires, but the arXiv body timeout and full-buffering behavior remain unchanged.

## Expected behavior

- The download deadline covers response headers, body transfer, and destination write.
- Session cancellation aborts the active body read promptly.
- The PDF is streamed to a temporary file instead of held entirely in memory.
- A configurable/bounded maximum PDF size is enforced.
- Timeout, cancellation, HTTP error, and oversized-body paths remove temporary files and preserve an existing destination.
- The temporary file is atomically renamed to the destination only after a complete successful download.

## Suggested tests

- response headers arrive, then the body stalls past the deadline;
- cancellation while reading the body;
- oversized `Content-Length` and chunked body exceeding the limit;
- partial download cleanup and preservation of an existing destination;
- successful streamed download and atomic replacement;
- parallel downloads do not require one full-PDF `ArrayBuffer` per call.

## Impact

This blocks Cortex completion and parent result propagation under realistic parallel research workloads. High concurrency amplifies the likelihood of a slow arXiv/Fastly stream, but reducing supported concurrency is not an adequate correctness fix; each network operation needs a complete lifecycle deadline and bounded memory behavior.

marzich:
## What does this PR do?

- keeps the arXiv download deadline active through response-body consumption and file completion;
- streams PDF chunks into an adjacent temporary file instead of materializing a full-response `ArrayBuffer`;
- enforces a 100 MiB bound using both `Content-Length` and streamed byte counts;
- atomically renames complete downloads and removes partial files on timeout, cancellation, HTTP failure, or size failure;
- adds deterministic regressions for a headers-then-stalled body, caller cancellation, chunked oversize responses, successful replacement, and temporary-file cleanup.

Closes #946.

## Why?

`arxiv_download` previously cleared its 60-second timer immediately after `fetch()` returned response headers. A slow or stalled PDF body could therefore leave a Cortex tool call running until the generic two-hour tool timeout. Parallel research sessions also held one full PDF `ArrayBuffer` per active download.

The destination is now changed only after a complete bounded download, so interrupted transfers do not corrupt or replace an existing paper.

## How was it tested?

- `bun test test/tool/arxiv-download.test.ts` from `packages/synergy` (4 passed)
- `bun test test/tool/timeout.test.ts test/tool/search-guard.test.ts` from `packages/synergy` (10 passed)
- `bun run quality:quick` from the repository root
- pre-push format, lint, and monorepo typecheck hooks

## Checklist

- [x] New tests live under `packages/synergy/test/tool`
- [x] No server API, generated SDK, localization, auth, or documentation contract changed
- [x] No secrets, local runtime data, or unrelated cleanup included


yzxoi:
PR #947 的整体修复方向是对的：下载 deadline 现在覆盖 response body，PDF 按块写入临时文件，并在完整下载后提交。但 review 和本地复现发现 3 个需要一起修复的问题：

1. **最终提交阶段没有继续受 timeout/cancellation 约束。** 当前代码在 `rename()` 前清除了 timer，并且没有在 rename 后重新检查 signal。延迟 rename 的复现中，20ms timeout 或 caller abort 后，调用仍在约 114ms 返回成功并替换目标。
2. **`overwrite=true` 改变了目标文件的元数据和链接语义。** 默认权限创建的临时文件经 rename 后会让现有 `0600` 文件变成 `0644`；symlink 路径会被普通文件取代，原 symlink target 保持旧内容。覆盖实现需要明确并保留既有文件语义，而不是无条件更换 inode。
3. **临时文件清理失败被静默吞掉。** `rm(...).catch(() => {})` 会让调用只报告 timeout/原始错误，但实际留下 `.synergy-arxiv-*.tmp` partial file。清理失败应与主错误一起向上报告，并有回归测试。

我会在这个 PR 分支上补充针对上述场景的回归测试和修复，包括安全的提交边界、元数据/链接处理与 cleanup error 聚合，然后重新跑聚焦测试、typecheck 和质量检查。

yzxoi (approved):

Issue #951, opened 2026-07-29
bug: reconnect generation race leaves completed session messages missing until hard refresh


marzich:
## Summary

A completed session can persist and submit its full result successfully while the browser omits a contiguous tail of assistant/tool messages until a hard refresh.

This is a follow-up to #509 and #331. Durable message/part reload after reconnect exists, but the page can miss the reconnect generation that is supposed to trigger it.

## Production evidence

Affected session: `ses_05338ceceffeL1tgzWSbTOZlJ7` on a Tailscale HTTP deployment.

- `GET /session/:id/message/page?limit=200` returned all 30 messages with `hasMore=false`.
- All omitted assistant messages were persisted with `visible=true`, the correct `rootID`, complete text/tool parts, and terminal tool states.
- Clarus submission completed successfully; this was not a task/runtime failure.
- The existing browser omitted the synthesis/write/submission tail.
- A clean headless browser loaded the same URL and rendered all 53 timeline items, including every omitted message.

Therefore storage, message paging, root grouping, and static rendering are healthy. The stale state is confined to the already-open browser's live sync store.

## Root cause

The session page reacts to:

```ts
[params.id, sdk.connected()]
```

and calls `sync.session.sync()` when the socket reports connected.

`sync.session.sync()` decides whether to force a durable message/part snapshot by comparing `sessionReconnectVersions` with `globalSync.reconnectVersion()`.

However, `globalSync.reconnectVersion` is incremented by a separate reactive effect after `globalSDK.connected()` becomes true. This creates an ordering race:

1. WebSocket reconnects and `connected=true`.
2. Session-page effect runs first and calls `sync()` with the old reconnect version.
3. The sync plan sees the session as current and short-circuits, so no message snapshot is fetched.
4. Global-sync effect then increments `reconnectVersion`.
5. The session page does not observe that value, so it never retries.

Any unsequenced `message.part.*` events missed during the disconnect remain absent from the browser store even though durable storage is complete. A hard refresh fixes the display because initial message paging loads the authoritative parts.

## Expected behavior

The actively viewed session must run one authoritative message/part sync for every reconnect generation, regardless of reactive effect ordering.

## Proposed fix

- Expose the current reconnect generation through the scope sync context.
- Include that generation in the session page's sync trigger dependencies.
- Keep the existing generation comparison/deduplication so the second invocation is cheap unless a new reconnect generation actually requires a snapshot.
[...]

PR #952, opened 2026-07-29
fix(app): resync session parts after reconnect

marzich:
## Summary

- make the active session react to reconnect generation changes, not only the WebSocket `connected` boolean
- force the existing authoritative message/part snapshot path to run after every reconnect generation
- preserve the existing inflight deduplication and freshness checks
- add a focused regression test that locks reconnect generation into the session sync watch key

Relates to #951. This PR intentionally does not close the issue until it has been validated against a live reconnect.

## Root cause

The page and global sync contexts react independently when the WebSocket reconnects. The page could call `sync.session.sync()` before global sync incremented `reconnectVersion`; that call saw the previous generation and short-circuited. The later generation increment was not observed by the page, leaving any missed unsequenced `message.part.*` events absent until a hard refresh.

The affected production session was complete in durable storage: its page endpoint returned all 30 messages and all omitted parts. A clean browser rendered all 53 timeline items, while the already-open browser omitted the tail. This isolates the failure to reconnect convergence in the live frontend store.

## Changes

| File | Change |
|------|--------|
| `packages/app/src/context/sync.tsx` | expose the reactive reconnect generation through the scope sync context |
| `packages/app/src/pages/session.tsx` | include reconnect generation in the active session sync trigger |
| `packages/app/src/context/session-sync-plan.ts` | define the canonical session sync watch key |
| `packages/app/test/context/session-sync-plan.test.ts` | verify generation advances change the watch key |

## Verification

- `bun test test/context/session-sync-plan.test.ts test/context/session-part-snapshot-freshness.test.ts` (18 passed)
- `bun run typecheck` (Turbo, 11 packages passed)
- `bun run --cwd packages/app build` passed
- Prettier check passed for all changed files
- `git diff --check` passed


yzxoi:
Addressed the reconnect-generation race from this review.

- Split the global recovery-start generation from a per-Scope recovery-completed generation. The completed generation publishes only after replay/full resync succeeds, and it is suppressed for failed, stale, or released Scope recoveries.
- Delayed the session page's durable message/part refresh until that completed Scope generation advances instead of triggering it at the raw WebSocket reconnect.
- Made in-flight session sync generation-aware: newer or stronger requests queue behind insufficient pending work, and only an accepted latest message snapshot advances the captured generation watermark.
- Preserved permission and volatile refresh behavior and documented the resulting reconnect contract.

Validation:

- `bun test test/context/scope-reconnect-recovery.test.ts test/context/session-sync-plan.test.ts` — 24 passed
- `bun run --cwd packages/app test` — passed
- `bun run --cwd packages/app typecheck` — passed
- `bun run --cwd packages/app build` — passed

yzxoi (approved):

PR #953, opened 2026-07-29
fix(lattice): allow pristine workflow re-enable

EricSanchezok:
## Summary

- cancel pristine Lattice runs when the workflow is disabled so Lattice can be enabled again immediately
- preserve pause and explicit resume semantics after durable Lattice work has started
- recover legacy pristine runs left paused by the previous disable behavior
- serialize lifecycle decisions under the Lattice lock and document the updated contract

## Verification

- `bun test test/session/workflow.test.ts test/lattice/`
- `bun run quality:quick`
- independent maintainability review


yzxoi:
Addressed the lifecycle edge cases in a follow-up commit:

- include pending in-memory model calls when deciding whether a Run is pristine;
- persist the initial goal prompt effect atomically with the Run record;
- roll back only the exact newly created pristine Run when Session workflow projection fails, while preserving an existing active Run;
- add regression coverage for pending-call disable/re-enable, projection retry, existing-Run preservation, and the interrupted-create persistence window.

Validated with:

- `bun test test/lattice/run-service.test.ts test/session/workflow.test.ts` (62 passed)
- `bun test test/lattice/controller.test.ts test/lattice/store.test.ts test/lattice/policy.test.ts test/server/lattice-route.test.ts` (31 passed)
- `bun run quality:quick`

yzxoi (approved):

PR #966, opened 2026-07-30
fix(files): defer and degrade workspace search

### LINKED ISSUE #965: Prompt mount eagerly scans large workspaces and file-search timeout becomes a global frontend error
## Summary

Mounting the session prompt immediately starts an empty-query `GET /workspace/files/search`, even when the user has not opened the `@file` picker. On large workspaces, path enrichment can exceed its 20-second timeout. The backend returns 500, and the frontend surfaces the response through the global error boundary as `UnknownError: Internal server error`.

This was observed immediately after enabling a Clarus account, but Clarus itself connected successfully. The channel config PATCH returned 200 and the Clarus runtime connected; the actual failing route was workspace file search.

## Evidence

Current `dev`: `6d88486f8`

Workspace involved:

- approximately 21 GB
- approximately 45,000 files

Server evidence:

```text
GET /workspace/files/search status=500 duration=20227
error=The operation timed out.

PATCH /config/domains/channels status=200 duration=40
channelType=clarus ... channel connected

GET /workspace/files/search status=500 duration=20211
error=The operation timed out.
```

Frontend symptom:

```text
Error: Unknown error
Caused by:
UnknownError
Internal server error
```

## Root cause

`useFilteredList()` creates its resource immediately. The prompt's `@file` provider passes an async item source, so mounting `PromptInput` calls `searchFilesAndDirectories("")` before the picker is used.

The backend then:

1. builds or refreshes the workspace file index;
2. selects the first page for the empty query;
3. enriches every result with filesystem metadata and Git status;
4. races enrichment against a 20-second abort signal;
5. lets the abort rejection escape, producing HTTP 500.

A settings/config refresh can remount or refocus the session UI, making this unrelated file-search failure look like the setting change broke the frontend.

## Expected behavior

- Prompt mount should not scan workspace files until the `@file` picker is actually activated.
- A file-search enrichment timeout should return a bounded/truncated response or a route-local recoverable error, not crash the page through the global error boundary.
- Enabling a channel should not appear to break unrelated session pages.

## Proposed minimal fix

1. Lazily activate the prompt's async `@file` source on first picker input rather than on hook creation.
2. Make file-result enrichment bounded and timeout-tolerant; reuse a single Git status snapshot instead of resolving Git status independently for every result.
3. Add focused tests for lazy async-list activation and timeout-safe file search.

No concurrency reduction or workspace-size restriction is proposed.

marzich:
## Summary

- defer the prompt `@file` async source until the picker is actually activated
- reuse one Git status snapshot while enriching file-search results
- degrade internal enrichment timeouts to basic truncated results instead of HTTP 500
- preserve caller-driven abort behavior

Fixes #965.

## Changes

| Area | Change |
|---|---|
| `packages/ui` | Add an opt-in `deferInitialLoad` mode to `useFilteredList`, with a browser-runtime regression test |
| `packages/app` | Enable deferred loading for prompt `@file` suggestions |
| `packages/synergy` | Bound file metadata enrichment, reuse `statusMap()`, and return basic results on internal timeout |
| tests | Cover deferred activation, external cancellation, and timeout degradation |

## Verification

- `bun test test/workspace-file/service.test.ts` (24 passed)
- `bun test --conditions=browser ./test/hooks/use-filtered-list.test.tsx` (1 passed)
- `bun run typecheck` (11 packages passed)
- focused Prettier check passed
- focused oxlint passed
- `bun run --cwd packages/app build` passed (6,842 modules)
- production service restarted successfully; Tailscale HTTP endpoint returned 200
- file search in the observed ~21 GB workspace returned HTTP 200 in 0.56 s
- Clarus channel reconnected successfully after restart

## Scope

The PR does not include local sandbox environment overrides, configuration changes, or audit worktrees.

### DIFF
--- a/docs/architecture/workspace-and-files.md
+++ b/docs/architecture/workspace-and-files.md
@@ -43,7 +43,7 @@ Search has three independent modes:
 
 File-index scans consume and retain at most 50,000 complete paths, deduplicate retained paths, preserve results collected before a subprocess output limit or scan timeout, and mark the search response as truncated whenever a bound is reached. A workspace that is too large for one bounded index scan therefore returns partial file matches instead of failing the route with a 500 response or retaining an output-sized object graph indefinitely.
 
-File-result path enrichment has its own finite timeout and races caller cancellation. A stalled shared VCS status lookup therefore cannot keep a canceled file search pending indefinitely.
+File-result path enrichment has its own finite timeout and fails open to the complete basic path results when optional metadata is unavailable. Caller cancellation still propagates, while `truncated` remains reserved for an incomplete index or result page.
 
 The classic debug file search is lazy and reuses this same bounded project index. Starting a project Scope does not launch a second fire-and-forget repository scan.
 
--- a/packages/app/src/components/prompt-input/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input/prompt-input.tsx
@@ -1222,6 +1222,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
       const paths = await files.searchFilesAndDirectories(query)
       return paths.map((path): AtOption => ({ type: "file", path, display: path }))
     },
+    deferInitialLoad: true,
     key: atKey,
     filterKeys: ["display"],
     onSelect: handleAtSelect,
--- a/packages/app/src/components/prompt-input/submit.ts
+++ b/packages/app/src/components/prompt-input/submit.ts
@@ -29,6 +29,7 @@ import {
   formatSessionPreview,
   formatSessionReference,
   inlineLength,
+  inlineText,
   SESSION_PREVIEW_MAX_MESSAGES,
 } from "./content"
 import { setCursorPosition } from "./editor-dom"
@@ -924,7 +925,7 @@ export function usePromptSubmit(input: PromptSubmitInput) {
     const textPart = {
       id: Identifier.ascending("part"),
       type: "text" as const,
-      text,
+      text: inlineText(currentPrompt),
     }
     const requestParts = [
       textPart,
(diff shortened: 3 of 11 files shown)


yzxoi:
Thanks for the fix — the issue and the two-layer direction are valid: the prompt should not trigger a workspace search before the user invokes `@`, and optional file-result enrichment must not turn a path search into an HTTP 500.

I found a few issues that should be addressed before merging:

1. `useFilteredList.onInput()` activates the resource before updating the filter. A first non-empty input such as `foo` therefore starts both `""` and `"foo"` searches. The empty request is not canceled and can race the real query on a cold workspace index.
2. The `statusMap()` refactor is not needed. `WorkspaceFileStatusCache` already shares one in-flight status build across per-item `statusForPath()` calls, so this does not reduce the expensive Git work. It also introduces two regressions: empty result pages still run the full status lookup, and non-timeout status failures now escape the existing `maybeNode()` fail-open boundary as HTTP 500s.
3. The new browser-only UI test classification uses `path.join()` while matching POSIX path literals, so the test runs in the wrong Solid condition on Windows.

I’m going to push a follow-up that:

- keeps the opt-in deferred initial load;
- updates the filter before activating the resource and adds a first-non-empty-input regression test;
- restores the original `maybeNode()`/status-cache flow and adds only the minimal timeout fail-open wrapper around optional enrichment;
- keeps caller cancellation propagating;
- preserves `truncated` for actual result/index truncation rather than missing optional metadata;
- fixes the UI test runner to use POSIX logical paths;
- adds coverage for ordinary enrichment failure and timeout fallback.

yzxoi (approved):

PR #977, opened 2026-07-31
fix(app): align quick-switcher recommendations

### LINKED ISSUE #976: fix(app): keep quick-switcher recommendations consistent with Models settings
## Summary

The Models settings page can show a model as enabled for the quick switcher while the session top-bar picker does not list it and cannot find it through search. The provider and model are available; the mismatch is entirely client-side.

Observed with `openai-codex/gpt-5.6-luna`:

- `/provider` returns Luna and reports `openai-codex` connected.
- Settings shows the Luna quick-switch switch enabled.
- The top-bar picker only contains Sol/Terra and cannot search Luna.
- The persisted `quick_switcher.models` contains no Luna override.

## Root cause

The settings model manager and the runtime local-model context independently implement the same recommendation heuristic. Their input ordering differs:

- Settings sorts selectable models by display name before choosing the newest reasoning/free/multimodal model.
- Runtime evaluates provider catalog/object order.
- Models with the same `release_date` therefore produce different winners. For the GPT-5.6 family, Settings chooses Luna while runtime chooses Sol.

Because Settings believes Luna is recommended by default, the checked switch produces no explicit `state: add` preference. No config patch is sent, so the top-bar list remains unchanged.

## Expected behavior

Settings and the top-bar picker must derive the same default recommendation set regardless of provider/model input order. Equal release dates need deterministic tie-breaking.

## Proposed fix

- Move recommendation selection into a shared model-catalog helper.
- Use stable tie-breaking: release date descending, then display name and model ID ascending.
- Use the shared helper in both Settings and the runtime quick switcher.
- Add a regression test with Sol, Terra, and Luna sharing a release date and reversed input order.

## Impact

This affects any provider that publishes multiple same-day models, not only OpenAI Codex. It makes a checked Settings control misleading and prevents users from selecting a model they have apparently enabled.

marzich:
Fixes #976

## Summary
- centralize quick-switcher recommendation selection in the provider model catalog
- use deterministic release-date/name/id tie-breaking across Settings and the session model picker
- cover equal-release-date models with an order-independent regression test

## Root cause
Settings sorted the connected catalog by model name before selecting the newest model, while the runtime picker used provider insertion order. For GPT-5.6 models sharing a release date, Settings displayed Luna as enabled but the session picker recommended Sol, so Luna was absent despite the provider being connected.

## Validation
- `bun run typecheck`
- `cd packages/app && bun test test/components/provider/model-catalog.test.ts`
- restarted the local dev service and confirmed HTTP 200
- verified against live `/provider` data that `openai-codex/gpt-5.6-luna` is available and returned by the shared recommendation helper


marzich:
Added commit `38c61529a` to cover the second consistency failure reported during live validation.

The picker no longer treats recent history as configured quick-switch membership. A recent-only model is shown only in the `Recent` group; provider groups are generated from the resolved recommendation + explicit preference set.

Local validation after rebuilding and restarting:
- repository typecheck: 11/11 packages passed
- focused tests: 6 passed
- production frontend build passed
- Tailscale HTTP endpoint returned 200 with the new asset
- live OpenAI Codex provider group resolved to Luna/Sol/Terra; GPT-5.5 resolved only to Recent

yzxoi (approved):
LGTM. Reviewed with focus on the recent/quick-switcher separation and provider recommendation behavior. No blocking issues found for this change. The quick-switcher duplicate display behavior in Recent vs configured section is intentional and acceptable given the UX intent. Approve.

PR #978, opened 2026-07-31
fix: defer provider catalog runtime schema load

yzxoi:
## Why
- Prevent provider catalog process startup crash when `./models` has runtime schema/import issues during module load.
- Keep `./models-schemas` as the compile-time dependency while moving `./models` loading to runtime.

## What changed
- Lazy-load `ModelsDev` from `./models` in `packages/synergy/src/provider/catalog.ts`.
- Guard models.dev refresh listener registration with error handling around the lazy import.
- Add regression test `packages/synergy/test/provider/catalog-resilient-load.test.ts` to ensure `ProviderCatalog` module import still succeeds when the `models` module export is missing.

## Tests
- `bun test test/provider/catalog-resilient-load.test.ts`


yzxoi:
@codex review

PR #999, opened 2026-07-31
feat: deliver Feishu outbound text as CardKit markdown cards

yzxoi:
## Summary

- Ordinary Feishu/Lark outbound text (terminal answers, proactive pushes, and non-streaming fallbacks) was sent as `msg_type: "text"`, so Feishu displayed raw markdown syntax (`**bold**`, `# heading`) instead of rendering it.
- Terminal answers now ship as a CardKit card with a single `markdown` element — the same rendering already used by streaming cards, response cards, and question cards. Sanitization reuses the existing `sanitizeFeishuCardMarkdown`.
- Robust by default: falls back to plain text when the serialized card exceeds the 30 KiB card budget or the card API fails; media attachments keep their native message types; streaming card behavior is unchanged.
- New `responseFormat` setting (`"text"` | `"markdown"`, default `"markdown"`) at both provider level (`channel.feishu`) and account level for chats that prefer raw text. Docs updated in `docs/reference/configuration.md` and `docs/product/connections.md`; config JSON schema and SDK types regenerated.

## Test plan

- `cd packages/synergy && bun test test/channel/feishu-provider.test.ts` — 46 pass, including 6 new markdown-reply tests (card by default, push path, account/provider `text` override, card-API failure fallback, oversize fallback, config parse defaults)
- `cd packages/synergy && bun test test/channel/` — 333 pass
- `cd packages/synergy && bun test test/config/` — 123 pass
- `bun run quality:quick` — 11/11 tasks green (format, lint, typecheck, monorepo/package checks, localization)
- `./script/generate.ts` run twice; generated OpenAPI/SDK output stable

Note: `bun run test:changed` reports one pre-existing failure in `test/tool/arxiv-download.test.ts` ("reports temporary-file cleanup failures") that also fails on the clean base tree and is unrelated to this change. `test/channel/command.test.ts` passes in isolation and in the full channel suite.


yzxoi:
## Review: URL regression in the default markdown reply path

The new `sendFeishuMarkdownCard` reuses `sanitizeFeishuCardMarkdown`, but that sanitizer was designed for **card UI text** (response/question cards), not for free-form LLM answers. Before this PR it was only used by `response-card.ts` and `question-card.ts`; the streaming card — the path that actually renders LLM answers — uses its own `normalizeMarkdown`, which only blanks empty content and keeps everything else.

The sanitizer strips:

- `[text](url)` → `text` (link label only, URL gone)
- bare `https?://...` → **deleted entirely**

So after this PR, every ordinary answer (push/reply/non-streaming fallback, all defaulting to `responseFormat: "markdown"`) silently drops URLs: \"see https://example.com/docs\" becomes \"see\". Streaming cards (default on) keep URLs, so the two modes also render inconsistently. The new tests never asserted URL handling, which is why this slipped through.

Fix direction: align the markdown reply card with the streaming card — keep the content as-is (only blank-content fallback), leave `sanitizeFeishuCardMarkdown` to the UI cards. I will implement this and push the fix to this branch.

yzxoi:
## Fix pushed: keep URLs in markdown reply cards

`c117d2637` changes the markdown reply card to use the same `normalizeMarkdown` as the streaming card (only substitutes a blank for empty content) instead of `sanitizeFeishuCardMarkdown`. Bare URLs and `[text](url)` links now survive in ordinary replies; the sanitizer stays exclusive to response/question card UI text.

**Verified:**
- `bun test test/channel/feishu-provider.test.ts` — 46 pass (default card test now asserts URLs and links are preserved verbatim)
- `bun test test/channel/` — 333 pass
- `bun run typecheck` (synergy package) — clean
- Prettier — clean

Note: the earlier pre-push hook failure was an unrelated dependency-install issue (`ghostty-web` not materialized after a certificate error during `bun install`); a reinstall resolved it.

PR #1008, opened 2026-08-01
feat(provider): single connection resolver for profile semantics

yzxoi:
## Summary

Introduces `packages/synergy/src/provider/connection.ts` as the **single owner** for mapping a connection ID to its runtime semantics, addressing the design review findings on #990 (see [review comment](https://github.com/SII-Holos/synergy/pull/990#issuecomment-5148950376)).

Currently each consumer resolves connection semantics independently and drifts:

- `catalog.ts refresh()` gates profile/config completion on `ScopeContext.tryScope()` — a bare `refresh(connectionID)` outside a scope silently returns `bundled` without refreshing
- the same "connection → profile" resolution is re-implemented in `provider.ts` init, `auth.ts` retargetHook, `usage-service.ts`, `setup.ts`, `transform.ts`, with different invalid-profile handling per module
- `profileID` is optional on `Info`; forgetting to thread it silently degrades to direct lookup
- snapshot eviction protection in `persistSnapshots` is computed from the **canonical profile set** even though snapshot keys are connection-scoped — with many connections near `MAX_SNAPSHOT_ENTRIES` (100), low-activity connections' live snapshots are evicted as LRU garbage
- `composeProviderSpec` (provider init / `resolveModelOptions` / `configureWorkerProvider`) has three subtly different merge orders; worker re-runs profile hooks with `provider: undefined` while the main process passes catalog metadata

## What this PR adds

1. **`ProviderConnection.resolveConnection(connectionID, config)`** — the only resolution path: `{ profile, catalogSourceID, baseURL, env }` with explicit `unknown_profile` failure instead of silent fallback.
2. **`ProviderConnection.composeProviderSpec(...)`** — the single composition path with a fixed layer order: catalog projection → profile runtime options → connection overrides. Used by init, getSDK, import probes, and (via a serialized definition) the agent worker, so all merge in the same order.
3. **`ProviderConnection.ConnectionStateManager`** — per-connection snapshot lifecycle whose eviction protection is computed from the **active connection set** (fixes the B1 finding), with register/unregister/invalidate semantics for future catalog/auth/scope reload wiring.

Fully self-contained: no existing call sites are changed yet. Consumers (catalog, provider init, auth, usage, setup, transform) are migrated to this module in follow-up PRs.

## Verification

- `bun test test/provider/connection.test.ts` — 14 pass / 0 fail (resolution matrix, composition precedence, eviction protection)
- `bunx tsc --noEmit` on the package — no errors in changed files
- `bunx oxlint` and `bunx prettier --check` — clean

## Follow-ups

- Migrate `catalog.ts`, `provider.ts`, `auth.ts`, `usage-service.ts`, `setup.ts`, `transform.ts` to use this module (removes the scope-gated refresh fallback and the per-module drift)
- Worker plan serialization round-trip consistency test (main-process plan → worker replay via the same `composeProviderSpec`)
- Make `refresh()` fail loudly outside a scope instead of returning bundled


yzxoi:
@codex review

PR #1009, opened 2026-08-01
feat: render markdown images in Feishu reply cards

yzxoi:
## Summary

Feishu CardKit markdown only accepts `alt` — an `image_key` from the image upload API, not an HTTP URL. Any reply containing `alt` therefore failed card validation and the whole message degraded to raw plain text.

This PR materializes markdown images before building the card:

1. Extracts `alt` references from the reply text
2. Downloads the URL through the existing outbound-media pipeline (30s timeout, HTTP status check)
3. Uploads it to Feishu via `im/v1/images` (reusing `uploadImage`)
4. Replaces the URL with the returned `image_key` so the card renders the image

**Graceful degradation** (single image failure never fails the card):
- Download failure / non-`image/*` content type → keep as a plain link `[alt](url)`
- Non-HTTP destinations (`data:`, `file:`, relative) → keep only alt text
- No image syntax → no download or upload happens

## Files

- `packages/synergy/src/channel/provider/feishu/send-card.ts` — `materializeMarkdownImages` runs before card build
- `packages/synergy/src/channel/provider/feishu/outbound-media.ts` — export `uploadImageFromUrl`
- `packages/synergy/test/channel/feishu-markdown-image.test.ts` — 7 new tests (replace, degrade, multiple, no-op)
- `docs/product/connections.md` — outbound delivery behavior documented

## Test plan

- `bun test test/channel/feishu-markdown-image.test.ts` — 7 pass
- `bun test test/channel/feishu-provider.test.ts test/channel/feishu-message.test.ts test/channel/feishu-question-card.test.ts test/channel/feishu-response-card.test.ts test/channel/feishu-streaming-card.test.ts` — 81 pass
- `bun test test/channel/` — 340 pass
- `bun run typecheck` (synergy) — clean
- `bun run lint` (root oxlint) — 0 errors/warnings
- Prettier — clean


yzxoi:
## Fix pushed: structure-aware image matcher (`f439a9f05`)

Addresses the review finding that the pure-regex `MARKDOWN_IMAGE_PATTERN` mis-matched `` in non-rendering contexts (inline code, fenced code blocks, escaped positions), causing:

- unnecessary outbound downloads/uploads (and a wider SSRF surface)
- code examples being rewritten into links/alt text

**Fix:** a lightweight structure-aware scanner (`collectNonRenderedSpans`) tracks three non-rendered contexts before matching:

| Context | Handling |
|---|---|
| Escaped punctuation (`\!`) | skipped as a 2-char literal span |
| Inline code spans (``...````, ```...```) | skipped entirely |
| Fenced code blocks (``` / `~~`~, ≥3 chars, info string, ≤3-space indent) | skipped until matching close fence |

Only image syntax **outside** those spans is materialized; code-context examples are kept verbatim with zero requests.

**New tests (4):** inline code, fenced block, escaped, and a real-image + code-example mixed case verifying only the real image is fetched/uploaded and code content is untouched.

**Verified:** 11/11 image tests pass, 59 feishu provider/streaming tests pass, typecheck + prettier clean.

Issue #1011, opened 2026-08-01
feat(channel): isolate Feishu Sessions by thread


marzich:
## Problem

Feishu group conversations are currently scoped by one of `group`, `group_sender`, `group_topic`, or `group_topic_sender`. None of these modes supports the desired interaction model for a project bot:

- every accepted non-thread `@bot` request starts independent work;
- the bot creates a Feishu thread under that current request;
- only messages inside the same Feishu thread continue the same Synergy Session;
- independent requests in the same group can execute in parallel.

The current `group_topic` mode uses `root_id ?? thread_id` and falls back to `chatId`. This causes top-level requests to share one group Session and treats ordinary replies as Session continuity. In addition, the Feishu provider awaits the full conversation handler through a queue keyed by `chatId`, so separate Sessions in one group would still execute serially.

## Required behavior

Introduce an opt-in Feishu group Session mode with the following semantics:

1. **Non-thread request creates a Session**
- When an accepted group message has no `thread_id`, create a new Session keyed by the current message's `message_id`.
- Do not use `root_id` to select an existing Session.
- Preserve `root_id`, `parent_id`, and quoted content only as input context and historical lookup signals.

1. **Reply in a new thread**
- The first response to a non-thread request must use `reply_in_thread: true`, creating a Feishu thread under the current request.
- The response must remain anchored to the current request rather than being delivered under an older quoted message.

1. **Thread is the only continuity signal**
- Messages with the same `thread_id` must reuse one Session regardless of sender.
- Persist a restart-safe `thread_id → Session` association when the bot creates a thread for an existing non-thread request.
- If Atlas is first mentioned inside an existing thread, create one Session for that thread and reuse it thereafter.

1. **Parallelize independent work**
- Different non-thread requests and different threads in the same chat must be able to execute concurrently.
- Messages within one thread/Session must remain serialized.
- Replace or narrow the current full-lifecycle `perChatQueue`; an optional account/Scope concurrency limit may protect resources but must not serialize the entire chat.

1. **Do not debounce across Session boundaries**
- The current `chatId + senderId` debounce key must not merge separate top-level requests.
- Non-thread requests must be isolated by their own `message_id`; thread messages may be grouped only within the same thread.

### Independent top-level requests

```text
A: @Atlas investigate PR #123
B: @Atlas prepare the meeting summary
C: @Atlas check the pre environment
```

Expected:

- three Sessions;
- three Feishu threads created under the respective requests;
- all three may execute in parallel.

[...]

PR #1023, opened 2026-08-01
fix(channel): persist busy Feishu messages

### LINKED ISSUE #202: bug: queued inbox messages not processed after session execution ends (race between drainReady and enqueueUser)
## User report

> @chaos 现在执行过程中再发一条新消息，新消息会先暂存。但等执行结束后，这个暂存的新消息并不会被执行

Source: group:oc_46d261becbf063b7e0b667667d210297; message:om_x100b6b467cdf0ca8b36c7d22bfa2bb8

## Triage

- Type: bug_report
- Related area: session

## Deduplication

Related but distinct from #193 (Session inbox can auto-drain mid-run). Issue #193 describes the inbox item disappearing and being inserted mid-run unexpectedly. This report is a different failure mode: the queued inbox item is neither drained mid-run nor processed after the run finishes — it simply remains in the inbox indefinitely.

### Session input route (`packages/synergy/src/server/session.ts`)

The `submitInput` function enqueues user input to the inbox when the session is running:

```ts
async function submitInput(input: InvokeInput): Promise<SessionInbox.InputResult> {
  const messageID = input.messageID ?? Identifier.ascending("message")
  const next = { ...input, messageID }
  if (SessionManager.isRunning(input.sessionID)) {
    const item = await SessionInbox.enqueueUser(next)
    return { status: "queued", item }
  }
  SessionInvoke.invoke(next).catch(...)
  return { status: "started", messageID }
}
```

### Session invoke loop (`packages/synergy/src/session/invoke.ts`)

After each inner turn, the loop drains ready inbox items:

```ts
// Inner loop finished — drain queued user input and agent updates
const readyItems = await SessionInbox.drainReady(sessionID)
const readyResult = await materializeInboxItems(sessionID, readyItems)
if (readyResult.needsReply) continue outer
```

### Race condition

There is a race between `drainReady` reading/removing inbox items and `enqueueUser` writing a new item:

1. `loop()` inner turn completes → calls `drainReady(sessionID)` → reads and removes all queued items from storage → finds nothing → outer loop breaks
2. Meanwhile (between drainReady reading and the outer loop break), a new message arrives → `submitInput` → `SessionManager.isRunning` → true → `SessionInbox.enqueueUser(next)` → writes to inbox
3. `loop()` function returns → `SessionManager.run()`'s finally block only checks `runtime.mailbox.length`, not persistent inbox items → runtime is unregistered
4. The inbox item remains in persistent storage but nobody processes it

The `SessionManager.run()` finally block (`packages/synergy/src/session/manager.ts`):
```ts
finally {
    const runtime = getRuntime(session.id)
    if (runtime && !runtime.abort && runtime.mailbox.length === 0) {
        unregisterRuntime(session.id)
    }
}
```

No subsequent check is made for leftover inbox items.

## Expected behavior

When a session is running and a user submits another message:

1. The message should be queued in the inbox with `deliveryTarget: "after_turn"`.
2. Once the current agent run reaches its final response and becomes idle, the queued inbox item should be drained and submitted as the next turn.
3. If no new messages arrive after the run, the queued message should still be processed.

## Actual behavior

1. During execution, the user's new message is correctly queued to the inbox.
2. After execution ends, the queued message remains in the inbox and is never processed.
3. The message is only processed if the user sends yet another message (which triggers a new `invoke` → `loop()` → `drainReady` picks up both).

## Proposed fix

The fix should ensure that when a session invoke finishes, any remaining inbox items are checked and processed. Options:

**Option A**: In `SessionManager.run()`'s finally block, after checking the runtime mailbox, also check for persistent inbox items and process them:

```ts
finally {
    const runtime = getRuntime(session.id)
    if (runtime && !runtime.abort) {
        const inboxItems = await SessionInbox.list(session.id)
        if (inboxItems.length > 0) {
            // Trigger a new invoke to process queued inbox items
            await run(session.id, () => processMailbox(session.id))
        } else if (runtime.mailbox.length === 0) {
            unregisterRuntime(session.id)
        }
    }
}
```

**Option B**: In `submitInput`, after enqueuing to the inbox, atomically check if the session is still running. If the session became idle during the enqueue, immediately trigger a new invoke.

**Option C**: In `loop()`, after the outer loop decides to break, re-check the inbox one more time before returning.

Suggested labels: `bug`, `area/session`

marzich:
## Problem

PR #1013 added thread-scoped Feishu Sessions, but the provider lane still awaited the complete conversation execution. A second message in the same thread could therefore remain blocked behind a long-running generation; if it reached a busy Session near loop completion, the message could also be left in the persistent inbox without a reliable wake or restart recovery path.

This is a reliability follow-up to #1011 and PR #1013. It also closes the remaining Feishu-specific form of the queued-inbox race described in the already-closed #202; there is no separate open Issue for this report.

## Summary

- Separate Channel conversation acceptance from background execution so the Feishu per-thread lane waits only for durable acceptance.
- Persist busy-session messages in the existing `SessionInbox` with stable provider delivery identity and reply correlation metadata.
- Preserve same-thread ordering and different-thread parallelism without introducing a provider-owned durable queue.
- Recover runnable inbox tasks on startup through the ordinary `SessionDrive` loop.
- Make inbound attachments durable before acceptance and skip attachment reconstruction on duplicate replay.
- Track accepted background executions during bounded Feishu account drain.

## Implementation

- `ChannelHost.conversations.receive()` now returns an acceptance result plus a separately tracked execution Promise.
- `ChannelConversationAcceptance` reserves the Session loop lease for idle work, queues behind existing runnable work, and deduplicates direct and queued delivery using the same stable identity.
- `SessionManager` exposes one terminal lease operation so externally reserved leases have a single release owner.
- `SessionInbox` discovers runnable Sessions for startup recovery and retains its existing peek/materialize/commit ownership.
- Feishu retains the existing `group_thread` queue key, thread binding, debounce isolation, and account drain timeout.

## Related work

- Follow-up to #1011
- Follow-up to #1013
- Related to #202

## Verification

- `bun test test/channel/` — 375 passed, 0 failed
- `bun test test/session/restart-while-queued.test.ts test/session/inbox.test.ts test/session/recovery.test.ts test/session/manager.test.ts test/session/working.test.ts` — 90 passed, 0 failed
- `bun run --cwd packages/synergy typecheck` — passed
- Second-pass maintainability review — passed with no blocking findings

## Compatibility

- No provider-side durable queue or persisted-state schema migration is introduced.
- Existing Feishu thread routing and legacy scope behavior remain unchanged.
- Duplicate delivery keys remain idempotent after inbox materialization and process restart.
- No production deployment or Atlas runtime/configuration change was performed.


yzxoi:
I reviewed the merged behavior of this PR against the latest `dev`. The durable busy-session queue is the right direction, but I found three correctness gaps that should be fixed before merge:

1. **Idle-session acceptance is not durable yet.** `ChannelConversationAcceptance.accept()` returns `accepted: true` after acquiring an in-memory lease, while `input.execute()` is deferred to a later microtask. At that point there may be neither an inbox delivery record nor a materialized user message. A process exit in that window loses an already-accepted Feishu message, and a concurrent replay can be accepted twice because `findDelivery()` still sees nothing.

1. **Same-thread commands can overtake a running generation.** The Feishu per-thread lane now waits only for acceptance and detaches `execution`. A following `/new` or `/reset` therefore reaches `ChannelCommand.execute()` while the previous generation still owns the Session lease. `/new` can archive and unregister that Session while its generation is still writing messages and sending a reply. Ordinary messages pass through lease arbitration, but commands currently run before that boundary.

1. **Startup inbox recovery includes archived Sessions.** `SessionInbox.listRunnableSessions()` scans inbox files without reading Session info or excluding `time.archived`. Archiving does not clear queued task items, so restart recovery can wake an archived conversation and execute delayed work against the old Channel endpoint.

I’m preparing a follow-up on this PR branch with behavioral coverage for these three cases and the focused fixes. I’ll run the relevant Channel/Session tests and report the exact verification in the commit/PR update.

yzxoi (approved):

PR #1025, opened 2026-08-01
fix(synergy-link): share canonical Holos accounts

EricSanchezok:
## Summary

- read the active Holos identity from the canonical multi-account store
- write and clear Link credentials without losing inactive accounts
- retain legacy credential fallback only when the canonical store is unavailable
- migrate legacy Link credentials into the canonical account store

## Testing

- bun run --cwd packages/synergy-link test
- bun run --cwd packages/synergy-link typecheck
- bun run --cwd packages/synergy-link build
- bun run quality:quick


yzxoi:
Thanks for moving Synergy Link to the canonical Holos multi-account store. We agree with **Option A**: Synergy and standalone Synergy Link should share one global active Holos identity.

Before merging, the current branch needs the following corrections:

1. **Use the canonical Synergy home resolution.** `SynergyLinkHolosAuth.synergyRoot()` ignores `SYNERGY_HOME`, while Synergy's canonical account owner prioritizes it. With a custom or isolated home, the two runtimes currently read and write different `holos-accounts.json` files.
2. **Do not fall back to legacy credentials when the canonical store is malformed or unreadable.** Legacy fallback should only be considered when the canonical file is genuinely absent (`ENOENT`). Parse, schema, permission, and other I/O failures must remain visible; otherwise stale credentials can silently become active again.
3. **Serialize cross-process read-modify-write operations.** Atomic rename prevents partial files but does not prevent lost updates when Synergy and Synergy Link update the shared account collection concurrently. The canonical account owner needs a cross-process lock or equivalent consistency mechanism.
4. **Reconcile with current `dev`.** The PR currently conflicts in the Holos auth implementation and documentation. Conflict resolution must preserve the current canonical `synergy.d` Holos configuration, endpoint resolution/normalization, and endpoint-environment validation.

I will push a follow-up to this PR implementing these corrections while retaining Option A semantics: Link login changes the global active account, and Link logout removes the global active account.

PR #1028, opened 2026-08-01
feat: harden Synergy Link v2 end to end

### LINKED ISSUE #982: bug: Synergy Link Holos client and auth/config use different endpoint sources of truth
## Summary

Synergy Link currently has inconsistent Holos endpoint definitions:

- `packages/synergy-link/src/holos/client.ts` hardcodes `www.holosai.io` for token exchange and WebSocket connection.
- `packages/synergy-link/src/holos/auth.ts` defines `api.holosai.io` as `HOLOS_API_HOST`/`HOLOS_WS_URL` and writes those values into Synergy config.

The two hostnames resolve to different infrastructure. Even when both currently serve compatible token responses, this drift makes environment selection and Agent Tunnel routing difficult to reason about and can create multi-instance routing failures.

## Observed evidence

For Synergy Link 3.0.9:

```text
client.ts: https://www.holosai.io + wss://www.holosai.io
auth.ts:   https://api.holosai.io + wss://api.holosai.io
```

DNS resolution also yielded different address sets for the two names.

## Expected behavior

- Token exchange, WebSocket connection, generated config, doctor output, and portal links should use a single explicit endpoint configuration model.
- The runtime should not duplicate or hardcode competing API/WS hosts.
- Environment overrides should be supported and validated together.

## Acceptance criteria

- One source of truth defines Holos API, WebSocket, and portal endpoints.
- `client.ts` consumes the same endpoint configuration used by auth/config.
- Login, doctor, reconnect, and server startup show the effective sanitized endpoint.
- API and WS endpoint overrides cannot silently target inconsistent environments.
- Tests verify default endpoints and custom environment endpoints.
- Existing issue #527 is cross-linked because endpoint drift may affect WebSocket routing, but this issue remains focused on configuration ownership and consistency.

### LINKED ISSUE #983: bug: expired Synergy Link collaboration sessions remain opened in the sender registry and cannot be closed cleanly
## Summary

A remote Synergy Link collaboration session expires after its idle timeout (currently 10 minutes by default), but the sending Synergy connection registry can continue to report the session as `opened` and reuse its old Session ID.

Subsequent RPC calls are rejected by the Link with:

```text
session_invalid: No active collaboration session.
```

An attempted `connect close` can also time out because it targets the already-expired Session ID. Re-running `connect open` may reuse the stale local record rather than establishing a fresh session.

## Observed sequence

1. A Link session is opened successfully.
2. No activity occurs for more than the remote idle timeout.
3. The remote logs:

```text
session.expired.idle_timeout
runtime.session.expired
```

1. The sender still reports the connection as `opened` with the old Session ID.
2. A remote command is sent with that Session ID.
3. The Link returns `session_invalid`.
4. `connect close` also fails or times out.

## Expected behavior

- Session expiry should converge on both endpoints.
- The sender should remove or invalidate the local connection record when the remote reports `session_invalid` or expiry.
- A subsequent `connect open` should establish a new Session ID without requiring a new Link ID or process restart.
- `connect close` should be idempotent when the remote session is already gone.

## Acceptance criteria

- Remote idle expiry causes the sender connection to become closed/expired within a bounded interval.
- Receiving `session_invalid` invalidates the cached Session ID immediately.
- The next `connect open` creates a new session.
- Closing an already-expired session succeeds as an idempotent no-op or returns a non-timeout terminal state.
- Tests cover expiry, stale RPC, stale close, and reopen.
- The error shown to the user distinguishes session expiry from Link/Agent offline status.

### LINKED ISSUE #984: security: bind Synergy Link process handles and output to the creating caller and collaboration session
## Summary

The Synergy Link process registry is shared across collaboration sessions. Process records are addressed by process ID, but are not bound to the creating `sessionID`, caller Agent ID, or caller owner user ID.

A later valid collaboration session can potentially list, read logs from, write to, signal, kill, or remove processes created by another caller or an earlier session.

## Security impact

This is a cross-session authorization boundary violation. A caller with valid Link access should not gain control over another caller's processes or output merely by obtaining a process ID or invoking `process list`.

Potentially affected operations include:

- list
- poll/status
- log/output
- stdin/write
- send-keys
- signal/kill
- clear/remove

## Expected behavior

Every process record must be owned by the identity that created it:

```text
sessionID + callerAgentID + callerOwnerUserID
```

All subsequent process operations must enforce that ownership.

## Acceptance criteria

- Process records persist creator session and caller identity.
- `process list` only returns processes visible to the current authorized caller/session.
- Every read/control operation verifies ownership before exposing output or mutating a process.
- Session close/expiry/revocation terminates and removes all processes owned by that session, including descendants/process groups.
- Finished output is not visible across sessions.
- Process IDs remain unguessable, but authorization does not rely on secrecy of the ID.
- Tests include two callers/two sessions and verify that each cannot list, read, write, signal, kill, or remove the other's processes.
- Security logs record denied cross-session access without exposing process output.

## Priority

P0/security. The current remote execution surface is arbitrary shell execution; process ownership isolation is a required defense boundary, not only lifecycle cleanup.

### LINKED ISSUE #985: bug: synergy-link doctor treats the expected standalone local_owner=none state as a failure
## Summary

In standalone mode, Synergy Link intentionally owns and maintains its own Holos connection. No local Synergy owner is required. However, `synergy-link doctor` unconditionally treats a missing local owner as an issue.

## Observed behavior

A healthy standalone Link reported:

```text
✔ mode — standalone
✘ local_owner — No local owner declared
✔ auth — agent <id> (shared)
✔ service — pid <pid>
✔ connection — connected
✔ holos_secret — Credentials accepted by Holos

✘ Issues found
```

The Link was concurrently verified through a successful Collaboration Session and remote RPC.

## Expected behavior

Doctor checks should be mode-aware:

- `standalone`: no local owner is expected; report `not applicable` or pass.
- `managed`: require a valid, non-expired local owner and report failures precisely.

## Acceptance criteria

- A healthy standalone Link exits successfully even when no local owner exists.
- Output explains that local ownership is not applicable in standalone mode.
- Managed mode continues to fail if its required owner is missing, stale, or inconsistent.
- Tests cover standalone, healthy managed, missing managed owner, and expired managed owner.
- Doctor's exit code and summary agree with the individual checks.

### LINKED ISSUE #986: bug: Holos presence caches a recovered target as offline for 24 hours and hides the delivery failure reason
## Summary

After a Holos `ws_failed` response, Synergy marks the target Agent offline in an in-memory Presence map. The offline entry remains valid for approximately 24 hours. Subsequent sends to that Agent are rejected locally without another network delivery attempt, even after the target Link fixes its credentials and successfully reconnects to Holos.

The user-facing remote transport also discards the concrete send failure reason and reports only:

```text
Synergy Link request <request-id> was not delivered
```

Together, these behaviors make a transient or configuration-induced delivery failure persist until the sending Synergy process is restarted.

## Reproduction

1. Start a Synergy Link whose displayed Agent ID does not match its Secret-derived Holos identity, or otherwise cause the target to return `TARGET_OFFLINE`.
2. From Synergy, call `connect open` for that target Agent.
3. Observe `request was not delivered`.
4. Correct the remote credentials and confirm:
- WebSocket open;
- `synergy-link reconnect` succeeds;
- the token identity is correct;
- the target Link is connected.
5. Retry from the same Synergy process.

Observed: delivery still fails immediately because the target remains cached offline. Restarting the sending Synergy process clears the Presence map and the same `connect open` succeeds immediately.

## Root cause

The sender handles `ws_failed` by marking the target Agent offline. `HolosProvider.send` then short-circuits whenever Presence reports offline. Reconnecting the sender or target does not invalidate that target-specific cached state, and no bounded probe is attempted before the long TTL expires.

Separately, `HolosSynergyLinkTransport` throws a generic `was not delivered` error for all `{sent:false}` results, dropping distinctions such as `not_connected`, `offline`, and `delivery_failed`/`TARGET_OFFLINE`.

## Expected behavior

- An offline observation must not suppress delivery attempts for 24 hours after a target may have recovered.
- Presence should be refreshed or invalidated on relevant reconnect/online events.
- A short TTL plus bounded reprobe/backoff should replace permanent local short-circuiting.
- Transport errors should preserve a safe, actionable reason and server failure code.

## Acceptance criteria

- A target that reconnects can receive requests without restarting the sender.
- A failed target is reprobed after a bounded interval.
- Sender reconnect clears or revalidates stale target Presence entries.
- Inbound target activity marks the target online.
- Errors distinguish at least:
- sender not connected;
- locally cached offline;
- server returned `TARGET_OFFLINE`;
- delivery failed for another reason.
- Tests cover failure → remote recovery → successful retry in the same process.
- Existing issue #527 should be cross-linked, but this issue remains separate: #527 concerns competing WebSocket reconnect cycles rather than target Presence TTL and error propagation.

### LINKED ISSUE #987: bug(security): reject Synergy Link credentials when the imported Agent ID does not match the ws_token identity
## Summary

`synergy-link login` accepts an operator-provided Agent ID and Agent Secret as independent values. It verifies that the Secret is accepted by Holos, but it does not verify that the returned WebSocket token belongs to the provided Agent ID.

This allows the CLI and status output to report one Agent ID while Holos registers the tunnel under another Agent ID derived from the Secret. The Link then appears logged in and connected, but requests sent to the displayed Agent ID fail as offline/not delivered.

## Observed behavior

During a real standalone Link deployment:

```text
Configured Agent ID: 29fbc816-b5a8-49d0-920f-59ef7756956f
ws_token agent_id: 770a9e2c-f563-4ea8-a0ad-60aed5b300ff
```

The CLI reported:

```text
Logged in         yes
Agent ID          29fbc816-b5a8-49d0-920f-59ef7756956f
Holos             connected
```

However, Holos routed the WebSocket under `770a...`, so collaboration requests addressed to `29fbc...` returned `request was not delivered`.

The mismatch was independently confirmed against both:

- `https://www.holosai.io/api/v1/holos/agent_tunnel/ws_token`
- `https://api.holosai.io/api/v1/holos/agent_tunnel/ws_token`

Both tokens contained the old Agent ID.

## Root cause

The imported `agentID` is persisted and used for local display/logging, while the server-side tunnel identity comes from the Secret-derived token. Login validates Secret acceptance but does not compare the token's `agent_id` claim to the imported Agent ID.

## Expected behavior

When importing existing credentials, Synergy Link should:

1. Exchange the Secret for a short-lived WebSocket token.
2. Decode or introspect the authenticated Agent identity.
3. Compare the authenticated Agent ID with the operator-provided Agent ID.
4. Reject the login without writing credentials when they differ.
5. Display an actionable error that identifies an ID/Secret mismatch without printing either Secret or token.

Preferably, the Agent ID should be derived from the authenticated Secret rather than accepted as an independent source of truth.

## Acceptance criteria

- A matching Agent ID and Secret are saved successfully.
- A mismatched pair is rejected before auth/config files are modified.
- Existing credentials are not overwritten on mismatch.
- `whoami`, `status`, runtime logs, and the actual Holos tunnel identity cannot disagree.
- Tests cover matching credentials, mismatched credentials, malformed/opaque tokens, and token exchange failure.
- No Secret or complete WebSocket token is written to logs or errors.

## Security impact

This is an identity-binding and diagnostic integrity issue. It can cause requests to be routed to the wrong Agent identity and can mislead operators into debugging networking or Link state instead of credentials.

### LINKED ISSUE #988: bug: synergy-link status presents stale state.json values as live when the control socket is unreachable
## Summary

When the Synergy Link control socket cannot be reached, `synergy-link status` can fall back to persisted `state.json` and display `Service running` and `Holos connected` without clearly indicating that these values are only a last-known snapshot.

This can mislead operators during recovery: the recorded PID may be stale, the process executable may have been deleted, or the runtime may be unable to service control requests even though status appears healthy.

## Observed behavior

```text
Service           running
PID               <pid>
Holos             connected
```

while:

```text
Timed out connecting to control socket at .../control.sock
```

In one migration, the recorded process existed but `/proc/<pid>/exe` pointed to a deleted binary, and the control socket could not complete a reconnect request.

## Expected behavior

Status must distinguish authoritative live runtime data from persisted last-known state.

## Acceptance criteria

- When the control socket responds, output is marked live.
- When the control socket is missing/unreachable/timed out, output clearly reports runtime status as unknown or stale.
- Snapshot-derived fields are labeled `last known` and include the snapshot timestamp/age.
- A stale PID is checked against process existence and, where possible, executable identity/start time.
- `status` exits non-zero when live status was requested but cannot be obtained.
- `--json` exposes a source field such as `live | snapshot` and the control socket error.
- Tests cover missing socket, stale socket, dead PID, reused PID, and healthy live runtime.

### LINKED ISSUE #992: bug: remote Bash yield at the transport deadline loses the process handle
## Summary

Remote Synergy Link Bash requests can deterministically time out when `yieldSeconds` equals the transport timeout. The current sender transport abandons every request after 30 seconds, while the Link host waits the full requested yield duration before returning an auto-background result.

With `yieldSeconds: 30`, a command that runs longer than 30 seconds cannot return its background `processId` before the sender has already deleted the pending request.

The remote command is not cancelled, so it may continue running without a process handle available to the caller.

## Reproduction

1. Open a valid Synergy Link collaboration session.
2. Run a remote command that takes longer than 30 seconds with:

```json
{
  "yieldSeconds": 30
}
```

1. Observe that the sender fails after approximately 30 seconds:

```text
Timed out waiting for Synergy Link response <request-id>.
```

1. The host may continue executing the process, but the caller never receives the `processId` that would have been returned by auto-backgrounding.

## Root cause

The Holos transport uses a fixed 30-second response deadline. The Link process registry independently waits `yieldSeconds * 1000` before constructing an auto-background response. Network delivery and serialization necessarily happen after that wait.

The timeout path only removes the local pending request and rejects the promise. The protocol has no cancellation request, no `AbortSignal`, and no request-ID-to-process-ID recovery operation.

## Expected behavior

- A valid `yieldSeconds` value must always leave enough time for the host response to arrive before the transport deadline.
- If the sender times out after the remote process starts, the process must either be cancelled or recoverable by request ID.
- Long-running commands must not become untracked orphan work merely because the result crossed a response deadline.

## Acceptance criteria

- Reject or clamp `yieldSeconds` values that are not safely below the transport timeout.
- Prefer immediate background execution for commands expected to exceed the response budget.
- Add a safety margin between host yield and sender timeout.
- Add protocol-level cancellation or idempotent request-ID-to-process-ID lookup.
- Distinguish `request timed out before dispatch` from `remote process started but result was lost`.
- Tests cover the exact 30-second boundary, late responses, process recovery, cancellation, and duplicate retries.

## Evidence

Relevant paths:

- `packages/synergy/src/remote/holos-transport.ts`
- `packages/synergy-link/src/exec/process-registry.ts`
- `packages/synergy-link-protocol/src/bash.ts`
- `packages/synergy-link-protocol/src/client.ts`

### LINKED ISSUE #993: bug: connect list, status, and open diverge for raw or stale Link sessions
## Summary

`connect list`, `connect status`, and `connect open` can report mutually contradictory Synergy Link state because they use different local views and do not verify the remote collaboration session.

A raw session opened with `linkID + targetAgentID` can produce all three results at the same time:

```text
connect list   -> No active Synergy Link sessions
connect status -> status: opened
connect open   -> Connection is already open
```

Remote Bash requests can nevertheless time out or fail because the host session is no longer valid.

## Root cause

- `connect list` filters every in-memory session through `SynergyLinkTargetStore.findByLocator(...)`, hiding raw sessions that do not have a matching persisted target.
- `connect status` reads only the in-process session map and performs no heartbeat.
- `connect open` immediately returns `already open` when the cached record says `opened`; it does not probe or heartbeat the host.
- `lastUsedAt` is updated before the remote request succeeds, so failed requests make stale sessions appear recently active.
- A failed or timed-out `connect close` does not clear or downgrade the cached session.

This overlaps with, but is broader than, #983. That issue covers remote idle expiry and stale cached Session IDs. This issue covers contradictory command surfaces, hidden raw sessions, false freshness, and lack of remote verification.

## Expected behavior

All connect surfaces should derive from a coherent session model and clearly distinguish:

- locally cached;
- remotely verified alive;
- unknown after timeout;
- expired/invalid;
- closed.

## Acceptance criteria

- `connect list` includes raw sessions owned by the current local agent, with an explicit `unregistered` marker.
- `connect status` reports both local cache state and last remotely verified state.
- `connect open` heartbeats an existing cached session before returning `already open`.
- `session_invalid` clears the local record immediately.
- Transport timeout changes the session to `unknown`, not `opened`.
- Split `lastAttemptAt` from `lastVerifiedAt`; failed attempts do not refresh verified freshness.
- Provide a local-only force-clear operation for unrecoverable stale records.
- Tests cover raw sessions, timeout, stale close, heartbeat failure, reopen, and list/status/open consistency.

## Evidence

Relevant paths:

- `packages/synergy/src/tool/connect.ts`
- `packages/synergy/src/tool/synergy-link-execution.ts`
- `packages/synergy/src/synergy-link/target-store.ts`

Related: #983.

### LINKED ISSUE #994: bug: Synergy Link targets cannot be relinked and idle does not indicate reachability
## Summary

A persisted Synergy Link target cannot be repaired after its Agent identity or Link ID changes. The target patch contract allows only `name`, `enabled`, and `allowedAgents`; it cannot atomically update `targetAgentID + linkID`.

At the same time, `list_targets` can display the stale target as `idle` merely because a local Holos execution client exists, even when that target's locator is obsolete or unreachable.

This leaves operators with a target that looks available but routes connection attempts to the wrong Agent/Link and cannot be relinked in place.

## Reproduction

1. Create and approve a target for Agent A / Link A.
2. Rotate or replace the Link so the valid locator becomes Agent B / Link B.
3. Open Agent B / Link B using a raw locator.
4. Observe:
- the old target may still appear `idle`;
- opening by target ID still routes to Agent A / Link A;
- the raw session is not associated with the persisted target;
- target patch cannot update either locator field.

## Root cause

- `SynergyLinkTarget.PatchInput` does not accept `targetAgentID` or `linkID`.
- Target availability is computed as `connected` only for a matching cached session; otherwise it is `idle` whenever any local Synergy Link client exists.
- `idle` therefore represents local client presence, not target reachability or a successful recent probe.

## Expected behavior

- Operators should be able to atomically relink a target after identity or endpoint rotation.
- Relink must validate the new locator before committing it.
- Availability must reflect target-specific evidence, not merely the existence of a local transport client.

## Acceptance criteria

- Add an explicit relink operation accepting `targetAgentID + linkID` together.
- Probe the new locator and verify host identity before committing.
- Invalidate or reconcile sessions associated with the old locator.
- Preserve authorization, name, and allowed-agent policy during relink.
- Track target-specific states such as `unknown`, `unreachable`, `reachable`, and `connected`.
- Include last probe result and timestamp in target views.
- Do not report `idle` as an availability signal when no target-specific probe has succeeded.
- Tests cover Agent rotation, Link rotation, failed relink rollback, session reconciliation, and stale target rendering.

## Evidence

Relevant paths:

- `packages/synergy/src/synergy-link/types.ts`
- `packages/synergy/src/synergy-link/target-runtime.ts`
- `packages/synergy/src/synergy-link/target-service.ts`
- `packages/synergy/src/tool/connect.ts`

### LINKED ISSUE #1004: feat: bash 经 synergy-link 远端执行时工具展示未标明远端来源,本地/远端无法区分;应在工具渲染中根据 metadata.backend 显示"通过 synergy-link 执行"标识
## Triage

- Type: feature_request
- Area: area/frontend
- Summary: bash 经 synergy-link 远端执行时工具展示未标明远端来源,本地/远端无法区分;应在工具渲染中根据 metadata.backend 显示"通过 synergy-link 执行"标识

### LINKED ISSUE #1021: Cross-host Agent Tunnel identity takeover silently redirects existing Synergy Link targets
## Summary

A Synergy Link instance can be started on a different physical host using shared Link state and the same Holos Agent credential. Holos currently treats the newest WebSocket authenticated as that Agent ID as authoritative, unconditionally replaces the previous active tunnel, and routes subsequent Agent RPCs to the new socket.

Because Synergy targets are currently bound to `targetAgentID + linkID`, not to a verified physical-host identity, an existing target can silently begin executing on the wrong server while its local target ID, label, Link ID, Agent ID, and connected/opened status all remain internally consistent.

This was reproduced during an Atlas recovery on 2026-08-01. The incorrectly started Link process has been removed; this issue documents the underlying cross-host identity-takeover gap.

## Impact

A host that can read a valid Agent Secret and persisted Link state can become the active execution endpoint for that Agent from another machine. Existing clients do not need to change their target records: once the Holos Agent Tunnel route changes, subsequent RPCs follow the new tunnel.

This creates a dangerous failure mode for deployments using shared filesystems, copied Link homes, restored backups, container images, or otherwise portable credentials/state:

- RPCs can execute on an unexpected physical host.
- Target labels and IDs can continue to look correct.
- A normal reconnect can be indistinguishable from a cross-host takeover.
- The old tunnel is displaced with the same replacement mechanism used for expected reconnects.
- Stale collaboration state can obscure the original tunnel failure and make the wrong-host tunnel appear to be a successful recovery.

## Confirmed incident topology

The two Link deployments were intentionally on different servers:

- Atlas target: `target_26b6341a-bb6a-4772-aec4-f148c0057213`
- Link: `link_31a610c6-8fda-471e-bb62-e606e5d0ef77`
- Agent: `8c673455-826a-461a-9ee1-a08c4a8116a8`
- `Clarus 内测群助手` target: `target_52f7323c-8c11-4972-94cf-5f075f9a2af0`
- Link: `link_4fda00ed-239e-4eba-b385-235a4d5c0975`

Both servers mounted:

```text
/inspire/hdd/project/multi-agent/czxs253130466
```

The healthy Clarus helper target returned hostname:

```text
synergy-chaos--e8962f2a1d55-x6avdzsyy5
```

The shared Atlas `instances/atlas-qz` path was visible on that host even though it was not the Atlas server.

## Reproduction / causal timeline

1. The existing Atlas collaboration session was stale: the sender still reported the target/session as connected/opened, while short RPCs and `connect close` timed out. This is the separate stale-session behavior tracked in Synergy issue #983.
2. From the healthy `Clarus 内测群助手` target, the shared Atlas instance directory was visible.
3. That shared-path visibility was incorrectly interpreted as evidence that execution was occurring on the Atlas host.
4. `atlas-qz/scripts/serve.sh` was started on the Clarus helper server.
5. The process loaded the Atlas-specific HOME, `SYNERGY_LINK_HOME`, state, and credential.
6. Startup identified the process as Atlas Agent `8c673455-826a-461a-9ee1-a08c4a8116a8` and Link `link_31a610c6-8fda-471e-bb62-e606e5d0ef77`.
7. Holos logs showed repeated socket replacement/reconnect behavior for that Agent identity.
8. After the sender cleared its stale collaboration cache with `POST /holos/reconnect` and freshly opened the existing Atlas target, an Atlas-target RPC returned:

```text
synergy-chaos--e8962f2a1d55-x6avdzsyy5
```

1. This proved that the unchanged Atlas target was now routed through the new Agent tunnel on the wrong host.
2. The mistakenly started tmux window/process was removed, and the wrong host was verified to have no remaining `atlas-qz` process.

The resulting chain was:

```text
shared persistent state visible on both servers
  -> Atlas credentials/state loaded on the wrong server
  -> second Link process authenticates as the correct Atlas Agent
  -> Holos replaces the active tunnel for that Agent ID
  -> existing Synergy Atlas target still resolves by the same Agent ID
  -> fresh RPCs execute on the wrong physical host
```

## Root cause

This requires gaps at three layers.

### 1. Holos Agent Tunnel: last writer wins by Agent ID without host binding

Investigated at `SII-Holos/holos-server` `origin/main`:

```text
800b41672f3483fdb9e1d5577048ebf7a0bdc8f8
```

The Agent Tunnel WebSocket identity is derived from JWT claims including `agent_id`, `owner_user_id`, and `scope="agent_ws"`:

- `holos/modules/agent_tunnel/security.py:109-146`
- `holos/modules/agent_tunnel/api.py:974-988`

On a new connection for an existing Agent ID:

- `_prepare_agent_ws_connection` calls `crud.replace_active_session`.
- Existing active sessions are marked `connection_status="replaced"` with `close_reason="replaced_by_new_session"`.
- The new session is inserted as active.
- The old socket is closed with code `4002`, reason `replaced_by_new_session`.
- The in-memory routing map is overwritten by `agent_id`.

Evidence:

- `holos/modules/agent_tunnel/api.py:336`
- `holos/modules/agent_tunnel/crud.py:217-249`
- `holos/modules/agent_tunnel/api.py:1004-1028`
- `holos/modules/agent_tunnel/service.py:40-91`
- `holos/modules/agent_tunnel/models.py:83-86`

`remote_ip` and `gateway_instance_id` are recorded, but no expected host identity, machine fingerprint, or cross-host lease is checked. A valid latest connection wins regardless of physical origin.

### 2. synergy-link: portable credential and state, no cross-host ownership

Investigated in Synergy monorepo `origin/dev`:

```text
103571b84c6e0a56435ac23f04accad59d8cfe8e
```

The Holos identity is loaded from the shared auth file as `agentId + agentSecret`, with no host binding:

- `packages/synergy-link/src/holos/auth.ts:32-37,109-121`
- `packages/synergy-link/src/holos/client.ts:26-34,147-158`

The Link runtime directly adopts the persisted `linkID` and `hostSessionID` from `state.json`:

- `packages/synergy-link/src/state/store.ts:104-125`
- `packages/synergy-link/src/runtime.ts:53-58`

Hostname is capability metadata only, not an authorization or lease key:

- `packages/synergy-link/src/platform.ts:40,247-250`
- `packages/synergy-link/src/types.ts:7-13`

Local process protection does not provide cross-host ownership:

- the guard is a short control-socket ping;
- the socket is unlinked before listen;
- PID reconciliation uses the local PID namespace;
- there is no server-visible cross-host lease.

Evidence:

- `packages/synergy-link/src/service.ts:37,54-76,220-222`
- `packages/synergy-link/src/control/server.ts:20,47`
- `packages/synergy-link/src/service/local.ts:7-13`

In this incident, the Clarus helper Link and the mistakenly started Atlas Link used different instance homes/control sockets, so they could coexist normally on the same physical host.

### 3. Synergy target: Agent/Link binding but no verified host pin

Investigated in the same Synergy `origin/dev` commit:

```text
103571b84c6e0a56435ac23f04accad59d8cfe8e
```

Target persistence and lookup bind by `targetAgentID + linkID`:

- `packages/synergy/src/synergy-link/types.ts:41-66`
- `packages/synergy/src/synergy-link/target-store.ts:42-73`

`connect open` routes using those values. Host information is recorded only after a successful connection as an advisory probe:

- `packages/synergy/src/tool/connect.ts:123,131-182,213-229`
- `packages/synergy/src/synergy-link/target-store.ts:83-107`

The probe guard only verifies that the remote self-reported `linkID` matches the target. It does not verify a stable expected host identity or reject a changed hostname/host instance.

### Not target-label drift

The two local target names did not merge. Target names are local labels and are not the routing authority.

### Not an Agent ID/Secret mismatch

The wrong-host process authenticated as the correct Atlas Agent identity. An invalid or mismatched Secret would not have replaced the Atlas Agent tunnel.

### Direct cause: duplicate correct Agent identity on the wrong host

The wrong host started a second Link process with the correct Atlas Agent Secret and persisted Link state. Holos made it the active tunnel for that Agent ID.

### Separate interacting cause: stale collaboration session (#983)

The stale collaboration session did not move execution to another host. It obscured the dead/original execution path and required a sender reconnect/fresh open. The fresh open then followed the newly replaced Agent tunnel, making the wrong-host route look like successful recovery.

The stale-session issue and cross-host takeover should remain separate because they have different owners, root causes, and acceptance criteria.

## Expected behavior

A reconnect from the same installed Link instance should be distinguishable from a connection presenting the same Agent credential from a different physical/runtime host instance.

An existing target must not silently begin executing on another host merely because that host can read copied/shared credential and state files.

### Holos server — primary trust-boundary owner

1. Add a stable `hostInstanceID` to Agent Tunnel registration, backed by an installation-specific key rather than hostname/IP alone.
2. Persist the expected host binding or current lease owner for an Agent/Link identity.
3. When the same Agent ID connects from a different `hostInstanceID`, reject, quarantine, or require an explicit takeover/rebind flow instead of silently replacing the current tunnel.
4. Emit a high-signal structured takeover event containing old/new host instance, remote IP, gateway, and session IDs.
5. Distinguish same-host reconnect from cross-host identity takeover with separate server events and close reasons.
6. Add tests for two clients using the same Agent credential from different host instances.

### synergy-link

1. Generate and store a non-shared, installation-bound `hostInstanceID` and key.
2. Record state ownership and fail closed when shared/copied state is started from another host without explicit rebind.
3. Acquire/renew a server-visible Agent/Link ownership lease.
4. Store host identity with PID state; do not interpret a PID without its host namespace.
5. Replace best-effort socket probing with a reliable local exclusive lock for same-host duplicate starts.
6. Clearly display local host identity, server-verified owner, and rebind/takeover status.

### Synergy sender

1. Allow targets to pin an expected verified `hostInstanceID`.
2. Verify it during `connect open` and heartbeat before reporting success.
3. Fail closed or require confirmation when the verified host changes.
4. Display target label, claimed Agent ID, Link ID, observed hostname, and verified host identity as distinct fields.
5. Do not silently overwrite the target's previously trusted host observation.

## Acceptance criteria

- Two Agent Tunnel WebSockets using the same Agent credential but different `hostInstanceID` values cannot silently replace one another.
- A legitimate reconnect from the same host instance continues to work without manual intervention.
- Cross-host replacement requires explicit authorization or produces a rejected/quarantined connection.
- Holos emits an auditable event for every attempted or accepted cross-host takeover.
- `synergy-link` refuses to adopt shared/copied Link state on another host unless an explicit rebind/lease transfer is performed.
- A Synergy target pinned to one verified host cannot return `connected/opened` and execute RPCs on another host.
- Tests cover shared state/credential access from two hosts and verify that the original target cannot silently redirect.

## Issue relationship

Synergy issue #983 should remain focused on stale collaboration sessions and false connected/opened state. This cross-host identity-takeover issue should be tracked separately, with Holos server as the primary owner and linked synergy-link/Synergy follow-ups.

### LINKED ISSUE #1022: Design a shared-filesystem-safe Synergy Link deployment model and runbook for Qizhi
# Proposed repository

`SII-Holos/synergy`

# Proposed title

Design a shared-filesystem-safe Synergy Link deployment model and runbook for Qizhi

# Issue body

## Summary

Qizhi deployments can run multiple Synergy applications and Synergy Link instances on different physical or container hosts while mounting the same persistent filesystem. The current Synergy Link deployment model does not provide a complete host-ownership boundary for this topology.

A Link instance directory placed on the shared filesystem contains portable mutable state and credentials. If another Qizhi host starts that instance, it can authenticate as the same Holos Agent and reuse the same Link ID. Holos currently treats the newest same-Agent tunnel as authoritative, so existing Synergy targets may silently route RPCs to the wrong physical host.

We need two things:

1. an explicit, fail-closed operations manual for safely running multiple Synergy Link instances on Qizhi today; and
2. a maintainer-reviewed deployment architecture that makes this topology safe by construction rather than relying on operator discipline.

This issue asks Synergy, synergy-link, Holos, and Qizhi maintainers to jointly define that model.

## Why Qizhi is different

In the affected environment, different compute instances mount the same persistent root:

```text
/inspire/hdd/project/multi-agent/<user-id>
```

The following are therefore true at the same time:

- two hosts can see the same Link instance directory;
- a process-local PID does not identify the owning host;
- a Unix control socket or local `flock` is not a cross-host lease;
- a stable shared path is useful for binaries and durable configuration but is not a machine-identity boundary;
- container hostnames may change across rebuilds, while logical deployments such as `atlas-qz` must remain stable;
- `/root` and other container-local paths are not suitable durable homes;
- there may be no SSH, SSM, systemd user service, or secondary remote-control channel available when the Link itself is down.

## Confirmed incident

Two Link deployments were intentionally placed on different Qizhi servers but shared the same persistent filesystem. During recovery, the Atlas Link wrapper from the shared filesystem was mistakenly started on the other server.

That process loaded the correct Atlas Agent credential and persisted Link state. Holos accepted the new same-Agent tunnel and replaced the original active tunnel. The existing Atlas target still referenced the same Agent ID and Link ID, so a freshly opened target session executed on the wrong physical host while target label, Link ID, Agent ID, and connected/opened status all appeared internally consistent.

The wrong-host process was removed. The incident is documented in Synergy #983, but #983 should remain focused on stale collaboration-session reconciliation. This issue focuses on Qizhi deployment topology and safe operations.

Related evidence:

- Synergy #983: stale collaboration session can remain locally `opened` after the remote session is gone.
- Holos currently uses `replaced_by_new_session` for a new tunnel authenticated as an existing Agent ID.
- Synergy targets persist `targetAgentID + linkID`; host observation is advisory and not a pinned trust identity.
- synergy-link can load Agent credentials and Link state from a portable shared HOME/state directory.

## Current fail-closed runbook

Until the product supports shared-filesystem-safe ownership, every logical Link deployment must have a unique instance namespace:

```text
$BASE/agents/synergy-link/instances/<instance>/
├── bin/
├── home/
├── state/
├── tmp/
└── scripts/
```

Required environment:

```text
HOME=$BASE/agents/synergy-link/instances/<instance>/home
SYNERGY_LINK_HOME=$BASE/agents/synergy-link/instances/<instance>/state
TMPDIR=$BASE/agents/synergy-link/instances/<instance>/tmp
```

Each logical deployment must use a distinct Holos Agent identity and Link ID. Link identity must never be shared with the Synergy application on the same host or with another Link.

### Before starting a Link

1. **Hard-assert the expected physical/logical host.** Shared path visibility is not proof that the operator reached the intended Qizhi instance.
2. **Verify the instance namespace.** Confirm executable, wrapper, HOME, `SYNERGY_LINK_HOME`, and `TMPDIR` all belong to the intended instance.
3. **Verify canonical Agent identity.** The Agent ID displayed in local configuration must match the `agent_id` claim returned for the stored Agent Secret. Never print or copy the Secret.
4. **Check singleton state.** Abort if multiple matching Link processes exist, or if a process exists without the expected operator-owned tmux/service owner.
5. **Do not start from another Qizhi host merely because the shared directory is visible.** This can take over the existing Agent tunnel.
6. **Do not reuse mutable Link HOME/state across logical deployments.** Sharing only the binary is acceptable; sharing credentials, `state.json`, `owner.json`, control socket, logs, or runtime state is not.

### Starting/restarting today

- Run the Link as a foreground process inside an operator-managed tmux session because Qizhi does not provide a reliable user systemd service.
- Restart only the Link service; do not conflate it with the Synergy application runtime.
- Use a host-local recovery lock to prevent two operators on the same host from racing. Treat it as local protection only, not a distributed lease.
- Require the old child process to exit before starting another copy.
- Require exactly one new matching process after startup.
- Capture only non-secret evidence: hostname, executable, PID/PPID, HOME, `SYNERGY_LINK_HOME`, `TMPDIR`, Agent ID, Link ID, and sanitized recent logs.

### Sender-side verification

A target is trusted only after a short real RPC proves:

- expected hostname or, preferably, a stronger future host-instance identity;
- expected Link executable;
- expected HOME / `SYNERGY_LINK_HOME` / `TMPDIR`;
- expected Agent ID and Link ID.

`connected`, `opened`, target label, target ID, Link ID, Agent ID, shared path visibility, and persisted `state.json` are not independently sufficient.

Remote RPCs should use `yieldSeconds <= 5`; longer work must background and use process polling. A timeout or `session_invalid` must be treated as loss of verified execution, not as a reason to try another shared host.

### Recovery when no independent control channel exists

If the Link is the only remote-control path and it is down:

1. stop remote retries;
2. use the Qizhi platform terminal on the intended host;
3. fail closed on host mismatch;
4. restart only the intended Link instance;
5. then clear sender-side stale collaboration state;
6. open a fresh Session ID and re-attest the complete host/process/identity set before any privileged work.

### 1. Cross-host identity takeover

A second host that can read the same Agent credential can become the active Agent tunnel. Existing targets may follow it without changing target metadata.

### 2. Shared mutable state corruption

Two hosts can update the same Link state, owner snapshot, control socket path, or logs. Local PID and socket checks cannot arbitrate cross-host ownership.

### 3. False recovery and misleading status

A stale collaboration session can leave a target shown as connected/opened. A newly registered wrong-host tunnel can then make a fresh open appear to recover the target while silently changing execution host.

### 4. Application/Link identity collision

If Link and Synergy application reuse the same HOME or Agent identity, their Holos tunnels can replace one another.

### 5. Operational lockout

Qizhi may not provide SSH/SSM/systemd or a second control plane. When the only Link fails, recovery requires manual platform-terminal access.

### 6. Credential portability

An Agent Secret copied through shared storage or image restore is currently sufficient to recreate the Agent tunnel elsewhere. Hostname and remote IP are recorded metadata, not authorization boundaries.

## Maintainer design questions

We need agreement across Synergy, synergy-link, Holos, and Qizhi maintainers on the intended deployment contract:

1. What is the stable identity of a Qizhi Link installation when container hostnames can change but logical deployments persist?
2. Should one Holos Agent identity be allowed to have multiple active Link tunnels? If yes, how are targets routed? If no, how is an authorized host migration distinguished from takeover?
3. Which layer owns the cross-host lease: Qizhi deployment operator, synergy-link, or Holos server?
4. How should a target pin and display its expected verified host identity?
5. How should an operator intentionally transfer ownership to another Qizhi host without copying credentials or silently replacing the tunnel?
6. Which state may live on the shared filesystem, and which state must be machine-bound or server-issued?
7. What recovery path is supported when the Link itself is the only remote-control channel?

### Holos server

- Introduce an installation-bound `hostInstanceID`, proven by an installation key rather than hostname/IP alone.
- Persist and enforce the current Agent/Link lease owner.
- Distinguish same-host reconnect from cross-host takeover.
- Reject, quarantine, or require explicit approval for a different `hostInstanceID` instead of unconditional last-writer-wins.
- Emit auditable attempted/accepted takeover events with old/new host instance, gateway, remote IP, and session IDs.
- Provide an explicit ownership-transfer/rebind operation.

### synergy-link

- Generate a non-shared installation identity/key on the intended host, or obtain one from the platform/server.
- Fail closed if persisted Link state is opened from a different host instance without an explicit transfer.
- Acquire and renew a server-visible lease.
- Store PID ownership together with host-instance identity; do not interpret a bare PID from shared state.
- Use reliable same-host exclusive locking, while acknowledging that local locks are not distributed locks.
- Surface effective host identity, lease owner, Agent identity, Link ID, and takeover/rebind state in `doctor` and runtime logs.

### Synergy sender

- Allow targets to pin an expected verified `hostInstanceID`.
- Verify the pinned host during open and heartbeat before reporting connected/opened.
- Fail closed or request explicit confirmation when the verified host changes.
- Display target label, claimed Agent identity, Link identity, observed hostname, and verified host identity as separate fields.
- Reconcile stale sessions independently from host-attestation failures.

### Qizhi deployment integration

- Prefer a platform-managed service/lease or supervisor rather than relying only on tmux.
- Provide a stable Qizhi workload/instance identity that survives container recreation but cannot be read and replayed from another mounted host.
- Expose a safe platform-terminal recovery procedure and a way to attest which compute instance is executing the Link.
- Document whether the shared filesystem supports reliable distributed locking; if not, do not represent `flock` as cross-host protection.

## Acceptance criteria

- A documented, maintainer-approved Qizhi deployment runbook exists for one or multiple Link instances.
- Starting the same shared Link state from another Qizhi host cannot silently redirect an existing target.
- Two logical Link deployments sharing `$BASE` cannot accidentally share Agent credentials, Link state, control sockets, or leases.
- Same-host reconnect works automatically; cross-host migration requires explicit ownership transfer.
- A target pinned to one verified host cannot report connected/opened and execute on another host.
- Status surfaces separate claimed identity, verified host, lease owner, and collaboration-session state.
- Tests cover two hosts mounting the same state, duplicate same-Agent tunnels, stale session plus takeover, deliberate host migration, and recovery after the original host returns.
- The final proposal documents the supported Qizhi service manager, recovery channel, state layout, secret boundary, and rollback process.

## Request for maintainer collaboration

This cannot be solved reliably by adding another local wrapper alone. Please involve maintainers from:

- Synergy target/session lifecycle;
- synergy-link runtime and deployment;
- Holos Agent Tunnel identity/routing;
- Qizhi platform/runtime operations.

The goal is to agree on a deployable Qizhi architecture and migration path, then split implementation into owner-specific follow-up issues. The current runbook should be reviewed as a temporary safety boundary, not treated as the final platform contract.

## Related issues

- SII-Holos/synergy#983 — stale collaboration-session reconciliation and false opened state. Keep its acceptance criteria separate.
- A separate Holos security/trust-boundary follow-up may be needed for duplicate same-Agent tunnel policy and cross-host takeover prevention.

yzxoi:
## Summary

- bind remote execution and process control to live collaboration leases, reap session-owned process trees, deduplicate retries, and bound foreground waits below the transport deadline
- reconcile cached sender sessions, Holos presence, delivery failures, target relinking, host CLI identity/status, and Link-facing tool errors with verified current state
- add localized remote-execution provenance plus a builtin `qizhi-synergy-link` Skill and fail-closed Qizhi deployment/recovery runbook

### Host execution boundary

- Derive an execution lease from the authenticated Holos caller and current collaboration session; never accept lease identity from the wire payload.
- Scope process listing, output, stdin, signals, removal, and cleanup to the creating session, caller Agent, and caller owner. Session close, kick, disable, or expiry terminates that lease's process trees and retained output.
- Revalidate the lease immediately before dispatch, make same-request retries idempotent within a session, and reject request-ID reuse with a different payload.
- Apply a best-effort pre-spawn guard to obvious direct detached-daemon launches without treating shell text scanning as a security boundary; auto-background remote foreground commands after at most five seconds, clamp blocking polls, and preserve process handles across ambiguous timeouts.

### Sender and targets

- Heartbeat-verify cached sessions before reporting them open; clear definitive invalid-session state, report transport timeouts as unknown, keep `lastAttemptAt` separate from `lastVerifiedAt`, expose unregistered raw sessions safely, and provide an ownership-guarded local cache clear.
- Bound Presence observations to five minutes, close and drain old tunnels on stop/reconnect, avoid stale offline short-circuiting, and preserve structured not_connected, offline, delivery_failed, and dispatched-timeout reasons.
- Add probe-first atomic target relinking for `targetAgentID + linkID`, roll back failed verification, clear the old cached session, and replace ambiguous `idle` availability with `unknown`, `unreachable`, `reachable`, or `connected`.

### CLI, identity, and diagnostics

- Resolve API, WebSocket, and portal endpoints from one validated source; missing values use defaults while malformed, credential-bearing, or cross-environment endpoints fail closed.
- Verify imported credentials against Holos `/me` before persistence so the authenticated Agent ID must match the supplied ID without exposing a secret or token.
- Make `doctor` mode-aware, treating local ownership as not applicable in standalone mode and validating managed ownership. Refresh runtime start identity and distinguish authoritative live control-socket status from degraded last-known snapshots with nonzero status.

### UI and operator guidance

- Mark remote Bash and process cards with a localized `Executed via Synergy Link` provenance tag while leaving local cards unchanged.
- Update `connect`, remote `bash`, and `process` model-facing descriptions and errors with stable-target preference, verified-session semantics, no-local-fallback guarantees, bounded yield, and ambiguous-timeout retry guidance.
- Add the builtin `qizhi-synergy-link` workflow and Qizhi operations runbook for per-instance writable-state isolation, platform-supervised lifecycle, live verification, independent recovery, credential rotation, target relinking, and rollback.

## Compatibility and topology

- Protocol remains Synergy Link v2; no Holos server change is included.
- Existing target records remain readable; relinking adds optional paired locator fields and target availability now uses `unknown | unreachable | reachable | connected`.
- Cross-host reuse of one Holos Agent credential is explicitly unsupported because current Holos routing cannot distinguish an ordinary reconnect from another physical host presenting the same credential. Qizhi deployments isolate every writable identity root; a replacement device receives a new Agent ID and is adopted only through a verified target relink.

## Testing

- `bun run typecheck && bun run build && bun test` (`packages/synergy-link`) — 80 tests pass across 17 files (354 assertions).
- `bun run test:ci` (`packages/synergy`) — 1,744 tests pass across 144 files (6,094 assertions).
- `bun run --cwd packages/app test` and `bun run --cwd packages/ui test` — both pass, including the production CSS build contract and local/remote provenance rendering.
- `bun run release:test` — 23 tests pass; package tests also pass for `plugin` (43), `plugin-kit` (19), `browser` (38), `synergy-link-protocol` (1), and the browser crypto contract (1).
- `bun run quality:quick`, `bun run format:check`, `bun run typecheck`, `bun run deadcode`, `bun run skill:check`, `bun run localization:check`, `bun run test-layout:check`, `bun run package:check`, `bun run monorepo:check`, SDK/OpenAPI regeneration, and `git diff --check` pass.
- `bun run secrets:check` could not run locally because `gitleaks` is not installed; the unchanged CI secret-scan job installs and runs it.

## Issues

Closes #982
Closes #983
Closes #984
Closes #985
Closes #986
Closes #987
Closes #988
Closes #992
Closes #993
Closes #994
Closes #1004
Closes #1021
Closes #1022


yzxoi:
Implemented the remaining Holos Agent Tunnel liveness gap from #1030 and folded it into this PR.

- Synergy now uses a 30s heartbeat / 90s pong deadline; standalone Link hosts use 60s / 180s.
- A silent half-open socket is retired through the existing bounded reconnect path.
- Pending Link and native requests settle exactly once as dispatched-but-result-unknown (`transport_liveness_lost`), and late responses from the retired provider generation cannot resolve them.
- Updated `connect`, remote `bash`, and `process` guidance plus the Qizhi builtin Skill/runbook to prohibit automatic mutating retries after ambiguous liveness loss.
- Also registered the host's `ps` system dependency so the CI deadcode/Knip gate recognizes the privacy-preserving process-owner scan.

Validation:

- `packages/synergy-link`: `bun run typecheck && bun run build && bun test` — 87 tests, 371 assertions.
- `packages/synergy`: `bun run test:ci` — 1,746 tests, 6,099 assertions.
- root: `bun run quality:quick`, `bun run deadcode`, and `git diff --check` pass.
- Independent security review: approved with no blocking findings.

yzxoi:
@codex review

Please review the latest head (`f9a71bfc8`), including the missed-pong liveness handling, ambiguous settlement, transport-generation isolation, Qizhi guidance, and follow-ups to the earlier inline comments.

yzxoi:
@codex review

yzxoi:
@codex review

The latest head addresses the three review findings:

- resolves extensionless Windows commands using cwd/PATH/PATHEXT with bounded, fail-closed inspection;
- encodes atomic Synergy Link relinks in the source schema, OpenAPI, and generated SDK type;
- sanitizes control-socket failures at the shared client boundary so CLI output cannot expose the local socket path.

Please review the latest commit set.

yzxoi:
@codex review

Please re-review the latest head. This follow-up addresses the newest actionable findings around Windows `cmd` caret escaping, in-flight request deduplication and bounded capacity, configured Holos base-path preservation, session-policy serialization, blocking-kick/open serialization, post-dispatch disconnect semantics, and target mutation locking. The PR body is unchanged.

yzxoi:
@codex review

yzxoi:
@codex review

yzxoi:
@codex review

Please review the latest commit, which fixes kicked-session requests timing out when the Link host rejected the session. The session path now returns correlated error envelopes, and unexpected host errors are redacted unless they carry a valid Synergy Link protocol error code.

Verification:

- `bun test test/host-hardening.test.ts`
- `bun run test` (115 pass, 0 fail)
- `bun run typecheck`
- `bun run build`
- changed-file Prettier check and `git diff --check`

yzxoi:
@codex review

The latest follow-up addresses the P2 finding on unexpected RPC execution errors. Non-protocol failures now retain their full diagnostics only in host-local logs and return a fixed public `execution_failed` message; valid Synergy Link envelope errors keep their actionable messages. A regression covers an `ENOENT`-shaped error through the public RPC handler path.

Verification:

- focused RPC redaction and valid-envelope tests
- `bun run test`
- `bun run typecheck`
- `bun run build`
- changed-file Prettier check and `git diff --check`

yzxoi:
@codex review

The latest follow-up addresses the probe-versus-relink P2. Every runtime probe and registered `connect open` observation now carries the exact `linkID + targetAgentID` it tested; the target service rechecks that locator under the per-target lock and drops stale results without writing storage or publishing an update.

Regression coverage includes a delayed old-locator failure after a successful relink and an old-locator host observation after replacement.

Verification:

- focused probe/relink regression
- Synergy Link target domain tests: 36 pass
- target runtime/service plus connect tool tests: 46 pass
- `bun run typecheck`
- changed-file Prettier, oxlint, and `git diff --check`
- `bun run test:changed`: 4192 pass, 2 skip; one unrelated contact-test hook timeout, with that file passing 5/5 immediately in isolation

PR #1035, opened 2026-08-02
fix: render resent messages after rewind without a forced refresh

yzxoi:
## Summary

- Rewind 后重新发送(retry 重发或手动重发)的消息前端不渲染、需强制刷新才显示。根因:派生的 `canUnrollback` 只在 rollback/unrollback API 中重算,消息写入路径不更新也不推送,前端 rollback 前缀裁剪(`messagesBefore`)因此一直隐藏新分支。
- **后端(方案 A)**: `Session.updateMessage` 检测 root user 消息使活跃 rollback 失效时,持久化并通过 `session.updated` 发布 `canUnrollback: false`。已持久化为 false 的摘要优先于实时派生(防止回写旧时间戳的消息重新激活 redo);当实时派生与已持久化摘要不一致时,通过新增的 `SessionHistory.rewriteStoredInfo` 回写真实投影。
- **前端(方案 C)**: rollback 过滤提取为 `messagesHiddenByRollback`,一旦 cut 点之后出现不在 dropped 集合中的消息(即新分支已加载),前缀裁剪即退化为 dropped-set 过滤——即使 rollback 摘要滞后于消息窗口,重发分支也立即可见。

## Test plan

- `bun test test/session/rollback.test.ts`(packages/synergy,17 个测试含 2 个新增:新 root 消息恰好发布一次 rollback 失效事件;非 root 消息不触发)
- `bun test test/session/session.test.ts test/session/summary.test.ts test/session/manager.test.ts`(87 通过)
- `bun test test/components/session/session-message-order.test.ts`(packages/app,9 个测试含 5 个新增过滤场景)
- `bun run quality:quick` 通过
- 已知基线问题(与本次改动无关,已在 origin/dev 基线复现): `test/session/inbox.test.ts` 2 个超时;packages/app 11 个布局类测试失败


yzxoi:
I found two blocking issues while reviewing this change:

1. `messagesHiddenByRollback()` currently treats any post-cut message outside `droppedMessageIDs` as proof that the replacement branch is loaded. A post-rollback non-root steer/injection also satisfies that condition even though the backend deliberately keeps redo available and prefix-hides all post-cut messages in that state. This can show a ghost steer message while the UI still offers redo. The frontend condition should mirror the backend invariant: only a root user message created after the rollback should degrade the prefix cut.

1. `publishRollbackInvalidation()` calls `SessionHistory.storedInfo()` on every root message, including sessions that never rolled back. That scans and reads the complete raw message/event history on the awaited turn-start path. After a rollback has already been invalidated, the current branch calls `rewriteStoredInfo()`, causing a second full scan plus a `Session.update()` write chain on every later root message. We should first use the persisted `session.history.rollback` projection as the cheap gate, return immediately when there is no rollback or it is already false, and perform at most one fresh derivation when the persisted value is still true.

I’m adding regression coverage for the non-root injection case and for avoiding repeated history derivation, then will push the focused fix to this PR branch.

PR #1045, opened 2026-08-03
fix: harden Windows process, sandbox, and watcher stability

TTAWDTT:
## Summary

Harden Windows runtime stability across process management, sandbox setup, server locking, and file watching.

## Original problems

- Child stdout and stderr were drained sequentially. If one pipe became full, the child and parent could block each other, causing commands to hang intermittently.
- The Job Object imposed an active-process limit of one, preventing shells from starting normal descendants such as `node`, `git`, pipelines, or npm processes.
- Server lock acquisition used a read-then-write sequence, allowing concurrent server starts to pass the lock check.
- Desktop termination handled only the direct server process, leaving Windows descendants such as Bun, cmd, plugins, or browser processes behind and potentially retaining ports or file locks.
- File watcher failures did not reliably trigger re-subscription, allowing file changes to be lost and configuration reloads to stop.
- Sandbox deny roots could cover unsafe locations such as a user directory or volume root, while ACL rollback did not fail closed strongly enough during abnormal cleanup.

## Changes

- Drain child stdout and stderr concurrently and bound stdio shutdown.
- Terminate the associated Job Object before joining drain workers.
- Preserve shell child-process support while retaining Job Object cleanup semantics.
- Make server lock acquisition atomic and fail closed on malformed or uncertain lock data.
- Improve Windows Desktop server termination and health-check failure handling.
- Recover failed file watchers and preserve rename/re-subscription lifecycle.
- Reject unsafe sandbox deny roots and harden ACL rollback behavior.
- Add Windows stability verification to CI.

## Validation

- `packages/synergy` typecheck passed.
- `packages/desktop` typecheck passed.
- Focused watcher, server-lock, and Desktop server-manager tests passed.
- Rust helper tests passed: 206 passed, 0 failed, with the two WFP tests excluded.
- Rust release build passed.
- `git diff --check` passed.

The full Rust test run still has two existing WFP test failures because those tests assert non-Windows stub behavior while running on Windows. They are intentionally left for a separate follow-up PR.

## Scope

This PR focuses on confirmed Windows stability issues.

It does not change shell selection, Desktop random-port behavior, or the broader `OVERLAPPED I/O` / `CancelIoEx` redesign for elevated-session cancellation. The WFP test-platform issue will be handled separately.


yzxoi:
Thanks for the Windows stability work. I completed another full pass over the latest head and found three concrete regressions:

1. **Windows sandbox default deny root reached the helper.** The normal gate → tool resolver path supplied the shared profile's default `dataDenyRoots: [homedir]`. For workspaces under the user's home directory, the helper's fail-closed ancestor check rejected the profile before spawning the command.
2. **A persistently unreadable server lock could spin forever.** After lock reads exhausted their retries without ever obtaining contents, acquisition returned to the outer loop with nothing it could safely quarantine.
3. **Rename followed by deletion was published as a rename.** The drain preserved the original rename for a later destination deletion, causing consumers to add a destination that no longer existed and exposing the wrong final state.

I added focused regression coverage and fixed all three paths:

- Windows wrapper calls now identify the shared default home deny root so the Windows backend removes only that default entry; explicitly supplied deny roots, including an explicit home root, remain intact.
- Lock acquisition now distinguishes a missing lock from an existing lock that remains unreadable and fails conservatively with `LockFileUncertainError`.
- A rename followed by destination deletion now triggers the serialized watcher resync path instead of publishing an invalid rename.

Verification completed:

- `cd packages/synergy && bun test test/sandbox/phase3-windows-config.test.ts test/file/watcher-events.test.ts test/daemon/server-process-lock.test.ts` — 56 passed, 0 failed
- `cd packages/synergy && bun run typecheck`
- `bun run quality:quick`
- `git diff --check`

The follow-up fix commit `2f9d2f5cd` has been pushed directly to this PR branch.

yzxoi (approved):

PR #1046, opened 2026-08-03
fix(bash): inject GH_TOKEN per gh command instead of requiring all-gh invocations

### LINKED ISSUE #995: bug: GH_TOKEN injection skipped unless every parsed command is gh
## What happened?

When a GitHub credential is connected through the frontend, Synergy injects it as `GH_TOKEN` into bash tool executions **only when every parsed command in the invocation is `gh`** (`canInjectGitHubCliToken` in `packages/synergy/src/tool/bash/local.ts`). Any additional command — `echo`, `ls`, `tail`, `head`, `awk`, or shell operators like `|`, `;`, `&&` — disables injection for the entire invocation.

In practice agents frequently combine commands, e.g. `gh api user | head`, `gh auth status 2>&1`, or `gh repo clone ... 2>&1 | tail -5`. All of these run without `GH_TOKEN`, and `gh` fails with "please run `gh auth login`" / "populate the `GH_TOKEN` environment variable", which misleads debugging into believing the stored credential is missing or expired. The same stored token works fine for a pure `gh` command: `gh auth status` then shows "Logged in ... (GH_TOKEN)".

## Expected behavior

The managed token should be usable in the common `gh` patterns agents actually run. Looser injection is preferred:

- Inject when the first/primary command in a pipeline is `gh` (downstream commands only consume stdout, so the leakage surface is minimal).
- Optionally, also let plain `git clone` use the stored token via a credential helper (an existing `buildCredentialCommand` pattern already exists in `packages/synergy/src/github/app-auth.ts`).
- At minimum, emit a notice when a `gh` command was detected but injection was skipped, instead of letting `gh` print a misleading login prompt.

## Steps to reproduce

1. Connect a GitHub account in Settings → GitHub (token stored).
2. Run a pure command: `gh api user` → succeeds.
3. Run a combined command: `echo ok; gh api user` → fails with "please run gh auth login or set GH_TOKEN".
4. Run a piped command: `gh repo view owner/repo | head` → same failure.

## Environment

- OS: macOS 15 / Ubuntu 24.04
- Synergy version: 3.0.10

## Logs

The `bash.github.token.injected` observability event is emitted only when injection happens; when injection is skipped there is no trace indicating the reason, which makes this failure mode hard to diagnose.

yzxoi:
## Summary

Closes #995.

The bash tool previously injected the managed GitHub token only when **every** parsed command in an invocation was `gh` (`canInjectGitHubCliToken`). Any mixed invocation — `echo ok; gh api user`, `gh repo view owner/repo | head`, `gh auth status 2>&1` — skipped injection entirely, so `gh` failed with a misleading "please run `gh auth login`" prompt even though the stored credential was valid.

This change prefixes each detected `gh` command with a command-scoped `GH_TOKEN='...'` assignment instead of injecting the token into the shared child environment. The token is visible only to the `gh` process itself:

- pipeline peers (`head`, `tee`, …) cannot read it
- later commands in the same invocation (`echo $GH_TOKEN`) cannot read it
- `ps`/argv never contains it (prefix assignment is not an argument)

Windows `cmd` (non-bash) has no command-scoped assignment, so it keeps the previous env-injection path for all-gh invocations and now emits a clear notice (plus a `bash.github.token.skipped` telemetry event) instead of failing silently when mixed commands are used.

Also adds a `bash.github.token.skipped` observability event with the reason whenever a `gh` command was detected but injection was skipped, addressing the missing diagnostics called out in the issue.

## Behavior

| Command | Before | After |
|---|---|---|
| `gh api user` | token injected | token injected |
| `gh api user \| head` | ❌ no token, gh auth error | ✅ token injected (scoped to gh) |
| `echo ok; gh repo view` | ❌ no token | ✅ token injected |
| `GH_TOKEN=explicit gh ...` | injected (double write) | explicit assignment preserved, no double write |

## Verification

- `cd packages/synergy && bun test test/tool/bash-github-token.test.ts` — 4 pass (incl. new prefix/quoting/behavior tests)
- `cd packages/synergy && bun test test/tool/bash` — 61 pass, 1 pre-existing skip
- `cd packages/synergy && bun run typecheck` — pass
- `bunx prettier --check` on both changed files — pass
- `bunx oxlint -c oxlintrc.json --deny-warnings` on both changed files — 0 warnings
- root `bun run quality:quick` — pass


yzxoi:
Addressed all four review comments in 3565935e.

**P1 — Keep the GitHub token out of the shell argv (fixed):** The prefix-rewrite approach is removed entirely. The managed token is now passed only through the child process environment (`sandboxEnv.GH_TOKEN`), so it never appears in the command string, the shell `-c` argv, or process listings. No AST rewriting of `gh` commands is performed anymore.

**P2 — Restrict prefix assignment to shells that support it (fixed):** With environment-only injection there is no command rewriting at all, so the mechanism is shell-agnostic — it works for `bash`, `zsh`, `csh`, `tcsh`, and Windows `cmd` alike. The previous POSIX/csh incompatibility cannot occur.

**P2 — Preserve credentials established earlier in the invocation (fixed):** The injected value lives in the child environment and is overridden by any explicit `GH_TOKEN=...` prefix assignment or `export GH_TOKEN=...` earlier in the command, per normal shell semantics. Covered by a new test (`local bash does not override an explicit GH_TOKEN established in the command`).

**P1 — Document the changed GitHub CLI credential behavior (fixed):** Added a "Bash tool GitHub CLI credential injection" section to `.synergy/skill/git-guide/SKILL.md` describing the environment-only injection, mixed/chained/piped coverage, explicit-token precedence, the skip notice, and the `bash.github.token.injected` / `bash.github.token.skipped` telemetry events.

Verification: `bun test test/tool/bash` (60 pass), `bun run typecheck`, prettier, oxlint, and root `bun run quality:quick` all pass.

PR #1050, opened 2026-08-03
feat: add compact live reasoning display

EricSanchezok:
## Summary

- add a General setting for opting into compact live reasoning, backed by the canonical config schema and generated SDK contract
- render only the latest active reasoning block as a stable plain-text tail line while preserving complete reasoning data
- avoid Markdown-driven layout jumps, clipped glyphs, and structural-only tail lines such as code fences

## Verification

- `bun test test/components/session-turn-timeline.test.ts` from `packages/ui` — 47 passed
- `bun run --cwd packages/synergy typecheck`
- `bun run quality:quick`
- browser verification confirmed a stable single-line height with ellipsis for paragraph, list, quote, heading, and long-text inputs


yzxoi:
Implemented https://github.com/SII-Holos/synergy/pull/1114.
Compact live reasoning display is quite difficult, especially for ds-v4-flash whose tps is so high haha :)

PR #1053, opened 2026-08-03
feat(plugin): add runtime endpoint lifecycle support

EricSanchezok:
## Summary

- add a capability-gated loopback runtime endpoint Host Service for external local plugin processes
- deliver startup and install lifecycle events and trusted Settings context to plugins
- bind the managed Desktop runtime to 127.0.0.1 on every platform

## Tests

- bun test test/server-manager.test.ts (packages/desktop)
- bun run desktop:test
- bun run typecheck
- bun run quality:quick


yzxoi (commented):
## Review: runtime endpoint lifecycle support

Nice scoped change overall. Loopback-only enforcement is the right fail-closed posture. Findings below; I will push fixes for the code-level ones to this branch.

### Design issues (worth deciding on)

1. **CLI default `0.0.0.0` makes the feature unusable out of the box.** `cli/network.ts` defaults `hostname` to `0.0.0.0`, and `isLoopback()` rejects it, so `runtime.endpoint.get()` throws `PLUGIN_RUNTIME_ENDPOINT_UNSAFE` under the default `synergy server` invocation. Since `0.0.0.0`/`::` still serve loopback, the check should treat wildcard binds as loopback-reachable and return `http://127.0.0.1:<port>` (fixing this in the PR, not by changing the CLI default which would break LAN access to `synergy web`).
2. **`runtime.started` vs endpoint availability.** `peekRuntimeEndpointGeneration()` does not check loopback, so the hook fires (and install catch-up runs) even when `get()` will always throw. Should be gated on loopback-reachability too.
3. **`freshInstall` is scope-scoped.** `install.ts` derives it from `state()` (scope-isolated), so re-adding a disabled plugin skips `lifecycle.install` and catch-up, while cross-scope adds can run it twice — contradicts the "once after fresh install transaction commits" doc.
4. **No retry path after failed `lifecycle.install`.** Degraded state lives only in the in-memory `contributionHealth` map (lost on restart) and re-running `add` won't re-trigger install. Doc promises "explicit retry" but nothing implements it.
5. **Context injection is not contribution-gated for this capability.** `context-factory.ts`/`runner.ts` only filter `agent.call` by `requires`; every contribution of a plugin declaring `runtime.endpoint.read` gets the service object even without `requires`, failing at call time instead of type level.

### Code issues (I'll fix these)

- Error codes `PLUGIN_RUNTIME_ENDPOINT_UNAVAILABLE`/`UNSAFE` are not part of the public `PluginHostServiceErrorCode` enum — plugin authors can't handle them type-safely.
- Parameter check runs on normalized params (`[]`, `null`, `"x"` all pass as `{}`), so "does not accept parameters" is bypassable.
- Shutdown clears the endpoint before `server.stop`, leaving a window where the server still listens but `get()` throws.
- Tests only cover direct rejections; no wildcard-bind coverage, and the non-loopback test uses `0.0.0.0` which should now be accepted.

### Minor

- `runtime.started` lacks `redactErrors: true` unlike sibling observers.
- `SettingsPanel.tsx` uses `section().context!` non-null assertion for an optional field.
- `runtime.started` input schema is empty `{}`; should validate `endpointGeneration`.

Happy to adjust if any of the above conflicts with an intended design.

yzxoi (approved):

PR #1055, opened 2026-08-03
TTAWDTT: Add configurable interface and monospace fonts

TTAWDTT:
## Summary

Synergy previously had no user-facing way to choose an installed interface font, and several English/monospace surfaces continued using hard-coded font stacks after the interface preference changed.

This PR:

- Adds separate interface and monospace font settings under General settings.
- Scans installed fonts with `queryLocalFonts()` and matches both family and full name, including localized names.
- Falls back to the default font when the browser cannot scan fonts, access is denied, or the requested font is unavailable.
- Persists the preferences, including migration from the earlier single-font preference shape.
- Applies the selected fonts to notes, inline code, math, Mermaid editors, terminals, Monaco source views, HTML previews, diagrams, and session UI.
- Refreshes iframe and editor-backed surfaces immediately when a font changes.
- Adds localized copy and regression tests.

## Validation

- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run monorepo:check`
- Font preference and HTML rendering tests pass in the package test suites.

## Screenshot

Font settings screenshot


yzxoi:
## Review: issues found & fixes pushed

Thanks for the PR — the direction (user-facing interface/monospace font settings) is solid. During review I found several correctness bugs and pushed fixes to `TTAWDTT/font-settings` (commit `8ea5d1b`, rebased onto latest `dev`). Here is what was wrong and what changed.

### Bugs fixed (with evidence)

**1. Stale async check clobbers newer input / silently undoes reset**
`checkAndApply` captured the requested family up front, then unconditionally wrote it back after `queryLocalFonts()` resolved. If the user kept typing (or clicked "Use default") while the check was in flight, the stale result overwrote the newer draft and re-applied the abandoned font.
Fix: a monotonic per-kind request sequence invalidates in-flight checks; stale results no longer touch state. Covered by a regression test.

**2. Every keystroke wrote localStorage + broadcast `synergy:font-change` + dropped the applied font**
`setFamily` ran `setStore` (synchronous localStorage write via `makePersisted`), cleared `appliedFamily`, and dispatched a global event on every `onInput`. Typing a 10-char name meant 10 storage writes and 10 full-surface refreshes — and the currently applied font vanished while typing.
Fix: the input is now a non-persisted draft; typing never writes storage, never broadcasts, and never disturbs the applied font. Only verified preferences are persisted on Check/Reset.

**3. Empty input + Enter reported "Not found" instead of resetting**
`checkAndApply("")` returned `missing` with a misleading status label. Fix: empty input on Check now resets to default (same as "Use default").

**4. Terminal font change was a silent no-op**
`ghostty-web@0.3.0` only applies `fontFamily` at construction; `handleOptionChange` for `fontFamily`/`fontSize` just logs "not yet fully supported" and never re-renders. The listener calling `t.options.fontFamily = ...` did nothing (and `fit()` re-measured with stale metrics). Fix: removed the dead listener; new terminal instances pick up the configured font from the CSS variable.

**5. Diagram layouts did not recompute after a font change**
The text-measure cache removal was right, but `DiagramGraph`/`DiagramSequence` layout memos only depended on container width, so nodes kept stale sizes until a resize. Fix: layouts now track a sans-font version signal (listening to `synergy:font-change`) and clear the measure cache.

**6. font-feature-settings not reset for third-party fonts**
The default fonts' OpenType features (`ss03`/`ss01` from `theme.css`) kept applying to arbitrary third-party families. Fix: applying a custom font sets `--font-family-*-font-feature-settings: normal`; reset removes it.

**7. Localized style suffixes did not match**
`withoutStyleSuffix` only stripped English suffixes, so fullName `"霞鹜文楷 常规"` failed to match a query for `"霞鹜文楷"`. Fix: also strip CJK suffixes (常规/斜体/粗体/细体/…). Covered by a regression test.

**8. Module-level default object mutation across instances**
`createStore` wraps its initial object in place, so `setStore` mutated the module-level `DEFAULT_PREFERENCES` — state leaked across provider instances (HMR remounts, tests). Fix: every provider gets a private copy.

### Also addressed

- `statusLabel` now shows "Checking…" during checks and "Press Check to apply" while editing.
- Reset button stays enabled when only an applied font exists.
- Migration from the earlier single-font shape is preserved (harmless for fresh installs).

### Verification

- New regression tests: `packages/app/test/context/font-preference-provider.test.ts` (8 tests: race, reset-race, empty input, missing keeps applied font, event storm, CJK suffix, feature-settings reset) — all pass.
- Original PR tests still pass.
- `i18n:check`, `format:check`, `lint`, `typecheck` (app + ui), and app/ui test suites pass locally.
- Branch rebased onto latest `dev` (resolved the `settings-panel.css` conflict by adopting the MenuField styles and re-adding the font controls).

One follow-up worth noting: runtime terminal font switching is limited by ghostty-web 0.3.0 itself (it cannot re-render with a new `fontFamily` after `open()`). New terminals use the configured font; a future ghostty-web release that supports runtime option changes would enable live updates.

Please review the new commit; happy to adjust anything.

yzxoi:
## Product-logic issues found while testing the current UI

I ran the current branch in an isolated instance and tested the font setting flow. The interaction model has real product problems:

### 1. Users cannot know the exact font full name — the input is unusable in practice

The current flow asks users to type a font's full name into a text input and press "Check and apply" to see if it exists. Users do not know the exact full name (or the localized name) of fonts installed on their machine — e.g. is it "LXGW WenKai", "霞鹜文楷", "LXGW WenKai Medium"? The feature forces them to guess, and any mismatch shows "Not found; using default" with no way to discover what is actually installed.

**Fix:** Check should _load the local font list_ (via `queryLocalFonts`) and present it as a selectable list, instead of validating a guessed name.

### 2. The right side should have exactly one button with two phases

Today the row has two buttons (Check + Use default) and the Check button doubles as the apply action. The intended interaction is simpler:

- Before the first load, the single action button reads **Check** — clicking it loads/refreshes the local font list.
- Once the list is loaded and the user has selected a font, the same button reads **Apply** — clicking it applies the selected font.
- A **Use default** button sits on the left of the selector, enabled only while a custom font is applied.

### 3. Input becomes a selector

The free-text input should become a selector (combobox) populated from the loaded local font list. The selector should show the currently applied font when one exists.

I'll implement this now and push the changes to this branch. The provider state machine will be reworked around per-kind phases (`idle → loading → ready → applied`, plus `unsupported`/`denied`), with the font list loaded once per Check and deduplicated by family name. Regression tests will be updated to the new interaction model.

TTAWDTT:
> ## Product-logic issues found while testing the current UI
> 在测试当前界面时发现产品逻辑问题
> I ran the current branch in an isolated instance and tested the font setting flow. The interaction model has real product problems:我在隔离实例中运行了当前分支并测试了字体设置流程。交互模型存在实际产品问题：
> 
> ### 1. Users cannot know the exact font full name — the input is unusable in practice
> The current flow asks users to type a font's full name into a text input and press "Check and apply" to see if it exists. Users do not know the exact full name (or the localized name) of fonts installed on their machine — e.g. is it "LXGW WenKai", "霞鹜文楷", "LXGW WenKai Medium"? The feature forces them to guess, and any mismatch shows "Not found; using default" with no way to discover what is actually installed.当前流程要求用户将字体的全名输入到文本输入框中，然后点击“检查并应用”来查看是否存在。用户不知道他们机器上安装的字体确切的全名（或本地化名称）——例如是“LXGW WenKai”、“霞鹜文楷”还是“LXGW WenKai Medium”？这个功能强迫他们猜测，任何不匹配都会显示“未找到；使用默认字体”，且没有方法来发现实际安装的字体。
> 
> **Fix:** Check should _load the local font list_ (via `queryLocalFonts`) and present it as a selectable list, instead of validating a guessed name.修复方案：检查功能应该加载本地字体列表（通过 `queryLocalFonts` ），并将其呈现为可选择的列表，而不是验证猜测的名称。
> 
> ### 2. The right side should have exactly one button with two phases
> Today the row has two buttons (Check + Use default) and the Check button doubles as the apply action. The intended interaction is simpler:今天这一行有两个按钮（检查 + 使用默认），检查按钮同时充当应用操作。预期的交互更简单：
> 
> * Before the first load, the single action button reads **Check** — clicking it loads/refreshes the local font list.在首次加载之前，单个操作按钮显示为“检查”——点击它会加载/刷新本地字体列表。
> * Once the list is loaded and the user has selected a font, the same button reads **Apply** — clicking it applies the selected font.一旦列表加载完成且用户已选择字体，该按钮会显示为“应用”——点击它会应用所选字体。
> * A **Use default** button sits on the left of the selector, enabled only while a custom font is applied.一个“使用默认”按钮位于选择器的左侧，仅在应用自定义字体时启用。
> 
> ### 3. Input becomes a selector
> The free-text input should become a selector (combobox) populated from the loaded local font list. The selector should show the currently applied font when one exists.自由文本输入框应变为一个下拉选择框（combobox），其内容来自已加载的本地字体列表。当有字体被应用时，选择器应显示当前应用的字体。
> 
> I'll implement this now and push the changes to this branch. The provider state machine will be reworked around per-kind phases (`idle → loading → ready → applied`, plus `unsupported`/`denied`), with the font list loaded once per Check and deduplicated by family name. Regression tests will be updated to the new interaction model.我现在将实现这个功能，并将更改推送到这个分支。提供者状态机将围绕每种类型的阶段（ `idle → loading → ready → applied` ，以及 `unsupported` / `denied` ）进行重新设计，字体列表将每个检查时加载一次，并通过家族名称进行去重。回归测试将更新到新的交互模型。

This is great polishing advice. It touches on two common paradigms for font settings. In that UI section, I referred to Codex. You might want to take a look at Obsidian—its design closely aligns with your description, and on top of the dropdown selection, it also incorporates the option for manual input.

TTAWDTT:
> Migration from the earlier single-font shape is preserved (harmless for fresh installs).

Thanks for the detailed review. I reviewed `8ea5d1b` and found one persistence migration regression: the legacy key `font-preference.v1` was no longer included in `Persist.global(...)`.

I restored the legacy key migration and added a regression test covering migration into `synergy.global.dat:font-preference` and cleanup of the legacy key.

I also confirmed the `ghostty-web@0.3.0` limitation: existing terminals cannot update their font at runtime; newly created terminals use the configured font. This limitation should be documented in the PR.

The fix is included in `3b4dfe3`, now pushed to `TTAWDTT/font-settings`. Pre-push checks passed: formatting, lint, typecheck, and monorepo validation.

yzxoi (approved):
## Approved ✅

Verified the font-settings flow end-to-end in an isolated dev instance, plus the full automated suite.

**What was verified**

- **Check** loads the locally installed font families (deduplicated + sorted), the selector populates, and the applied family is pre-selected after re-check.
- **Apply / Use default** lifecycle: Apply persists and applies the selected family (resetting `font-feature-settings` for third-party fonts); Use default restores the default and returns the row to idle.
- **Last fix**: the action button was deadlocked — disabled until `ready`, but `ready` required clicking it. Now Check stays clickable while idle (and retryable after unsupported/denied); only loading, or ready without a selection, disables it.

**Automated verification**

- 11 provider tests + 6 helper tests pass (`font-preference-provider.test.ts`, `font-preference.test.ts`)
- `typecheck`, `lint`, `format:check`, `i18n:check` all green
- CI: Quality, Typecheck, Desktop Checks, Package Validation, Smoke Test, Secret Scan, Workflow Validation pass; Test / Windows Checks still running on the latest push

**Housekeeping before merge**

- The branch is currently **BEHIND dev** — a final sync (merge/rebase onto `dev`) is needed to clear it and let the merge button enable.
- CI for the latest push (`2187f01c`) is still in flight; wait for Test / Windows Checks to go green.

Great work on the interaction model — the list-based selector is much more usable than the free-text input.

PR #1057, opened 2026-08-03
fix(app): stop infinite terminal reconnect after panel close

yzxoi:
## Summary

Closing the Terminal panel could leave it stuck on "Reconnecting..." forever. The inline reconnect loop reset its attempt counter on every socket `open`, so any accept-then-drop cycle (PTY exit, close race) bypassed `MAX_RECONNECT_ATTEMPTS` and rescheduled retries indefinitely.

## Changes

- **`packages/app/src/components/terminal/reconnect.ts`** (new): bounded `ReconnectController` state machine — gives up after `maxAttempts` quick-cycle failures, resets only after a stable connection (>5s), exponential backoff capped at 10s, single scheduled timer, stops on dispose/unmount.
- **`packages/app/src/components/terminal/terminal.tsx`**: replaced inline reconnect logic with the controller; guarded async `onMount` init against late unmount; `onCleanup` now wraps serialization/socket close/terminal dispose so cleanup always completes.
- **`packages/synergy/src/process/pty.ts`**: `onExit` now closes and clears subscriber sockets so clients stop writing into a dead PTY and their reconnect loop can settle into the "session lost" state.
- **`packages/app/test/components/terminal/reconnect.test.ts`** (new): 8 state-machine tests (quick-cycle give-up, stable-connection reset, validation failure/error give-up, dispose/unmount stop, single timer, capped backoff).

## Verification

- `bun test test/components/terminal test/context/terminal test/context/workbench test/components/workspace test/addons` — 112 pass
- `bun test test/workspace-contract.test.ts` (packages/synergy) — 8 pass
- `bun run typecheck` in `packages/app` and `packages/synergy` — pass
- `bun run quality:quick` (repo root) — pass
- Prettier check on changed files — pass


yzxoi:
Both Codex review points were addressed in commit `344d48025` (fix(app): distinguish missing PTY from reconnect exhaustion):

1. **P1 — Do not delete a live PTY when reconnects are exhausted**: `onGiveUp` now distinguishes `"missing"` (PTY confirmed gone via REST validation 404 → `onGone` → tab close → `pty.remove`) from `"exhausted"` (retries ran out while the PTY still validates → only the presentation marks the session lost, the process stays). Covered by `packages/app/test/components/terminal/reconnect.test.ts` ("PTY 仍存在但连接耗尽时标记 exhausted,不当作 missing").

2. **P2 — Assert the delay actually scheduled**: the exponential-backoff test now asserts `fake.delays()` equals the exact `ms` values passed to the fake timer (`[100, 200, 400, 400, 400, 400]`), so a constant-delay implementation fails the test.

Additional hardening in this PR (commit `45d3b74f8`):
- `Terminal.onCleanup` no longer writes the store (sync or deferred) during Solid's dispose flush — the root cause of the `cleanNode` → `Cannot read properties of null (reading '1')` crash after closing the first terminal.
- ws-close state writes are deferred past the dispose flush and guarded by `disposed`.
- `closeTab` gained a `closingTabs` re-entrancy guard for the ws-close → onGone → onRequestClose concurrent path.

CI: all checks green (Typecheck, Test, Quality, Windows, Desktop, Package Validation, Secret Scan, Smoke Test, Workflow Validation).

yzxoi:
@codex review

PR #1058, opened 2026-08-03
feat: make plugin runtime timeouts configurable

EricSanchezok:
## Summary

Makes five previously hard-coded plugin runtime timeout ceilings configurable through the canonical `pluginRuntimePolicy.limits` config domain (`50-plugins.jsonc`). Each new key defaults to the existing 120s behavior, so nothing changes unless configured:

- `agentCallMaxRuntimeMs` — ceiling for plugin `agent.call` / `agent.start` model invocations (was the hard-coded `AGENT_CALL_MAX_RUNTIME_MS` cap)
- `hookTimeoutMs` — per-handler timeout for plugin hook points (was the hard-coded 120s registry default)
- `contributionInvokeTimeoutMs` — fallback for contribution invocations with no declared `timeoutMs` (tools, authProvider, lifecycle handlers)
- `shellRunTimeoutMs` — default for plugin `shell.run` commands when the plugin omits `timeoutMs`
- `taskRunWaitTimeoutMs` — ceiling for how long `task.run` waits for a delegated task to reach a terminal state

Also fixes a wiring gap: `pluginRuntimePolicy.limits` was parsed but never applied to the plugin process runtime. Resolved limits now flow into `pluginRuntimeManager.start()` at every production call site (`ensureRuntime`, `reloadDevelopmentGeneration`, `prepareUpgrade`), so startup timeout, heartbeat interval, host-service request timeout, memory limits, and shutdown grace actually take effect. `invoke()` now falls back to the configured `contributionInvokeTimeoutMs`.

Config is read at plugin runtime start; changes take effect on the next plugin start/reload.

## Verification

- `cd packages/util && bun run typecheck && bun run build`
- `cd packages/synergy && bun test test/config/config.test.ts test/plugin/agent-call-host-service.test.ts test/plugin/shell-host-service.test.ts test/plugin/hook-contract.test.ts test/plugin/hook-timeout-config.test.ts test/plugin/task-run-host-service.test.ts test/plugin-runtime/manager.test.ts test/plugin-runtime/limits-wiring.test.ts` — 104 pass
- `cd packages/synergy && bun test` — 6531 pass, 2 skip, 4 fail (3 pre-existing environment flakes: ServerProcessLock stale-pid race, two 5s timeouts under full-suite load; 1 config test fixed and re-verified green)
- Root `bun run typecheck`, `format:check`, `lint`, `package:check`, `quality:quick` — all pass

## Notes

- No changes to the plugin manifest schema or runtime protocol version (additive `RuntimeLimits` fields only).
- Config-file only surface; no Web Settings panel (consistent with current `pluginRuntimePolicy` handling).
- Regenerated `packages/synergy/schema/config.schema.json` and `packages/sdk/js/src/gen/types.gen.ts` to include the new keys.


yzxoi:
## Review: limits are read per-call, but documented as read-at-start

I reviewed the full diff against `origin/dev`. The wiring fix is real and correct (`start()` already accepted `limits` and propagated it to the spawned process activation, memory monitor, startup timeout, and shutdown grace — but no production call site was passing it). Two issues to resolve before merge:

### 1. The five new keys are read per invocation, not at runtime start — docs say the opposite

`docs/reference/configuration.md` and the PR description say *"Limits are read at runtime start; changes take effect on the next plugin start/reload."* But the five new keys are resolved **on every call** through `resolvePluginRuntimeLimits()` → `Config.current()`:

- `manager.invoke()` — fallback for every contribution invocation without an explicit `timeoutMs` (plugin tools, `authProvider`, `lifecycle.upgrade`)
- `executePluginHooks()` — resolved before the handler loop, so a config change is visible to the very next hook trigger
- `resolvePluginAgentCall()` — per `agent.call` / `agent.start`
- `runPluginShell()` — per `shell.run`
- `runPluginTask()` — per `task.run`

Meanwhile the pre-existing keys (`startupTimeoutMs`, `heartbeatIntervalMs`, `hostServiceRequestTimeoutMs`, memory limits, `shutdownGraceMs`) genuinely are read once at `start()` and reused (including across memory-limit recycling via `#startInputs`). So one PR ends up with two different semantics, and the docs describe the behavior that is not implemented.

Suggested direction (happy to implement): resolve limits once at `start()` and store them on the runtime entry; all consumers read from the entry/input instead of `Config.current()`. This also kills the per-call `Config.current()` awaits on hot paths (`invoke`, hooks, host-service dispatch) and removes the `Config.current().catch(() => undefined)` silent-fallback when no ScopeContext is active.

### 2. `shellRunTimeoutMs` cannot actually cap plugin-declared timeouts

Only `agentCallMaxRuntimeMs` is a hard clamp (`positiveConstraint`). The other four are fallbacks — plugin-declared values win. In particular a plugin can declare an arbitrarily large `shell.run` `timeoutMs` and bypass `shellRunTimeoutMs` entirely, while `agent.call` is clamped. `docs/plugins/runtime-and-permissions.md` describes all five as "cap". Either keep the semantics but make the docs honest, or decide whether these should be ceilings.

Minor nits (not blockers):
- `PluginHookPoint.timeoutMs` is now dead — nothing reads it after `executePluginHooks` switched to `hookTimeoutMs`. Every point was already 120s, so no behavior change, but the per-point override capability is silently gone.
- `toolInvocationTimeoutMs` (existing key) now overlaps conceptually with `contributionInvokeTimeoutMs` — different directions, same default; worth a doc note.
- `executePluginHooks` resolves limits before checking whether any handlers exist, so every `tool.execute.before/after` trigger pays a config read even with zero plugins.

Tests are solid (real process fixture for hooks, boundary mocks for shell/agent/task). Happy to implement the read-at-start refactor if you agree with the direction.

yzxoi:
@codex review

yzxoi (approved):

PR #1060, opened 2026-08-03
fix: recover desktop native browser presentation

EricSanchezok:
## Behavior

- Require Desktop-native presentation for managed-local Desktop and keep Web/remote Desktop on explicit WebRTC.
- Add typed native capability/ticket coordination, broker restart/origin rebinding, strict server ticket errors, and native-only retry UX.
- Recover stable native page slots from renderer, attach, unresponsive, CDP, navigation, and CONNECT establishment failures without WebRTC fallback.
- Preserve page/profile identity across generation replacement and publish page-scoped recovery status to commands and UI.

## Risk

- Touches Desktop IPC, Browser presentation negotiation, Electron page lifecycle, and proxy timeout handling.
- Explicit native and WebRTC modes are now strict; recovery remains retryable and does not cross presentation modes.

## Tests

- `bun test --cwd packages/browser`
- `bun test --cwd packages/app test/components/workspace/browser`
- `bun run --cwd packages/desktop test:runtime`
- `cd packages/synergy && bun test test/browser test/server/browser-route.test.ts && bun run test:ci`
- `bun packages/app/script/private-http-smoke.ts`
- `bun run quality`
- Isolated managed Desktop startup and Browser Host broker registration with separate runtime state.


yzxoi:
## Review findings: unbounded native recovery loops + strict-mode gaps

The generation-replacement, single-flight recovery, and page-scoped status design is the right shape. I found several issues while reviewing — fixes are incoming on this branch.

### High severity

1. **`handleUnresponsive` "healthy" reload is unbounded** (`packages/desktop/src/browser-native-page-pool.ts`). When the CDP liveness probe succeeds, the handler runs `stop() + reload()` with no bound. A permanently blocked main thread (JS hot loop/deadlock) still answers CDP because it runs on a separate thread, so the page reloads forever every ~5s and never reaches the bounded `recover()` path. The probe proves CDP is alive, not that the page is healthy.
2. **`navigationRetries` is reset by redirects → unbounded reload loop**. `bindPageEvents` resets the retry counter on every `did-start-loading` unless `watchdogReloading` is set — but that flag is consumed by the *first* loading event after the watchdog reload. Any page that redirects during load (redirect chains, captive portals, login loops) resets the counter on the second loading event, so the 30s watchdog → reload → redirect → reset cycle repeats forever.
3. **`recoverEntry` corrupts state when closing the previous generation fails**. `closeGeneration(previous)` runs *after* the new generation is installed and after `replacementListeners` fired. If closing the old generation fails (the 5s close timeout is most likely exactly when the old renderer is wedged), the error is swallowed by the round loop while `entry.generation` already points at the new view: each retry round leaks a `WebContentsView` + `app.on("login")` listener, `host.status: ready` never fires, and the entry ends up in a broken state.

### Medium severity

4. **`selectBrowserPresentation` ignores `capabilities.webrtc` for explicit requests** (`packages/browser/src/protocol.ts`). The `: true` fallback makes the route's new `browser_webrtc_host_unavailable` branch dead code and turns "host unavailable" into a non-retryable 400 (`Browser Host does not support webrtc presentation`) instead of the intended strict retryable error. The new protocol test even codifies the wrong behavior.
5. **`resolveBrowserClientPresentation` catches bridge failures and returns `"native"`** — the opposite of the safe direction. If the capability IPC fails we cannot confirm managed-local, so we should fall back to WebRTC; as written, a broken bridge pins the session into a permanent native-failed loop.
6. **`browser-ws.ts` `connect()`: non-retryable ticket errors never reconnect**. When ticket issuance fails with a non-retryable error, `connect()` returns without ever creating the socket, and nothing schedules a reconnect — the Retry button (which only resets the coordinator epoch) cannot recover a session whose WS never opened.

### Low severity

7. `probeWebContents` attaches the debugger and never detaches when it was the one that attached.
8. Desktop `managedLocal`/ticket origin matching is a string compare of `URL.origin`; a `localhost` vs `127.0.0.1` alias in `SYNERGY_DESKTOP_APP_URL` silently downgrades to WebRTC.
9. Docs: recovery is described as "at most three rounds" but the implementation tries 3 rounds × 3 delays = up to 9 rebuild attempts; the unbounded reload paths are not documented.

I'll push fixes for 1–7 plus doc updates shortly; 8 is worth a follow-up design pass (hostname normalization).

yzxoi:
## Fixes pushed (4 commits)

All review findings 1–7 are addressed on this branch. Summary of what changed and why:

### 1. Bounded native recovery (unresponsive + navigation loops)
`browser-native-page-pool.ts` now keeps a **shared per-page recovery budget** (default 5) consumed by every automatic recovery action — renderer rebuilds, healthy-path unresponsive reloads, and the single navigation-watchdog retry. When the budget is exhausted the slot emits `host.status: failed` + `page.error` and waits for explicit Retry (which resets the budget). This closes the two unbounded loops:

- **Healthy-path reload loop**: a CDP-live but wedged renderer (main thread dead, CDP alive on its own thread) previously reloaded forever. Each reload now consumes budget.
- **Redirect reset loop**: `navigationRetries` was reset on every `did-start-loading`, so a redirecting page bypassed the one-retry limit. It now resets only when a navigation completes (`did-stop-loading`) or fails terminally (`did-fail-load`).

### 2. Recovery no longer rolls back on cleanup failure
`recoverEntry` previously `await closeGeneration(previous)` *after* installing the replacement; a failure there (most likely exactly when the old renderer is wedged) was swallowed into the round loop while `entry.generation` already pointed at the new view — leaking a `WebContentsView` + `app.on("login")` listener per round and never emitting `ready`. Closing the replaced generation is now best-effort; a successful replacement stays ready. `closeWebContents` also force-crashes a wedged renderer after the close timeout so a generation can never leak.

### 3. `probeWebContents` detaches its own debugger attachment
The probe previously attached and never detached; it now tracks whether it performed the attach and detaches only then, leaving transport-owned attachments untouched.

### 4. WebRTC capability is required for explicit requests
`selectBrowserPresentation` treated any explicit non-native request as available (`: true` fallback). It now checks `capabilities.webrtc`, so the route's `browser_webrtc_host_unavailable` branch is live again and host-unavailable surfaces as a retryable error instead of a non-retryable `Browser Host does not support webrtc presentation`. The route itself keeps WebRTC **on-demand**: the Host process is started by the control path on the first interactive command, so an explicit `webrtc` request during bootstrap does not fail while the capability is coming up. Updated the two tests that had codified the old behavior.

### 5. Bridge failure falls back to WebRTC (not native)
`resolveBrowserClientPresentation` and the coordinator's `capability()` catch returned `managedLocal: true` — a broken bridge pinned the session into a permanent native-failed loop. Both now return `managedLocal: false` → WebRTC.

### 6. Ticket failure no longer strands the WS session
A non-retryable ticket error in `connect()` returned without ever opening a socket, and nothing scheduled a reconnect (the Retry button only reset the coordinator epoch). `connect()` now schedules a reconnect after ticket failure, and `retryNative()` reconnects immediately when no socket is open/connecting.

### Tests
- `packages/desktop/test/browser-native-page-pool.test.ts`: +4 cases — healthy reload bounded by budget, redirects don't reset navigation retry, rebuild loop stops at budget, cleanup failure keeps a successful recovery.
- `packages/browser/test/protocol.test.ts` + `packages/app/.../browser-protocol.test.ts`: explicit `webrtc` without capability → `null`; with capability → `webrtc`.
- `packages/app/.../native-presentation-coordinator.test.ts`: +2 cases — capability probe throw → WebRTC fallback; coordinator stops native recovery on bridge failure.

All green: desktop 20/20, browser 13/13, app browser 41/41, synergy browser 7/7; `tsc`/`tsgo` typecheck, oxlint, prettier, and test-layout all pass.

yzxoi (approved):
## ✅ Approved

Reviewed the full change set and the fix commits on this branch. The recovery design (generation replacement, single-flight recovery, page-scoped status) is sound, and the follow-up fixes address all issues found in review:

**Fixed (verified at head `15795c19c`):**

1. **Bounded native recovery** — a shared per-page recovery budget (default 5) now caps every automatic recovery action: renderer rebuilds, healthy-path unresponsive reloads, and the navigation-watchdog retry. Exhaustion emits `host.status: failed` + `page.error` and waits for explicit Retry, which resets the budget. This closes both unbounded loops:
   - A CDP-live but wedged renderer previously reloaded forever; each reload now consumes budget.
   - `navigationRetries` previously reset on every `did-start-loading`, so redirects bypassed the one-retry limit; it now resets only on navigation completion or terminal failure.
2. **Recovery no longer rolls back on cleanup failure** — closing a replaced generation is best-effort; a wedged old renderer can no longer leak a `WebContentsView` + login listener per round or leave the entry stuck in `restarting`. `closeWebContents` force-crashes after the close timeout.
3. **`probeWebContents` detaches its own debugger attachment**, leaving transport-owned attachments untouched.
4. **WebRTC capability is required for explicit requests** — `selectBrowserPresentation` no longer accepts `webrtc` when the broker reports no capability; the route's `browser_webrtc_host_unavailable` branch is live again, and WebRTC stays on-demand so bootstrap is unaffected.
5. **Bridge failure falls back to WebRTC** — a broken native capability probe no longer pins the session into a permanent native-failed loop.
6. **Ticket failure no longer strands the WS session** — `connect()` schedules a reconnect after ticket failure, and `retryNative()` reconnects immediately when no socket is open.

**Verification:**
- New tests: pool +4 (bounded healthy reload, redirect-safe retry, budget exhaustion, cleanup-failure resilience), protocol +2 explicit-capability cases, coordinator +2 bridge-failure cases, browser-ws +2 reconnect cases.
- Local: desktop 20/20, browser 13/13, app browser 43/43, synergy browser 15/15; `tsgo -b`, oxlint, prettier, and test-layout all pass.
- Remote tree matches local at head; PR mergeable.

**Non-blocking follow-ups (not required for merge):**
- Hostname normalization for the `managedLocal` origin match (`localhost` vs `127.0.0.1` alias in `SYNERGY_DESKTOP_APP_URL` silently downgrades to WebRTC) — worth a follow-up design pass.
- CI (Actions) final results still pending at the time of review; all local-equivalent gates pass.

Nice work — the fix set is focused, well-tested, and preserves the strict-mode semantics the original PR introduced.

PR #1061, opened 2026-08-03
feat: add Grok OAuth provider via xAI subscription login

EricSanchezok:
## Summary

- Add a `grok` provider that mirrors the `openai-codex` integration using xAI's standard OAuth device-code flow (PKCE S256) against `auth.x.ai`, reusing the official Grok CLI public client id (no third-party OAuth registration channel exists).
- Tokens persist in the shared auth store with automatic refresh (`refresh_token` grant), rate-limit exhaustion and dead-credential transitions, all through the existing `Auth` / `ProviderAuthRecovery` infrastructure.
- Requests use the OpenAI-compatible `https://api.x.ai/v1` API via `@ai-sdk/xai` (already bundled), with `providerOptions` routed under the `xai` key.
- CLI `synergy auth login` gains a Grok branch with fallback priority; Web Settings reuse the existing provider-agnostic OAuth flow; the UI icon aliases `grok` to the xAI brand mark.
- Usage is reported as unavailable (xAI has no public usage API).

## Files

- `packages/synergy/src/provider/grok.ts` — device-code request/poll/exchange, PKCE, refresh, `resolveToken`, `grokFetchFor`, `classifyError` (new)
- `packages/synergy/src/provider/builtin.ts` — `grok` profile registration (`oauth_external`, `@ai-sdk/xai`)
- `packages/synergy/src/provider/auth.ts` — built-in "Login with Grok" method
- `packages/synergy/src/cli/cmd/auth.ts` — Grok auth branch + fallback priority
- `packages/synergy/src/provider/transform.ts` — `@ai-sdk/xai` providerOptions key routing
- `packages/ui/src/components/provider-icon.tsx` — `grok` → `xai` icon alias
- `packages/synergy/test/provider/grok.test.ts` — device flow, PKCE, refresh rotation, rate-limit/invalid-grant, fetch binding (new)
- `packages/synergy/test/provider/oauth-providers.test.ts` — reset list includes `grok`
- `docs/reference/configuration.md`, `docs/reference/cli.md`, `packages/synergy/AGENTS.md` — provider docs

## Verification

- `bun test test/provider/grok.test.ts` — 10 pass
- `bun test test/provider/` — 358 pass (15 files)
- `bun run typecheck`, `bun run lint`, `bun run format:check`, `bun run monorepo:check`, `bun run localization:check`, `bun run test-layout:check`, `bun run package:check`, `bun run quality:quick` — all pass

## Known constraints

- xAI has no public OAuth client registration; Synergy reuses the Grok CLI public client id (consent screen shows "Grok Build / Grok CLI" rather than Synergy). The id was verified from the official install script, four independent community implementations, and a live device-endpoint probe.
- Some subscription accounts may see HTTP 403 on inference despite valid OAuth (xAI allowlist); `classifyError` deliberately does not mark credentials dead on 403.


yzxoi:
## Review: Grok OAuth provider

Reviewed the full diff against `dev` and ran the provider suites. Overall the implementation is solid — the mirror of the `openai-codex` flow is faithful, `ProviderAuthRecovery` / `Auth` reuse is correct, PKCE + refresh rotation + dead-credential transitions are covered by tests, and the `@ai-sdk/xai` wiring (`modelFactory: "openaiChat"` → `createXai().chat`, `providerOptions` under the `xai` key, `handled()` fetch) checks out.

### Fixed: `slow_down` test is genuinely flaky (could intermittently fail CI)

Reproduced locally: the test hits `slow_down`, which grows the poll interval by 5s and sleeps for a real 5 seconds — landing right on the 5s per-test timeout. With `--rerun-each 3` it failed with `this test timed out after 5000ms` on run #2, and it failed intermittently in full-suite runs (2 of 4 runs).

Fix pushed to `synergy/grok-oauth` (`ab70cdbd`):
- `pollDeviceCode` now accepts an injectable `sleepFn` (defaults to `Bun.sleep`, production behavior unchanged).
- The test injects an instant sleep and now asserts the RFC 8628 interval growth (`[0, 0, 5000]`), so it runs in ~4ms instead of ~5s while still verifying the `slow_down` semantics.

Verified: `bun test test/provider/grok.test.ts` 10/10 pass, `tsgo --noEmit` and Prettier clean.

### Follow-up (shared with codex, suggest tracking via a separate issue): rate-limit cooldown surfaces as "credentials rejected"

When refresh hits 429, `resolveToken` marks the credential `exhausted`; during the cooldown `Auth.select` returns `undefined`, so the next request throws `grok_auth_missing` (`reloginRequired: true`) and `ProviderAuthRecovery` drives the user to "These credentials were rejected. Reconnect the provider" — even though the credentials are fine and only rate-limited. This is copied verbatim from `openai-codex`, so it affects both providers; worth a dedicated issue rather than a per-provider patch.

### Minor (same shared pattern): Web cancel does not abort polling

`authorizeDeviceCode().callback()` ignores the `AbortSignal` passed by `ProviderAuth.callback`, so cancelling the Web flow leaves the poll running until the 15-minute cap. Same behavior as codex — could be bundled into the issue above.

### Note: external assumptions cannot be verified offline

The shared Grok CLI `client_id` (ToS gray area, shared rate-limit surface), the scopes string, and the `grok-4.5 / grok-4.3 / grok-build-0.1` model IDs rely on the live probes described in the PR body. Worth a comment in the provider docs that these are pinned to the current Grok CLI flow and may need updating if xAI changes it.

yzxoi:
### 🔴 Fixed: provider detail icon froze on the first rendered provider

Found while testing the dev instance: after this PR's alias refactor, the provider detail panel icon stopped updating when switching providers — every provider showed the same icon (the initially selected one, e.g. deepseek).

**Root cause**: the alias refactor moved `resolveIconID(local.id)` out of the JSX into a plain `const iconID` in the component body. SolidJS component bodies evaluate only once, and the detail panel uses a non-keyed `<Show>`, so `iconID` was frozen at its first value. The old code read `local.id` directly inside the JSX (`<use href=...>`), which is reactive.

**Fix** (`fa5e8f27d`, pushed to `synergy/grok-oauth`):
- `provider-icon.tsx`: wrap the resolved id in `createMemo` so both the `<Show when>` condition and the sprite `href` stay reactive when `props.id` changes.
- Added `packages/ui/test/components/provider-icon.test.ts` covering: switching the provider id updates the href, the `grok → xai` alias, and unknown ids hide the icon.

Verified: new tests fail on the pre-fix code and pass after; full UI suite passes; `tsgo --noEmit` and Prettier clean.

yzxoi:
### 🔴 Fixed: selecting a single-method OAuth provider auto-opened the browser verification page

Reported while testing the dev instance: clicking `grok` in the provider list immediately opened the xAI verification page, before clicking "Login with Grok".

**Root cause**: `ProviderConnectionFlow` auto-advances a provider with exactly one auth method on mount (`methods().length === 1` → `selectMethod(0)`), a behavior that predates this PR and is fine for API-key providers (it just reveals the key form). But grok registers a single `oauth` method, so selecting it silently called `authorize`, obtained a device code, and `platform.openLink()` opened the external verification page — no user click involved.

**Fix** (`20f1f8a58`, pushed to `synergy/grok-oauth`):
- Added `shouldAutoAdvanceConnection(methods)` in `provider-connection-model.ts`: auto-advance only when the single method does **not** launch an external flow (`api` / `import`). A single `oauth` method now waits for an explicit click on the login row.
- Wired it into `ProviderConnectionFlow` (still gated by `!connected()` and `skipAutoAdvance`).
- Added unit tests covering api/oauth/import/multi-method cases and updated the Providers panel contract assertion.

Verified: model + panel tests pass (16/16), `tsgo --noEmit` and Prettier clean. Selecting grok now shows the "Login with Grok" method and only opens the browser after the user clicks it.

yzxoi (approved):
## ✅ Approved

Reviewed the full diff against `dev` and tested the provider in an isolated dev instance. The Grok OAuth provider (xAI device-code flow) faithfully mirrors the proven `openai-codex` pattern — PKCE, refresh rotation, dead-credential transitions, `ProviderAuthRecovery` reuse, and `@ai-sdk/xai` wiring all check out.

Issues found during review were all fixed in this branch:

1. **Flaky `slow_down` test** (`ab70cdbd`) — the test slept a real 5s and intermittently hit the 5s per-test timeout. `pollDeviceCode` now accepts an injectable `sleepFn`; the test runs in ~4ms and asserts RFC 8628 interval growth `[0, 0, 5000]`.
2. **Provider detail icon froze** (`fa5e8f27d`) — the alias refactor moved `resolveIconID(local.id)` out of JSX into a plain constant, so SolidJS never re-evaluated it when switching providers. Now wrapped in `createMemo` with component tests.
3. **Auto-opened OAuth verification page on select** (`20f1f8a58`) — `ProviderConnectionFlow` auto-advanced single-method providers, so selecting grok silently called `authorize` and opened the xAI page before any click. Auto-advance now only applies to non-OAuth methods (`shouldAutoAdvanceConnection`), with unit tests.

Verified: grok tests 10/10, oauth-providers 27/27, provider-icon tests 3/3 (red before fix), connection-model + panel contract 16/16, `tsgo --noEmit` and Prettier clean, and end-to-end in a dev instance (selecting grok now waits for an explicit "Login with Grok" click).

Suggested follow-up (separate issue, shared with `openai-codex`): a 429 rate-limit cooldown surfaces as "credentials rejected, reconnect provider" because `resolveToken` marks the credential exhausted and `Auth.select` returns undefined during cooldown — worth fixing for both providers together.

PR #1066, opened 2026-08-04
fix: ignore unavailable model default variants

EricSanchezok:
## Summary
- hide configured default variants that the selected model does not support
- omit incompatible agent and role defaults when materializing new or legacy root messages
- preserve validation errors for explicitly selected unavailable variants

## Verification
- bun run quality:quick
- bun run release:test
- bun run --cwd packages/app test
- bun run --cwd packages/app build
- bun test ./test/session/input.test.ts ./test/session/migration.test.ts ./test/session/llm-variant.test.ts


yzxoi:
## Review: found a regression, added a fix

Thanks for the PR — the core change (omit configured defaults the selected model doesn't support, while preserving explicit validation) is the right approach. During review I found one regression that would break session import and the variant migration, and a small frontend edge case. Both are fixed and pushed to this branch.

### 🔴 Regression: `resolveLegacyRoot` now throws for stale model references

Before this PR, `resolveLegacyRoot` only read config/agent defaults and never threw. The new implementation routes through `resolveForRoot` → `Provider.getModel()`, which **throws `ProviderModelNotFoundError`** when the provider or model no longer exists (e.g. a removed provider, a renamed model, or an import into an environment without that provider configured).

That error propagates into two unguarded callers:

- **Session import** (`session-import.ts:335` in `remapMessage`): an export containing a legacy root whose model is gone now fails the entire import with a 400 — previously it imported fine.
- **Migration `20260726-session-root-variant`** (`migration.ts:1861`): the outer try/catch wraps the whole session block, so one stale root skips variant backfill for the *entire session* instead of just that message.

### Fix (commit `176db3e9`)

`resolveLegacyRoot` now catches `Provider.ModelNotFoundError` and returns `undefined` (same behavior as when the agent is missing), so the root stays unresolved instead of aborting the import/migration:

```ts
try {
  return await resolveForRoot({ agent, model: user.model })
} catch (error) {
  const { Provider } = await import("@/provider/provider")
  if (Provider.ModelNotFoundError.isInstance(error)) return undefined
  throw error
}
```

### 🟡 Frontend: filter while model is unknown

`local.tsx` passed `[]` as `availableVariants` when the current model wasn't loaded yet, which hid configured defaults during provider-list loading. It now passes `undefined` ("don't filter when unknown"), matching the backend's `!input.model → undefined` behavior.

### Tests added

- `session-import.test.ts`: importing a legacy root referencing a removed provider/model succeeds, leaving variant unresolved.
- `migration.test.ts`: the variant migration skips a stale-model root and still backfills a sibling root in the same session.
- `composer-intent.test.ts`: defaults are not filtered when the available variant set is unknown.
- Also gave the PR's own slow import test (`canonicalizes missing root variants…`, ~5.8s) a 15s timeout — it flaked at the default 5s in parallel runs.

### Verification

- `bun test test/session/input.test.ts test/session/session-import.test.ts test/session/migration.test.ts` — 34 pass / 0 fail
- `bun test test/context/prompt/composer-intent.test.ts` (packages/app) — 39 pass / 0 fail
- `bun run typecheck` in packages/synergy and packages/app — clean
- Prettier — clean

One unrelated note: the branch's `bun.lock` is stale vs `dev` (missing the `@ericsanchezok/synergy-util` entry under `packages/synergy-link`), so `bun install` touches it locally. CI uses plain `bun install` so it won't fail, but rebasing onto `dev` would pick it up. I did not include the lockfile change in the push.

yzxoi (approved):
LGTM — 验证了模型不支持默认变体时静默省略、显式变体仍校验的行为，导入/迁移回归已修复并有测试覆盖。

PR #1068, opened 2026-08-04
feat(email): harden email tools with hybrid search, MIME decoding, and attachment metadata

EricSanchezok:
## Summary

Hardens the Synergy email subsystem (`email_send` / `email_read` tools) based on a full functional test and dual code review against a real 163 mailbox, and adds attachment metadata to read results.

- **Correct MIME decoding**: bodies are now parsed with `mailparser` (transfer-encoding + charset decoding). Fixes the verified bug where quoted-printable emails came back with raw `send =` line-wrap artifacts.
- **Working from/subject/text search everywhere**: date/flag keys run on the IMAP server; sender/subject/body filters run locally over a bounded newest-first window. Fixes empty search results on servers that ignore header search keys (verified: 163 Coremail returns empty for `UID SEARCH SUBJECT/FROM/TEXT`). No per-vendor branches.
- **Attachment metadata**: `email_read` now lists attachment file name, size, and content type without downloading content.
- **Resource caps**: 10 MB per-message download cap with `truncated` signaling (envelope-only fallback), bounded scan windows (`200` → widen `1000`, text scan `50`).
- **Fail-fast date validation**: invalid `since`/`before` rejected with a clear error instead of silently becoming `Invalid Date`.
- **Preserved structured config errors**: `NotConfiguredError`/`DisabledError` (with `missing` fields) now propagate instead of being swallowed into a generic `FetchFailedError`.
- **Consistent ordering**: search results newest-first; `summaries`/`read` process UIDs in given order with `limit` capping.
- **No plaintext password in the SMTP transport cache key** (identity fields only).
- **First automated test coverage** for the email subsystem (39 tests): quoted-printable decode regression, base64, attachment metadata, hybrid-search orchestration matrix (server keys, local filters, window widening, truncation), config-error propagation, transport-key assertions, and tool output shape.

## Verification

- New suites: `bun test test/email/* test/tool/email-read.test.ts` — 39 pass, 0 fail
- Root gates: `typecheck`, `lint`, `format:check`, `localization:check`, `monorepo:check`, `package:check`, `test-layout:check` all pass
- `test:changed` full run: 4335/4337 pass; 2 failures are pre-existing session-migration timeouts under parallel load (isolated rerun green)
- Live round-trip on the configured 163 account (new code): subject `AAAI` and from `scholight` searches now hit (previously empty), quoted-printable decode clean, date search returns 5, `markSeen` round-trip OK

## Scope

- `packages/synergy/src/email/*` (imap rewrite, new search-filter module, transport key)
- `packages/synergy/src/tool/email-read.*` (hybrid search, date refine, attachment rendering)
- `packages/synergy/test/email/*` + `test/tool/email-read.test.ts` (new)
- `packages/synergy/package.json` (+`mailparser`, `@types/mailparser`)
- `docs/product/connections.md`

No changes to config schema, permission taxonomy, SDK, server routes, or the settings UI.


yzxoi:
## Review: search ordering, scan bounds, truncation detection

Thanks for this hardening pass — mailparser decoding, config-error propagation, fail-fast date validation, the transport-key cleanup, and the new test coverage are all solid, and CI is green. I found a few correctness issues in the hybrid-search path; fixes are incoming on this branch (see follow-up).

### 1. Body-text search scans the wrong end of the window (HIGH)
IMAP SEARCH returns **ascending** UIDs, and every candidate window here is ascending, so `scanTargets = matched.slice(0, TEXT_SCAN_LIMIT)` picks the **oldest** 50 messages inside the newest-200 window. The newest 150 can never be text-matched, and the "narrow by date" hint makes results worse, not better.

### 2. "Newest-first" was never actually implemented (HIGH)
`runSearch` returns `matched.slice(-limit)`, which is ascending — the opposite of the documented contract ("Results are returned newest-first" in `email-read.txt` and `docs/product/connections.md`). Since `read`/`summaries` process UIDs in the given order, the agent ends up reading oldest-first. The tests encode the same ascending assumption, which is why they pass.

### 3. Date/flag + from/subject triggers an unbounded fetchAll (MEDIUM)
The server-keys branch (`email-read.ts`) passes no limit, then `fetchSummaries` joins **every** matching UID into a single FETCH. `since: 2020` + `from` on a large mailbox can exceed server command-length limits — contradicting the PR's "bounded scan windows" goal.

### 4. Truncation detection relies solely on the server-reported size (MEDIUM)
`meta.expectedSize > EMAIL_MAX_BYTES` is the only signal. `LimitedPassthrough` (imapflow v1.3.2) silently drops bytes past the cap, so on servers that don't report a literal size, a capped message is parsed as if complete and returned with `truncated: false`.

### 5. SMTP transport cache key no longer detects password rotation (MEDIUM)
Dropping the password from `transportIdentityKey` means a rotated password reuses the pooled transport with the old credential for up to the 60s idle window (old implementation keyed on the full config and rebuilt correctly).

### Open items (not addressed in this pass)
- `text`-only searches never widen the window (only `from`/`subject` do); widening a text scan would need to scan a different slice, so it stays bounded at 50 with `truncated` flagged.
- `read` and the text scan open one IMAP connection per message (serial connect/login/logout per UID).
- `download()` fetches the full RFC822 message including attachment bytes, so "attachment metadata without downloading content" is approximate; true part-level fetching needs BODYSTRUCTURE + per-part downloads.
- `unseen: false` now means "seen only" (previously ignored) — flag if that's unintended.
- `candidateUids.length >= windowLimit` widens once when the mailbox is exactly at the window size (off-by-one, harmless).

yzxoi (approved):
Isolated-instance testing confirmed the fixes: email_read search now returns newest-first UIDs, text filters scan the newest candidates, truncation is detected via byte-count fallback, and SMTP transport rebuilds on password rotation. All CI checks green. Approving.

PR #1069, opened 2026-08-04
feat(browser): action settle, semantic tool results, and rich tool cards

EricSanchezok:
## Summary

- **Action-level settle**: `browser_action` and agent navigation now settle the page by default (quiet-network strategy, up to 30s) and return a fresh accessibility snapshot; a settle timeout reports `settled: false` instead of failing the action. Conditional `browser_wait` remains for business conditions the engine cannot infer.
- **Semantic tool results**: action/navigation/wait outputs and metadata now expose settle outcome, URL/title/loading, target summary, snapshot element counts, and elapsed time; tool descriptions read as decision manuals (when to settle, when to wait, when to opt out of snapshots).
- **Rich Web tool cards**: browser_action/navigation/wait/snapshot cards show localized subtitles, tone-coded settle/loading tags, and summary rows via shared scalar-safe extraction helpers (no raw input values or `[object Object]` in the DOM); permission-dock summaries use the same extraction.
- **Plumbing**: broker host timeouts and abort mapping cover settle windows; `browser_wait` timeout now honors `timeoutMs`; protocol fields are additive/optional; SDK types regenerated.

## Verification

- `packages/browser`: typecheck + 44 tests pass (new settle success/timeout/load/none/navigation/wait coverage)
- `packages/synergy`: typecheck + 130 browser/tool tests pass (new tool unit tests for action/navigation/wait)
- `packages/ui`: typecheck + full isolated test suite pass (new browser render contract tests)
- `packages/desktop`: typecheck + tests pass
- Root gates: `typecheck` 11/11, `lint`, `format:check`, `localization:check`, `quality:quick` all green


yzxoi (commented):
## Review findings + fix commit

Reviewed the full diff and traced the command flow end-to-end (controller → broker → command-service → tools → Web panel → Desktop Host). Found one high-impact behavior regression and several robustness issues. A fix commit `55622e80e` was pushed to this branch; details below.

### 🔴 1. User-initiated navigation/reload/history was forced through 30s settle + full AX snapshot (behavior regression)

The PR description says *agent* navigation settles, but the implementation applied settle to **all** commands regardless of `source`:

- `navigate()`/`history`/`reload` in `packages/browser/src/controller.ts` always ran `settle({ mode: command.settleMode ?? "networkquiet" })` and `snapshot()` unless `includeSnapshot: false` was explicitly passed.
- The Web panel (`browser-store.tsx`, `browser-panel.tsx`), session restore (`session.ts` `ensurePage`), and Desktop initial navigation (`browser-native-page-pool.ts`) never pass those fields, so **address-bar navigation, back/forward, reload, and session restore each waited up to 30s and paid a full `Accessibility.getFullAXTree` cost**. Because the Host serializes commands per owner key, one slow settle also blocks every subsequent command on that page.

**Fix**: `source === "user"` now defaults to `settleMode: "none"` + no snapshot (`settleModeFor`/`includeSnapshotFor` in controller). Also split the shared `history`/`reload` schemas into user variants (`source: "user"` default) and backend variants (`source: "agent"` default) so the default truly matches the caller.

### 🟠 2. Settle timeout still forced page state + snapshot, turning successful actions into errors

The PR promises "a settle timeout reports `settled: false` instead of failing", but after the timeout the code still ran `pageState()` (`Runtime.evaluate`) and `snapshot()` (`Accessibility.getFullAXTree`). During mid-navigation those CDP calls can throw (`browser_frame_context_unavailable`, context destroyed), so an action/navigation that *worked* would fail anyway. `reload` newly exposed this because dev returned `void` without touching the page state.

**Fix**: `safePageState()` / `safeSnapshot()` degrade gracefully; snapshot is only attempted when `settle.settled` is true.

### 🟠 3. browser_action output reported pre-action URL/loading

`browser-action.ts` used the `page` object captured by `resolvePage()` before the action. After a navigation-triggering click, the output showed the **old** URL with `Loading: no` — contradicting `settled: false`.

**Fix**: action results now carry a fresh `page` state; the tool prefers it (`result.page ?? page`).

### 🟠 4. Pre-existing inflight requests could stall every settle

`inflightRequests` is never cleared on navigation. A long-polling fetch, keep-alive beacon, or a request orphaned by navigation stays in the set forever, so every subsequent action/navigation ate the full 30s budget returning `settled: false`. Pages with periodic polling would effectively never settle.

**Fix**: settle only counts **new** requests against quietness (baseline snapshot at settle start).

### 🟡 5. Minor

- `ToolTimeout` reported a 30s operation timeout for `browser_navigation` even for `current`/`stop`/`close`; now limited to settling actions (`goto`/`back`/`forward`/`reload`).
- `settleReason: "interrupted"` remains declared but never produced (kept for forward compatibility).
- `reload` result shape `void → navigation` is a protocol-level behavior change — worth calling out explicitly in the PR description (SDK types were regenerated and internal consumers adapted).

### Verification

- `packages/browser`: typecheck + 18 controller tests pass (incl. new settle tests, unchanged behavior for `settleMode: "none"` contract tests)
- `packages/synergy`: typecheck + 13 browser tool/timeout tests pass
- `packages/app` and `packages/desktop`: typecheck pass
- Changed files: `packages/browser/src/controller.ts`, `packages/browser/src/protocol.ts`, `packages/sdk/js/src/gen/types.gen.ts`, `packages/synergy/src/tool/browser-action.ts`, `packages/synergy/src/tool/timeout.ts`

yzxoi (approved):

PR #1070, opened 2026-08-04
fix(app): isolate stale message renderer failures

EricSanchezok:
## Summary

- isolate message-part renderer failures from disposed Solid reactive owners by snapshotting stable part identity before child evaluation
- key session-turn and message-role boundaries so session changes replace subtrees instead of reusing stale owners
- label source-managed App builds with their Git revision and emit source maps, while preserving release version semantics
- add a real Chromium regression that reproduces the secondary `Stale read from <Show>` failure and verifies the original renderer error stays local

## Root-cause evidence

The production stack resolves to the message-part `ErrorBoundary` fallback. A part renderer fails first; while Solid is handling that error, the fallback reads the live `part.type` accessor from the disposed owner. That secondary stale read escapes the local boundary and replaces the original renderer error. The regression test recreates that lifecycle exactly and failed with the stale-read exception before this change.

## Verification

- `bun run quality:quick`
- `bun run release:test`
- production App build and source-managed App build, including build-label and source-map artifact checks


yzxoi:
## Review: keyed turn boundary rebuilds the whole subtree on message updates (P1), plus smaller findings

Thanks for the fix — the root-cause evidence is solid. I verified locally before commenting:

- `message-part-error-boundary.test.ts` passes (with an explicit 30s timeout)
- `build-identity.test.ts`, `dev.test.ts` and the full `packages/ui` suite pass
- `SYNERGY_APP_BUILD_KIND=local bun run build` emits the label into the bundle and produces source maps
- All 9 CI checks are green

### P1 — `keyed` on the turn boundary destroys and recreates the whole turn on every message update

`SessionTurn` now uses `<Show keyed when={message()}>` and `<Match keyed …>`. In Solid, keyed means the subtree is torn down and rebuilt whenever the `when` value changes identity. On the backend, each assistant turn end emits `message.updated` at least 2–4 times (summary `diffState` pending→ready in `session/summary.ts`, `updateAssistantContextUsage` in `session/processor.ts`). Each event lands as a fresh message object via `reconcile(..., { key: "id" })` in the sync layer, so the keyed boundary rebuilds the entire turn — including every tool card, the Markdown render, and the shell-mode view.

User-visible effects:

- Expanded `BasicTool` cards collapse (they hold `createSignal(props.defaultOpen)`), including the shell-mode bash view during updates.
- Scroll/anchor positions inside the turn content jump.

The stale-read fix actually lives in `Part`'s identity snapshot + ErrorBoundary fallback, which the new regression test covers directly — the outer `keyed` wrappers are not required for it. I'd keep the turn boundary non-keyed (responsive updates) and only key where identity genuinely changes.

### P2 — `Part` component lookup became a static snapshot

`createMemo(() => PART_MAPPING[props.part.type])` was replaced by a plain read. If a part's `type` can change at runtime (placeholder → final type), the wrong renderer would stay mounted. Low probability, but the memo is free and preserves the previous contract.

### P3 — the new regression test fails under the default 5s timeout

Running `bun test test/components/message-part-error-boundary.test.ts` with the default timeout fails: the `beforeAll` vite build exceeds 5s and the run ends with `killed 1 dangling process`. It only passes through `ui/script/test.ts`'s isolated 30s window. `test:changed` or direct runs will go red.

### P4 — nits

- `vite.config.ts` runs `git status --short` synchronously on every config load and counts untracked files as dirty; `git rev-parse --short=9 HEAD` avoids the extra slice and `--no-optional-locks` avoids taking the git index lock during builds.
- `error.tsx` / `fatal-error.tsx` `Show keyed` is unnecessary — the label is a build-time constant.

I'll push a follow-up commit addressing these.

yzxoi:
Pushed the review fixes as `c82c18d14`:

- **P1**: session-turn and message-role boundaries are non-keyed again (`Show when` / `Match when`), so per-reply `message.updated` reconciles no longer tear down and rebuild the whole turn (tool-card expansion, scroll, Markdown state are preserved). The stale-read isolation stays in `Part` via the identity snapshot + ErrorBoundary fallback, covered by the regression test.
- **P2**: `Part` resolves the renderer through `createMemo` again, keeping registration reactive while the fallback still reads only the captured identity.
- **P3**: the regression test's `beforeAll` now has an explicit 60s timeout, so direct `bun test` runs pass instead of timing out at the 5s default.
- **P4**: `vite.config.ts` git probes use `GIT_OPTIONAL_LOCKS=0`; error pages drop the unnecessary `keyed`.
- Docs (develop-frontend SKILL, PRODUCT.md) updated to state the non-keyed turn boundary rule.

Verified: ui suite, app typecheck, regression test (with and without explicit timeout), and a `SYNERGY_APP_BUILD_KIND=local` production build all pass.

yzxoi (approved):

PR #1082, opened 2026-08-05
fix(file): resolve Linux watcher binding in source mode

marzich:
## Summary

- resolve the Parcel watcher package without assuming the packaged `SYNERGY_LIBC` define exists
- fall back to glibc for direct Linux source runtimes while preserving explicit musl builds
- add cross-platform package-name regression coverage

## Root cause

Source-mode development servers execute `file/watcher.ts` without the build-time `SYNERGY_LIBC` define. Reading the undeclared identifier threw before `@parcel/watcher` could subscribe. With subscription recovery enabled, this retried continuously and left file/config watching unavailable.

## Changes

| File | Change |
| --- | --- |
| `packages/synergy/src/file/watcher-binding.ts` | Resolve the platform package name with a guarded libc fallback |
| `packages/synergy/src/file/watcher.ts` | Load the binding through the resolver |
| `packages/synergy/test/file/watcher-binding.test.ts` | Cover source-mode glibc fallback, musl injection, and non-Linux names |

## Verification

- `cd packages/synergy && bun test test/file/watcher-binding.test.ts test/file/watcher-events.test.ts` - 26 pass
- `bun run quality:quick`
- live Linux source-mode verification: watcher initialized with the inotify backend without entering subscription retry


yzxoi (approved):
## Review: approve — sound, focused fix, no blocking findings

### The change
- `packages/synergy/src/file/watcher-binding.ts` (new): `FileWatcherBinding.packageName()` resolves the `@parcel/watcher-*` package name with a guarded `typeof SYNERGY_LIBC === "string"` read, so source-mode runtimes without the build-time define fall back to `glibc` instead of throwing `ReferenceError` (which previously sent the subscription-recovery loop into infinite retries). Non-Linux platforms never append a libc suffix. This matches the existing pattern in `src/browser/install.ts:481`.
- `packages/synergy/src/file/watcher.ts`: the inline dynamic `require` now delegates to the resolver; the `declare const SYNERGY_LIBC` moved into the new module.
- `packages/synergy/test/file/watcher-binding.test.ts` (new): covers the glibc fallback, musl injection, and non-Linux package names.

### On the P1 comment about static discoverability
I ran a minimal experiment comparing the old vs new require shapes with `bun build --compile --define:SYNERGY_LIBC="'glibc'"` against fake `@parcel/watcher-*` packages:
- Both forms produce **identical bundles**: Bun emits the same dynamic `__require(...)` call for both — the old template-string require was never statically inlined either, so the PR does not change discoverability.
- Running the compiled binaries with `node_modules` removed: **both** fail with `Cannot find module '@parcel/watcher-...'` — the native binding was always resolved at runtime from `node_modules`, never embedded in the binary.
- The packaged build already handles this independently of the require shape: `script/build.ts` `ensureNativeBuildPackages()` filters all `@parcel/watcher-*` dependencies and stages them into each dist package.

Conclusion: the P1 concern does not apply — behavior parity with `dev` for packaged builds is exact.

### Non-blocking
- The undefined-define path is only exercised indirectly (unbundled test runtime), but the `libc` injection parameter covers the defined-value paths.
- CI is green across all 9 checks (Test, Quality, Typecheck, Windows, Desktop, Smoke, Package Validation, Secret Scan, Workflow Validation).

**Verdict: approve.**

PR #1086, opened 2026-08-05
fix(app): preserve prompt borders and corner radius in narrow windows

TTAWDTT:
## Summary

When the desktop app is resized to a narrow window, the prompt input was incorrectly treated as a mobile layout.

As a result, the prompt lost its left and right borders and its rounded corners, making the windowed desktop layout visually inconsistent with the maximized layout.

## Changes

- Preserve the prompt's left and right borders in narrow desktop windows.
- Preserve the prompt's rounded corners when the desktop window is resized below the mobile breakpoint.
- Keep the existing edge-to-edge behavior for narrow web layouts.
- Apply a subtle border color to maintain visual separation in narrow layouts.

## Validation

- Prettier passed
- oxlint passed
- Production build passed
- Verified in both Web and Electron desktop builds

### Before

In a narrow desktop window, the prompt input loses its left and right borders and becomes square-cornered.
屏幕截图 2026-08-05 143356
### After

The prompt input preserves its side borders and rounded corners, matching the maximized desktop layout.
屏幕截图 2026-08-05 150954


yzxoi:
@synergy-agent review

yzxoi:
Follow-up: unify the narrow-layout prompt border to 18% on all edges.

While testing in an isolated instance, the narrow-window layout showed inconsistent borders: top 12%, left/right 18%, and the bottom border removed entirely (the old `max-md:border-b-0` override). This follow-up makes all four edges use the same 18% `--text-strong` border in the `max-width: 767px` media query and drops the now-redundant `max-md:border-t/max-md:border-b-0` class overrides.

Verified: narrow Web layout in the isolated dev instance shows a uniform border on all edges; Prettier and the full typecheck pass.

yzxoi (approved):

PR #1089, opened 2026-08-05
fix(app): animate collapsible surfaces

TTAWDTT:
## Summary

Several collapsible surfaces were rendered by mounting and unmounting their contents immediately, which made expand/collapse interactions feel abrupt and inconsistent.

This change introduces consistent animated transitions while preserving the existing interaction and state behavior.

## Changes

- Add a reusable disclosure transition for sidebar sections such as Recent, Home, Channels, and Background.
- Animate agenda activity group expansion and collapse.
- Animate the collapsed and expanded states of question prompts.
- Preserve accessibility state with `aria-expanded`, `aria-hidden`, and `inert`.
- Respect `prefers-reduced-motion` by disabling transitions when requested.
- Keep the Project entry as a navigation item rather than treating it as a collapsible section.

## Validation

- Prettier check passed
- oxlint passed
- Workspace typecheck passed: 11 packages
- Targeted workbench surface tests passed: 15 tests
- Manually verified sidebar expand/collapse behavior in the Web build

The full app test suite currently reports 14 failures outside the files changed by this PR, including the frontend color-token contract and Markdown preview selection tests. These are unrelated follow-up issues.


yzxoi:
@synergy-agent review

yzxoi:
## Review — fix(app): animate collapsible surfaces

Solid approach overall: replacing mount/unmount `Show` shells with always-mounted animated disclosures is the right pattern, and the a11y wiring is correct. I verified `inert` is a boolean attribute in solid-js 1.9.10's runtime (it toggles properly), the Kobalte popover is portaled to `document.body` so the new `overflow: hidden` doesn't clip it, `aria-expanded` was correctly added to the activity group header, and `prefers-reduced-motion` is handled in all three CSS files.

**Verified passing**: `tsgo -b` typecheck (packages/app), Prettier on all 6 changed files, `workbench-surface.test.ts` (15 pass), `question-prompt-style.test.ts` (Playwright, 1 pass).

### Blocking — question prompt footer/content is clipped when the prompt is tall

`packages/app/src/components/session/question-prompt.css:38-48`

`.question-prompt-shell` is a `display: flex; flex-direction: column` with `max-height: min(62vh, 540px)` and `overflow: hidden`. The PR removes `min-height: 0` from `.question-prompt-expanded` and inserts a `display: grid` shell between them. The grid container is a flex item with default `min-height: auto`, so it refuses to shrink below its content height; the `min-height: 0; overflow: hidden` on its child (`question-prompt.css:61-65`) applies to the grid *item*, not the grid *container*.

I reproduced this in headless Chromium with the real CSS (400px-wide, 300px max-height shell, 600px content):
- Pre-PR layout: shell 302px, expanded 300px, content 227px with working `overflow: auto` scroll, footer visible.
- PR layout: shell 302px, expanded **675px** (clipped by shell), content 602px with **no scroll** (`scrollHeight == clientHeight`), footer cut off below the fold.

So with several options or a long question, the answer options/review footer become unreachable. This is exactly what the removed `min-height: 0` on `.question-prompt-expanded` (deleted at `question-prompt.css:104`) was protecting.

Smallest fix — add `min-height: 0` to the grid shells:

```css
.question-prompt-collapsed-shell,
.question-prompt-expanded-shell {
  display: grid;
  grid-template-rows: 0fr;
  min-height: 0;
  ...
}
```

I verified this restores the old behavior exactly (shell 302px, content 227px scrollable, footer visible at 47px). The sidebar/agenda disclosures don't have this problem because they live in block/scroll containers, not a constrained flex column.

### Minor — skip menu can stay open after collapsing

`packages/app/src/components/session/question-prompt.tsx:237-242`

Previously, collapsing unmounted the expanded subtree, destroying the portaled Kobalte popover. Now the expanded shell stays mounted (only `inert` + `aria-hidden`), and the collapse handler (`onClick={() => setCollapsed(true)}`) doesn't reset `menuOpen()`. If the user opens the "more" menu and then clicks collapse, the portaled menu remains open and interactive over the collapsed prompt. Small fix: `setMenuOpen(false)` in the collapse handler (or an effect on `collapsed`).

### Note — PR is conflicting with `dev`

GitHub reports `CONFLICTING`, and `git merge-tree` confirms real conflicts in `packages/app/src/components/sidebar/sidebar.tsx`: `dev` added the GitHub channel group (`githubGroupOpen` / `githubChannelGroups`, PR #1084) in the same region as `SidebarDisclosure`. The resolution is straightforward (keep `SidebarDisclosure`, iterate `feishuChannelGroups()`), but when rebasing, consider also wrapping the new GitHub group in `SidebarDisclosure` so the animation story stays consistent. `sidebar.css` also has a small overlapping change (`cursor: pointer`).

**Synergy**: review complete — 1 blocking, 1 minor, 1 note

yzxoi (approved):
Approve — conflict with `dev` resolved and the earlier review findings are addressed.

## Review findings addressed

- **Question prompt scroll clipping (blocking)**: added `min-height: 0` to the collapsed/expanded disclosure shells so tall prompts stay scrollable inside the constrained `max-height: min(62vh, 540px)` shell. Regression test added (`question-prompt-style.test.ts`).
- **Skip menu lingering after collapse (minor)**: the collapse button now resets `menuOpen`, so the portal'd menu no longer stays interactive when the prompt is collapsed.
- **Dead class (nit)**: removed the unused `sb-project-disclosure` class.
- **Countdown**: kept both countdown instances active (`active={true}`); gating `active` on visibility would freeze the timer while collapsed and show a stale remaining time on expand — worse than the harmless dual tick.

## Conflict resolution (merged latest `dev`)

- Kept `dev`'s channelType-based grouping (`feishuChannelGroups()` / `githubChannelGroups()`) from #1084, and the GitHub channel group is now also wrapped in `SidebarDisclosure` so it animates consistently with the Feishu group.
- Kept this PR's `SidebarDisclosure` animation around Recent / Home / Channel / Background / projects / partner groups, including the `inert` + `aria-hidden` handling and `prefers-reduced-motion` support.
- Dropped the obsolete standalone GitHub `RootNavSection` that `dev` replaced with the in-channel GitHub group.

## Verification

- `bun test` (question-prompt style + new disclosure-layout regression, sidebar, workbench-surface): **86 pass**
- `tsgo -b` typecheck, `oxlint --deny-warnings`, `prettier --check`, `sherif` monorepo check: all clean

PR #1109, opened 2026-08-08
fix(channel): do not suppress outbound bridge for providers without terminal streaming delivery

yzxoi:
# fix(channel): do not suppress outbound bridge for providers without terminal streaming delivery

## Root cause

`c0562ed77` introduced foreground registration so the outbound bridge skips replies already delivered by a streaming card, and persists `channelOutboundSent` after the card closes. It assumed **every** foreground session's streaming session owns the terminal reply — but that is only true for providers whose streaming `close()` actually posts the final text (Feishu cards, text fallback).

Providers with a **no-op streaming session** (GitHub, which relies entirely on the outbound bridge to post comments) were registered anyway:

1. `ChannelOutbound.beginForeground(...)` → the bridge skipped the terminal assistant message
2. `channelOutboundSent: true` was persisted on the assistant message → even after `endForeground`, the durable marker made the bridge skip it forever

Result: GitHub comments were never posted, while the status reactions (👀/🚀, which fire directly in the execution path) still worked — the bot appeared to run but never replied. Empirically reproduced on the deployed instance: sessions completed, reactions appeared, zero comments on GitHub.

## Fix

Add an optional `ownsTerminalDelivery()` capability to `StreamingSession`, and only `beginForeground` / persist `channelOutboundSent` when it returns `true`:

- **Feishu streaming card** → `true` (close() finalizes the card with the answer)
- **Feishu non-streaming send session** → `true` (close() sends the text)
- **Channel text fallback session** → `true` (close() posts via replyMessage)
- **GitHub `NonStreamingSession`** → `false` (default) — the outbound bridge stays the delivery path

## Files

- `packages/synergy/src/channel/types.ts` — `StreamingSession.ownsTerminalDelivery?()`
- `packages/synergy/src/channel/index.ts` — gate `beginForeground` and the `channelOutboundSent` persist on `ownsTerminalDelivery()`
- `packages/synergy/src/channel/provider/feishu/index.ts`, `streaming-card.ts` — declare ownership
- GitHub provider intentionally unchanged (no-op streaming session defaults to `false`)

## Verification

- `bun run typecheck` — clean
- `bun test test/channel/` — 448 pass / 0 fail

### DIFF
--- a/packages/synergy/src/channel/index.ts
+++ b/packages/synergy/src/channel/index.ts
@@ -759,12 +759,6 @@ export namespace Channel {
               })
               void reactionController.setQueued()
 
-              // The foreground streaming card owns this root's terminal reply:
-              // register the root so the outbound bridge skips it and the
-              // answer is not delivered twice. Queued (busy/recovered) roots
-              // are never registered and keep the bridge as their delivery path.
-              ChannelOutbound.beginForeground(sessionID, delivery.messageID)
-
               let streaming = createStreamingSession({
                 accountId: ctx.accountId,
                 chatId: ctx.chatId,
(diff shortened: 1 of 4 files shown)


yzxoi:
@synergy-agent review

PR #1111, opened 2026-08-08
feat(plugin): add manual refresh for the plugin marketplace

EricSanchezok:
## Summary

The plugin Marketplace sidebar refreshed the official plugin catalog only on the hourly cache TTL, so new plugin versions from the `synergy-plugins` registry were invisible for up to an hour. This adds a manual refresh action that bypasses the cache immediately.

- **Backend**: new `POST /api/registry/refresh` endpoint backed by `PluginMarketplaceRegistry.refreshNow()` — force-fetches the remote registry index (bypassing `cacheTtlMs`), atomically rewrites `registry.json`, and drops the entry cache so detail views re-fetch fresh data. On failure the stale caches are preserved and the route returns the existing structured `503` response.
- **Frontend**: a refresh button in the Marketplace header (all views) calls the endpoint through the generated SDK, shows a spinner while running, toasts on success/failure, and refetches the search and installed lists so "Update available" states appear immediately.
- **SDK/OpenAPI**: regenerated (`client.registry.refresh()`).
- **i18n + docs**: new copy for en / zh-CN / pseudo; `docs/plugins/marketplace.md` documents the manual refresh behavior.

## Verification

- `packages/synergy`: `bun test test/server/plugin-registry-routes.test.ts` — 16 pass (3 new tests: TTL bypass + entry cache drop, 503 with stale cache preserved, concurrent dedupe)
- `packages/app`: `bun run typecheck` + `bun run test` pass
- `bun run localization:check` pass
- `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run quality:quick` pass

### DIFF
--- a/docs/plugins/marketplace.md
+++ b/docs/plugins/marketplace.md
@@ -24,6 +24,8 @@ Registry identity, generated manifest ID, approval ID, lockfile key, signature p
 
 When an official registry read cannot reach its upstream source and no cached data can satisfy the request, the server returns a structured `503 Service Unavailable` response. The Marketplace keeps installed or cached plugin details usable when possible, shows a registry-unavailable state instead of a generic application error, and offers an explicit retry action.
 
+The Marketplace header offers a manual refresh action that re-fetches the official registry index immediately, bypassing the hourly cache TTL, and drops cached entry details so the next detail view is fresh. When the refresh cannot reach the upstream registry, the previous cached data stays intact and the action reports an error instead of clearing the list.
+
 ## Publish Flow
 
 ```bash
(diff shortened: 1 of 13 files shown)


yzxoi (approved):
Reviewed locally (pulled head into a worktree and ran the tests): `test/server/plugin-registry-routes.test.ts` 16 pass / 0 fail, including the 3 new tests (TTL bypass + entry-cache drop, 503 with stale cache preserved, concurrent dedupe).

The implementation is clean: failure path preserves stale caches, the dedupe map is cleaned in `finally`, and SDK/OpenAPI/i18n/docs are all regenerated consistently. One minor non-blocking observation: when the marketplace is disabled, `refreshNow()` returns `refreshedAt: null` and the UI gives no feedback (no success or error toast) — the button just silently refreshes the list. Approving.

PR #1124, opened 2026-08-10
fix(session): preserve structured stream errors across Agent worker IPC

### LINKED ISSUE #1122: Structured stream errors collapse to [object Object] across Agent worker IPC
## What happened?

Structured provider errors emitted through an AI SDK stream can be plain JSON objects rather than JavaScript `Error` instances. When such an event crosses the Agent worker IPC boundary, `AgentTurnProtocol.serializeError()` currently handles the non-`Error` branch with:

```ts
message: String(error)
```

For a structured object this becomes `[object Object]`. The host then rehydrates it as `Error("[object Object]")`, `SessionProcessor` throws it for the stream `error` event, and `MessageV2.fromError()` persists an `UnknownError` whose only useful field is:

```text
Error: [object Object]
```

The original provider `message`, `code`, `type`, nested error fields, and retry metadata are irreversibly lost. Besides producing an unactionable UI error, this prevents reliable retry classification and can turn retryable provider failures into terminal `UnknownError` turns.

The relevant path is:

1. `packages/synergy/src/session/agent-turn/protocol.ts` projects a stream `error` event.
2. `encodeEvents()` calls `serializeError()` for the event payload.
3. A plain object takes the `String(error)` branch.
4. `decodeEvents()` reconstructs an `Error` containing `[object Object]`.
5. `packages/synergy/src/session/processor.ts` throws `value.error` for the stream event.
6. `packages/synergy/src/session/message-v2.ts` stores it as `UnknownError`.

This remains reproducible on current `dev` (`f8b83b2ca`). Existing Agent turn protocol tests cover `Error`, `APICallError`, `DOMException`, and structured errors implemented as `Error` subclasses, but not plain object stream errors.

Observed runtime characteristics across multiple occurrences:

- the provider stream started successfully;
- no text or tool output was produced;
- the stream emitted an error after approximately 5-8 seconds;
- adjacent requests using the same provider/model and comparable or larger request payloads completed successfully;
- every occurrence lost its diagnostic payload and became the same `[object Object]` message.

## Expected behavior

Plain object errors crossing the Agent worker protocol should retain a bounded, useful diagnostic representation.

At minimum:

- prefer a string `message` or nested `error.message` when present;
- preserve bounded `name`, `type`, `code`, status, retryability, and structured data where safe;
- use bounded safe JSON as the fallback instead of `String(object)`;
- preserve existing size limits and redaction rules;
- never fail serialization on circular, unsupported, or oversized values;
- allow downstream retry classification to inspect the recovered fields.

For example, this source event:

```ts
{
  type: "error",
  error: {
    type: "server_error",
    code: "upstream_unavailable",
    message: "The upstream service is temporarily unavailable",
  },
}
```

should remain actionable after `encodeEvents()` and `decodeEvents()`, rather than becoming `Error("[object Object]")`.

## Steps to reproduce

Add a protocol-level regression test using the existing public encode/decode boundary:

```ts
const [decoded] = AgentTurnProtocol.decodeEvents(
  AgentTurnProtocol.encodeEvents([
    {
      type: "error",
      error: {
        type: "server_error",
        code: "upstream_unavailable",
        message: "temporarily unavailable",
      },
    },
  ]),
)

expect(decoded.type).toBe("error")
expect(decoded.error).toBeInstanceOf(Error)
expect(decoded.error.message).toContain("temporarily unavailable")
expect(decoded.error.message).not.toContain("[object Object]")
```

Current behavior produces an error whose message is `[object Object]`.

Additional focused coverage should include:

- a nested `{ error: { message, code } }` payload;
- a plain object without a `message` field;
- circular and oversized data;
- retention of retry-relevant fields without exposing request credentials.

## Environment

- OS: Ubuntu Linux
- Runtime: Bun 1.3.14
- Synergy: source-mode development server
- Verified against `origin/dev` at `f8b83b2ca`

## Relevant logs or screenshots

Representative sanitized output:

```text
service=session.processor error=[object Object] process
finish=error error={"name":"UnknownError","data":{"message":"Error: [object Object]"}}
```

The raw provider payload cannot be recovered after the current serialization step, which is the core diagnostic impact of this bug.

yzxoi:
Closes #1122

## Summary

- Structured provider errors emitted through an AI SDK stream can be plain JSON objects rather than `Error` instances. When such an event crossed the Agent worker IPC boundary, `AgentTurnProtocol.serializeError()` handled the non-`Error` branch with `String(error)`, collapsing it to `[object Object]`. The host then rehydrated `Error("[object Object]")`, `SessionProcessor` threw it, and `MessageV2.fromError()` persisted an `UnknownError` whose only useful field was `Error: [object Object]` — losing provider `message`, `code`, `type`, nested error fields, and retry metadata, and turning retryable provider failures into terminal `UnknownError` turns.

## Changes

- `packages/synergy/src/session/agent-turn/protocol.ts`
  - `serializeError()` now extracts structured fields (`message`, `code`/`type`, `statusCode`, `isRetryable`) from plain-object errors, including nested `{ error: { ... } }` payloads.
  - Fallback is bounded safe JSON (never `String(object)`), and circular/oversized values never fail serialization.
  - Preserves existing size limits and reuses `boundedSerializable()` for the `data` payload.
- `packages/synergy/src/session/message-v2.ts`
  - `fromError()` now recognizes protocol-rehydrated structured errors (identified by the `isRetryable` boolean marker that plain native `Error`s never carry) and converts them into `APIError`, so `SessionRetry.retryable()` can classify and retry them instead of persisting `UnknownError`.

## Tests

- `test/session/agent-turn-protocol.test.ts`: plain-object stream error regression (issue reproduction), nested payload, retry metadata retention, bounded-JSON fallback, circular safety, oversized data bounding.
- `test/session/retry.test.ts`: downstream `fromError()` + `SessionRetry.retryable()` closed-loop coverage, plus unknown-error preservation when no retry metadata survives.

## Verification

- `bun test test/session/agent-turn-protocol.test.ts` — 24 pass
- `bun test test/session/retry.test.ts` — 18 pass
- `bun run test:changed` — 3801 pass (2 unrelated pre-existing timeout flakes in migration tests, pass in isolation)
- `bun run typecheck`, `bun run lint`, `bunx prettier --check` — clean

### DIFF
--- a/packages/synergy/src/session/agent-turn/protocol.ts
+++ b/packages/synergy/src/session/agent-turn/protocol.ts
@@ -389,9 +389,23 @@ export namespace AgentTurnProtocol {
 
   export function serializeError(error: unknown): SerializedError {
     if (!(error instanceof Error)) {
+      const object = isPlainRecord(error) ? error : undefined
+      const nested = isPlainRecord(object?.error) ? object.error : undefined
+      const message = firstString(
+        object?.message,
+        nested?.message,
+        isPlainRecord(nested?.error) ? nested.error.message : undefined,
+      )
+      const code = firstString(object?.code, nested?.code, object?.type, nested?.type)
+      const statusCode = firstInteger(object?.statusCode, nested?.statusCode)
+      const isRetryable = firstBoolean(object?.isRetryable, nested?.isRetryable)
       return {
-        name: "Error",
-        message: String(error).slice(0, ERROR_MESSAGE_MAX_CHARS),
+        name: firstString(object?.name, nested?.name) ?? "Error",
+        message: (message ?? safeStringify(error)).slice(0, ERROR_MESSAGE_MAX_CHARS),
+        code: code?.slice(0, ERROR_MESSAGE_MAX_CHARS),
+        statusCode,
+        isRetryable,
+        data: object === undefined ? undefined : boundedSerializable(object, ERROR_DATA_MAX_BYTES),
       }
     }
     const code =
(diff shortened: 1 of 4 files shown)


yzxoi:
Duplicated #1123

PR #1125, opened 2026-08-10
fix(tool): expose github_deliver_fix to the GitHub channel agent

yzxoi:
# fix(tool): expose github_deliver_fix to the GitHub channel agent

## Root cause

The `github_deliver_fix` tool was never visible to the `github-channel-agent`, so fix delivery never happened — the agent could only commit locally:

1. **Permission gate**: `github-channel-agent` runs under a strict whitelist (`"*": "deny"` + explicit allows). `PermissionNext.disabled()` disables every tool not explicitly allowed — empirically verified that `github_deliver_fix` (plus `expand_tools`/`search_tools`) landed in the disabled set. The agent had no activation path at all.
2. **Exposure gate**: the tool's exposure was `mode: "search"`, which keeps it invisible until activated via `expand_tools`/`search_tools` — both of which were themselves disabled by the whitelist.

## Fix

- `github-deliver-fix.ts` — expose the tool as `resident` so it is visible without activation. The tool already validates that the session is bound to a GitHub channel thread and errors otherwise (`GITHUB_TOOL_NOT_IN_CHANNEL_SESSION`).
- `tool-mode-policy.ts` — restrict `github_deliver_fix` to Channel sessions (same pattern as `response_card`), so the resident exposure does not leak into non-channel sessions.
- `builtin-internal.ts` — add `github_deliver_fix: "allow"` to the `github-channel-agent` whitelist so the permission system does not disable it.

## Verification

- `PermissionNext.disabled()` no longer contains `github_deliver_fix` for the agent's ruleset
- `bun run typecheck` — 11/11 packages clean
- `bun test test/session/tool-mode-policy.test.ts test/tool/github-deliver-fix.test.ts` — 13 pass / 0 fail (new coverage: channel-only visibility of `github_deliver_fix`)

### DIFF
--- a/docs/architecture/github-channel.md
+++ b/docs/architecture/github-channel.md
@@ -119,10 +119,11 @@ When the agent fixes an issue, the fix is delivered as a pull request by default
 
 1. Verifies the session is bound to a GitHub channel thread and the workspace checkout exists.
 2. Resolves the repository default branch as the PR base.
-3. Verifies the local branch exists and has commits ahead of the base.
-4. **When the thread is a pull request**, pushes the fix to that PR's head branch first, updating the PR in place (Codex-style behavior). Same-repository PRs push directly; fork PRs push to the fork only when the App has an installation there. If the push is impossible (fork without App access, protected or moved head branch), it falls back to the next step.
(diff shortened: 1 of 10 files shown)


yzxoi:
@codex review

PR #1138, opened 2026-08-13
fix(ui): reveal tool error details inline in activity steps

yzxoi:
## Summary

Tool call failures in the balanced activity display previously required two clicks to inspect: expand the collapsed activity step, then open a details dialog. This changes the error card to an inline collapsible rendered in the compaction failure card style, so a single step click reveals the full error (monospace details + copy action) directly in the timeline.

## Changes

- `packages/ui/src/components/error-card.tsx` / `error-card.css`: replace the dialog-based error card with an inline `Collapsible`; header shows the error preview, expanded body shows the full error, optional tool input, and a copy-details button. `compact` mode defaults to collapsed.
- `packages/ui/test/components/activity-trace.dom.test.ts`: add DOM coverage for a failed step — collapsed by default, full error visible after one click.
- Locale catalogs: drop the retired `ui.errorCard.viewDetails` / `ui.errorCard.errorMessage` entries.

## Verification

- `bun test test/components/activity-trace.dom.test.ts` (packages/ui): 20 pass
- `bun run typecheck` (packages/ui): pass
- `bun run i18n:check` (packages/app): pass
- `bun run quality:quick` (root): pass
- Manually verified in an isolated dev instance: failed tool step collapses by default; one click reveals the full error inline.

### DIFF
--- a/packages/app/src/locales/en/messages.po
+++ b/packages/app/src/locales/en/messages.po
@@ -19173,20 +19173,12 @@ msgid "ui.errorCard.detailsTitle"
 msgstr "Error details"
 
 #. js-lingui-explicit-id
-msgid "ui.errorCard.errorMessage"
-msgstr "Error message"
-
-#. js-lingui-explicit-id
 msgid "ui.errorCard.toolInput"
 msgstr "Tool input"
 
 #. js-lingui-explicit-id
 msgid "ui.errorCard.unknownError"
 msgstr "Unknown error"
-
-#. js-lingui-explicit-id
-msgid "ui.errorCard.viewDetails"
-msgstr "View details"
 
 #. js-lingui-explicit-id
 msgid "ui.imagePreview.close"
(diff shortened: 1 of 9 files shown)


yzxoi:
Thanks for the review — all three suggestions are valid and addressed in `a050577b9`:

1. **Keep ordinary tool errors compact by default** — `ErrorCard` now defaults to collapsed (`defaultOpen` instead of `compact`). The Full timeline path (`tool-result-body.tsx` via `message-part.tsx`) renders the one-line preview; only the already-expanded Activity Trace step path opts in with `defaultOpen` so a step click still reveals the full error inline.

2. **Resolve clipboard labels reactively** — `copyLabel` / `copiedLabel` / `failureDescription` are now getters (same pattern as `CompactionCard`), so copy feedback follows locale changes while the card stays mounted.

3. **Disable copy-feedback scaling for reduced motion** — the `prefers-reduced-motion` override now covers the copy button and its icon (`transition: none; animation: none`).

Verified: `packages/ui` typecheck passes, 115 related tests pass (including the failed-step DOM coverage), and the prior commit's gates remain green.

PR #1140, opened 2026-08-13
fix(ui): show approval audit icon on balanced activity steps

yzxoi:
## Summary

In balanced activity display, tool calls are collapsed into step rows that only render the family icon — the approval audit indicator shown in full mode (e.g. the smart-allow / auto-allowed orbit icon in autonomous mode) was dropped. This derives the audit icon from the step's approval metadata using the same `getApprovalAudit` rules as the full path and renders it inline before the family icon, so auto-allowed tools stay visibly annotated.

## Changes

- `packages/ui/src/components/activity-trace.tsx`: read `part.state.metadata.approval` per step, compute the audit via `getApprovalAudit`, and render `data-component="tool-audit-icon"` (with the approval tooltip as `title`) ahead of the family icon when an icon resolves.
- `packages/ui/src/components/activity-trace.css`: inline (non-absolute) audit icon layout inside the step row, with hover affordance.
- `packages/ui/test/components/activity-trace.dom.test.ts`: DOM coverage asserting the audit icon renders for an `auto_allowed` / autonomous step.

## Verification

- `bun run typecheck` (packages/ui): pass
- `bun test test/components/activity-trace.dom.test.ts test/approval-audit.test.ts ...`: 126 pass
- `bun run format:check` + `bun run lint` (root): pass

### DIFF
--- a/packages/ui/src/components/activity-trace.css
+++ b/packages/ui/src/components/activity-trace.css
@@ -11,6 +11,10 @@
 }
 
 [data-component="activity-trace"] {
+  [data-component="collapsible"] {
+    overflow: visible;
+  }
+
   [data-slot="activity-step-trigger"] {
     position: relative;
     display: flex;
--- a/packages/ui/src/components/activity-trace.css
+++ b/packages/ui/src/components/activity-trace.css
@@ -68,13 +68,15 @@
   }
 
   [data-slot="activity-step-audit-icon"] {
-    position: static;
-    display: inline-flex;
-    flex: 0 0 auto;
+    position: absolute;
+    left: -28px;
+    top: 50%;
+    transform: translateY(-50%);
+    display: flex;
     align-items: center;
     justify-content: center;
-    width: 18px;
-    height: 18px;
+    width: 24px;
+    height: 24px;
     cursor: help;
     border-radius: var(--radius-full);
     animation: none;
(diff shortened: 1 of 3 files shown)


yzxoi:
Thanks for the review — I checked all four suggestions against the current code and addressed the two that still applied:

1. **Reset the audit icon's inherited absolute positioning** — No longer applies. The final CSS explicitly sets `position: absolute; left: -28px; top: 50%` (anchored to the step row, outside it), verified live in the isolated instance: icon renders at the left edge of the step, inside the viewport.

2. **Localize the audit tooltip with the active catalog** — Applied. `getApprovalAudit(approval(), i18n())` now receives the active `i18n` from `useLingui()`, so the approval/risk labels and smart-allow line follow the interface locale like the rest of the row.

3. **Disable the inherited audit animation for reduced motion** — No longer applies. The activity-row variant sets `animation: none` unconditionally, so the inherited `audit-icon-enter` slide/fade never runs, including for reduced-motion users.

4. **Make the audit details reachable without hover** — Applied. The audit icon moved out of the collapsible trigger (an element inside a button cannot be focused) into the shared `Tooltip` primitive with `tabindex="0"`, `role="img"`, and an `aria-label` on the trigger, matching the full-mode tool card pattern — keyboard and touch users can now reach the approval details.

Verified: `packages/ui` typecheck passes; 34 related tests pass; live check in the isolated instance confirms the icon stays positioned outside the step row with the tooltip reachable.

PR #1142, opened 2026-08-13
fix(enforcement): classify read-only gh api as shell_read and filter slash-only pseudo paths

yzxoi:
## Summary

Fixes an enforcement false-positive that blocked a **read-only** `gh api` command under the `autonomous` control profile.

- **`gh api` classification** (`packages/synergy/src/enforcement/shell-safety.ts`): `gh api` is an authenticated HTTP client — default method is GET, but `-f`/`-F`/`--raw-field`/`--field`/`--input` auto-switch the request to POST and `-X`/`--method` can select any verb. Only statically confirmed GET/HEAD requests are now classified as `shell_read`; fields without an explicit GET method, `--input`, GraphQL, and all explicit write methods remain `shell_remote_write`, so `autonomous` still denies real remote mutations.
- **Slash-only pseudo-path filter** (`packages/synergy/src/enforcement/gate.ts`): `extractAbsolutePaths` now filters slash-only candidates (`//`+). jq's null-coalescing operator (`//`) was being extracted as an external path, producing a non-bypassable `file_external_write` capability that blocked even read-only `gh api --jq '... // ...'` commands in `autonomous`.
- **Docs**: updated the GitHub CLI permission matrix in `.synergy/skill/git-guide/SKILL.md` with the new `gh api` rows.

## Verification

- `bun test test/enforcement/shell-safety.test.ts test/enforcement/gate.test.ts` — 956 pass, 0 fail
- `bun test test/control-profile/autonomous.test.ts test/enforcement/classify.test.ts test/permission` — 289 pass, 0 fail
- `bun run typecheck` (packages/synergy) — clean
- `bunx prettier --check` on changed files — clean
- `bun run quality:quick` (repo root) — exit 0

## Test coverage added

- `gh api` default GET → `shell_read` (including the original jq + `2>&1` repro)
- `gh api -X GET`/`-X HEAD` with fields → `shell_read`
- `gh api -f`/`-F`/`--input` (auto-POST) → `shell_remote_write`
- `gh api -X POST/PATCH/DELETE` → `shell_remote_write`
- `gh api graphql` → `shell_remote_write` (mutations cannot be ruled out statically)
- Gate-level: jq `//` no longer yields `file_external_write`; autonomous allows read-only `gh api`, denies mutating `gh api`

### DIFF
--- a/.synergy/skill/git-guide/SKILL.md
+++ b/.synergy/skill/git-guide/SKILL.md
@@ -120,6 +120,11 @@ The permission system classifies `gh` commands before applying the active contro
 | `gh issue create/comment`            | `shell_remote_publish` | ⚠️ ask   | ✅ allow   | ✅ allow    |
 | `gh issue edit/close/reopen`         | `shell_remote_write`   | ⚠️ ask   | ❌ deny    | ✅ allow    |
 | `gh pr merge/close/reopen`           | `shell_destructive`    | ⚠️ ask   | ❌ deny    | ✅ allow    |
+| `gh api <endpoint>` (GET, no fields) | `shell_read`           | ✅ allow | ✅ allow   | ✅ allow    |
+| `gh api -X GET` / `-X HEAD`          | `shell_read`           | ✅ allow | ✅ allow   | ✅ allow    |
+| `gh api -f/-F/--input` (auto-POST)   | `shell_remote_write`   | ⚠️ ask   | ❌ deny    | ✅ allow    |
+| `gh api -X POST/PATCH/PUT/DELETE`    | `shell_remote_write`   | ⚠️ ask   | ❌ deny    | ✅ allow    |
+| `gh api graphql` (mutation possible) | `shell_remote_write`   | ⚠️ ask   | ❌ deny    | ✅ allow    |
 
 Checkout type does not change ordinary `shell_remote_publish` into `shell_remote_write`. Unknown write-capable `gh` subcommands default to `shell_remote_write`. Full Access silently allows permission-system capabilities but does not override task authorization, protected-branch rules, GitHub permissions, validation failures, or network/runtime errors.
 
--- a/packages/synergy/src/enforcement/gate.ts
+++ b/packages/synergy/src/enforcement/gate.ts
@@ -313,8 +313,19 @@ function isDestructive(command: string): string | null {
 function extractAbsolutePaths(command: string): string[] {
   const paths: string[] = []
   const pathPattern = /(?:\s|"|'|>|<|^|\|)(\/[^\s"'|;&]+)/g
+  // Mask gh api jq expression values before extracting absolute paths:
+  // jq syntax such as the null-coalescing operator (//) is not a filesystem
+  // path. Masking only jq arguments (--jq/-q) inside gh api invocations keeps
+  // genuine slash-only targets (e.g. `rsync file //`, which writes to the
+  // filesystem root) classified as external writes instead of dropping them
+  // globally.
+  let masked = command
+  if (/\bgh\s+api\b/.test(command)) {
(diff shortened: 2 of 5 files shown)


yzxoi:
Both review findings are valid security regressions in the first revision; both are fixed in 394117c430.

**P1 — attached gh api write flags (shell-safety.ts):** gh uses Go pflag, so write flags can be attached (`--method=DELETE`, `-XDELETE`, `-Fbody=hi`, `--field=body=hi`, `--input=file.json`); exact-token matching missed these and classified them `shell_read`. The classifier now parses both separated and attached option values, and only statically confirmed GET/HEAD (separated or attached) is read-only. Added tests: attached write forms → `shell_remote_write`, attached `-XGET`/`-XHEAD`/`--method=GET` → `shell_read`.

**P2 — global slash-only path filter (gate.ts):** dropping `//` globally removed genuine filesystem targets such as `rsync file //` (which writes to the filesystem root), eliminating the non-bypassable `file_external_write` capability. The filter is now scoped: only `--jq`/`-q` expression values inside `gh api` invocations are masked before absolute-path extraction, so jq's `//` operator is not treated as a path while real slash-only destinations keep their external-write classification. Added a gate test asserting `rsync file //` keeps `file_external_write` and is denied under autonomous, and that the original read-only `gh api --jq '... // ...'` command stays free of `file_external_write`.

Verification: shell-safety + gate suites 961 pass, 0 fail; typecheck and prettier clean.

yzxoi:
Thanks for the thorough review. Verified each point against the current HEAD (`394117c430`), which already contains the fix revision for the earlier P1/P2 inline comments.

**Blocking — attached/`=` spellings: fixed in `394117c430`**

The classifier now parses both separated and attached pflag forms. Runtime probe on HEAD:

| Command                                   | Classification       |
| ----------------------------------------- | -------------------- |
| `gh api repos/x -XPOST`                   | `shell_remote_write` |
| `gh api repos/x --method=DELETE`          | `shell_remote_write` |
| `gh api repos/x -fbody=hi`                | `shell_remote_write` |
| `gh api repos/x -Fstate=closed`           | `shell_remote_write` |
| `gh api repos/x --field=state=closed`     | `shell_remote_write` |
| `gh api repos/x --raw-field=state=closed` | `shell_remote_write` |
| `gh api repos/x --input=file.json`        | `shell_remote_write` |

Gate-level: `gh api repos/foo/bar/issues/1/comments --field=body=hi` → **deny** under `autonomous`. Tests for all attached forms were added in the fix commit.

**Blocking — `-i file.json` row: not applicable**

`-i` is not the `--input` shorthand. In gh's own source (`pkg/cmd/api/api.go`):

- `-i, --include` is a read-only bool (`ShowResponseHeaders`)
- `--input` is registered without a short flag (`StringVar(&opts.RequestInputFile, "input", ...)`)

Additionally, `cobra.ExactArgs(1)` makes `gh api <endpoint> -i file.json` fail as an extra positional argument, so it has no write capability.

**Suggestion — jq `//` operator beyond bare token: gh api case covered**

`extractAbsolutePaths` now masks the entire `--jq`/`-q` argument value inside `gh api` invocations before path extraction. Probe: `gh api ... --jq '.line //.original_line'` (no-space form) → allow under autonomous, no `file_external_write`.

The residual false positive for non-gh jq commands (`bun x jq '.a //.b'` → `file_external_write`) predates this PR — the original `/^\/\/+$/` filter never matched `//.b` either. I'd rather not broaden the jq-context heuristic in this PR because of the `//host/path` POSIX-path collision you flagged; happy to do it as a focused follow-up with a dedicated jq-context rule.

Verification on HEAD: shell-safety + gate suites 961 pass, 0 fail; typecheck and prettier clean.

PR #1158, opened 2026-08-14
fix(session): re-execute settled tool tasks on replayed dispatch

yzxoi:
## Summary

Fixes the CI flaky in `SessionProcessor` settlement tests that also affected the merged #1152 run (`does not persist a duplicate tool part when a settled call is replayed before tool-error` and `records LLM invalid_arguments failures that never enter the executor`).

### Root cause

`ToolScheduler` is a module-level singleton whose `dispatch()` deduplicates by key (`sessionID/generation/messageID/callID/executor/attempt`) and kept **settled tasks in the map for 30 minutes** (`sweepTerminal` cutoff). A later dispatch with the same key — retry, replay, or a fresh `SessionProcessor` instance (each test run reuses the fixed `ses_test`/call IDs) — returned the **stale settled promise** instead of running the tool again. The tool slot then never resolved, producing:

- `Tool execution did not return a final result` on the new part
- duplicate/empty settlement paths under full-shard CI load

Locally this reproduced deterministically: `bun test test/session/processor.test.ts --rerun-each=50` failed 98 times before the fix (execution layering group), and passes **2500/2500** after.

### Change

`packages/synergy/src/session/tool-scheduler.ts` — stop deduplicating a key once its task settles (`this.tasks.delete(key)` on settlement). In-flight dedup is preserved: duplicate dispatches of a still-running call still collapse to one execution (existing test `deduplicates a replayed call before invoking the executable tool` keeps passing).

This is also a production bug: a replayed/retried tool call with the same key would silently receive a stale result instead of executing.

### Tests

- New regression test: `re-executes a settled call when the same key is dispatched again` (asserts 2 executions)
- `tool-scheduler.test.ts`: 11 pass
- `processor.test.ts --rerun-each=50`: 2500 pass / 0 fail (was 98 fail)
- `test/session/` full suite: 1030 pass (2 pre-existing `SessionInbox` timeouts, verified present without this change)
- CI-like shard 4/4 full run: 2213 pass; 5 LibraryDB vector flakes verified pre-existing (pass in isolation)
- typecheck / prettier / lint: clean

### DIFF
--- a/packages/app/test/components/app-shell/mobile-drawer-root.test.tsx
+++ b/packages/app/test/components/app-shell/mobile-drawer-root.test.tsx
@@ -108,8 +108,8 @@ beforeAll(async () => {
     plugins: [solidPlugin()],
     server: {
       host: "127.0.0.1",
-      port: 0,
-      strictPort: false,
+      port: 5199,
+      strictPort: true,
       fs: { allow: [path.resolve(import.meta.dir, "../../../..")] },
     },
   })
--- a/packages/app/test/components/menu-field/menu-field.test.ts
+++ b/packages/app/test/components/menu-field/menu-field.test.ts
@@ -57,8 +57,8 @@ beforeAll(async () => {
     plugins: [solidPlugin()],
     server: {
       host: "127.0.0.1",
-      port: 0,
-      strictPort: false,
+      port: 5202,
+      strictPort: true,
       fs: { allow: [path.resolve(import.meta.dir, "../../..")] },
     },
   })
--- a/packages/app/test/components/settings/panels/BossModePanel.test.ts
+++ b/packages/app/test/components/settings/panels/BossModePanel.test.ts
@@ -69,8 +69,8 @@ beforeAll(async () => {
     plugins: [solidPlugin()],
     server: {
       host: "127.0.0.1",
-      port: 0,
-      strictPort: false,
+      port: 5203,
+      strictPort: true,
       fs: { allow: [path.resolve(import.meta.dir, "../../../..")] },
     },
   })
--- a/packages/app/test/components/settings/settings-dialog-dismiss.test.tsx
+++ b/packages/app/test/components/settings/settings-dialog-dismiss.test.tsx
@@ -71,8 +71,8 @@ beforeAll(async () => {
     plugins: [solidPlugin()],
     server: {
       host: "127.0.0.1",
-      port: 0,
-      strictPort: false,
+      port: 5204,
+      strictPort: true,
       fs: { allow: [path.resolve(import.meta.dir, "../../../..")] },
     },
   })
--- a/packages/synergy/src/session/tool-scheduler.ts
+++ b/packages/synergy/src/session/tool-scheduler.ts
@@ -246,7 +246,6 @@ export class ToolTaskScheduler {
         const remaining = (this.activeByExecutor.get(executor) ?? 1) - 1
         if (remaining > 0) this.activeByExecutor.set(executor, remaining)
         else this.activeByExecutor.delete(executor)
-        this.activeControllers.delete(task.key)
         this.drain()
       })
     }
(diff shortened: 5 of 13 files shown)


synergy-agent (commented):
Re-review of head `e986794e` (3 commits: `50564ed7`, `e406d19d6`, `e986794e0`; base `a4bf64bd9`). **Conclusion: approve — both tool-scheduler fixes are correct and the UI test change is sound. No blocking findings, no actionable non-blocking findings.**

## Commit 1 — `50564ed7` re-execute settled tool tasks on replayed dispatch

Diagnosis verified against the code: `dispatch()` dedupes by key (`tool-scheduler.ts:81-83`) and, before this PR, a settled promise stayed in `tasks` until the 30-minute `sweepTerminal()` (pre-PR `:375-384`), so a same-key replay returned the stale terminal result without invoking the tool. The fix deletes the key from `tasks` on settlement, identity-guarded (`tool-scheduler.ts:99-101`) so it never removes a newer task, and still pushes to `terminalTasks` for retention/eviction accounting. In-flight dedup is untouched — the pre-existing "deduplicates a replayed call before invoking the executable tool" test still passes. `sweepTerminal`'s identity check (`:376`) simply never matches now, which is harmless.

Regression test `tool-scheduler.test.ts` (`re-executes a settled call when the same key is dispatched again`) correctly asserts two executions for two sequential dispatches of the same key.

## Commit 2 — `e406d19d6` replay-safe controller cleanup

This directly addresses the codex P2 on the old head: `run().finally()` previously cleared `activeControllers.delete(task.key)` unconditionally (`drain()` at old `:243`, `run()` finally at old `:323`). With commit 1 releasing the key on settlement, a same-key replay can install a newer controller before the old invocation finishes its finally block, so the stale cleanup would remove the replay's controller and `stop()` could no longer abort it. The fix guards both deletions by ownership identity (`tool-scheduler.ts:322-326`) and removes the unconditional deletion in `drain()` (`:243`), since `run()`'s finally now owns that cleanup. This matches the codex-recommended remedy ("make that cleanup conditional on controller identity"). The `stop()` path (`:210-211`) iterates whatever maps contain, so it can still abort the replay's controller — the race the codex comment described is closed.

One minor residual (non-blocking, pre-existing): in the `finally` after an aborted tool, the delete is skipped when ownership changed, but the replay's controller keeps running to completion (its task is still active and correctly tracked). That is the intended new behavior — the replay is a live execution and must not be cancelled by the old invocation's cleanup.

## Commit 3 — `e986794e0` stop stubbing solid-js in pure projection tests

Verified: `session-turn-projection`/`-timeline`/`-activity` tests import `session-turn.tsx`, which imports `solid-js`, `solid-js/web` (`Dynamic`), and transitively `lucide-solid`/`@solidjs/meta` via icon chains. The removed partial mocks exposed only a subset of exports, so a real-module import of `isServer`/`createComponent`/`untrack` (e.g. from `@lingui/solid`'s dist, which imports `createComponent` from `solid-js/web`) fails ESM linking whenever the real module loads before the mock — exactly the ordering-dependent flake described. Replacing them with the real solid-js modules is the right call for pure projection tests, and `Empty` remains in use for component mocks, so nothing is left dangling.

The workspace's own `packages/ui/script/test.ts` runs `session-turn-projection.test.ts` and `session-turn-timeline.test.ts` in an `isolated` set (one file per process), which is precisely the cross-file module-caching hazard the commit describes; CI (`bun turbo test --filter='!synergy'`) exercises these and is green at head (Test, Typecheck, Quality, Windows Checks all pass).

## Verification performed

- `bun test --cwd packages/synergy test/session/tool-scheduler.test.ts --rerun-each 10` → 100 pass / 0 fail (includes both the new replay test and the in-flight dedup test).
- Extracted the head versions of the three changed UI test files plus head `packages/ui/src` and ran them with the real solid-js modules: 99 pass / 0 fail (40 activity + projection + timeline).
- CI on head `e986794e`: all checks pass (Test 10m23s, Typecheck, Quality, Windows Checks, etc.).

Note: the only existing formal review (codex, COMMENTED on `50564ed7`) raised the controller-cleanup concern that commit 2 resolves; the earlier `synergy-agent` issue comment covered commit 1 only. This review covers the full current head.

**Synergy**: review complete — approve, no blocking findings.

PR #1159, opened 2026-08-14
fix(embedding): resolve local model paths to buffers in standalone wasm shim

yzxoi:
## Summary
Fixes local embedding models failing to load in the standalone binary with `fetch() URL is invalid`.

## Root Cause
- transformers.js runs in Node mode under Bun (`process.release.name === "node"`), so it passes **local model file paths (strings)** to `InferenceSession.create()`.
- The standalone build plugin (`standaloneEmbeddingBuildPlugin`) replaces `onnxruntime-node` with the WASM backend (`onnxruntime-web/wasm`).
- The WASM backend's `node:fs` branch is compiled out (`if(!1)`), so it falls back to `fetch(path)`.
- Bun's `fetch()` rejects bare file paths with `TypeError: fetch() URL is invalid` (verified empirically: `file://` URLs work, bare paths don't).

Result: local embedding model download succeeds, but the extractor fails to load at pipeline initialization — both the network and `local_files_only` retry paths hit the same failure. The `wasmPaths`/`wasmBinary` preloading fix was working correctly; the failure happens one step later, when loading the ONNX **model file**.

## Fix
The build plugin shim (`script/embedding-onnxruntime-web.ts`) now wraps `InferenceSession.create` to resolve string inputs before reaching the WASM backend:
- disk paths / `file://` URLs → read into `Uint8Array` via `node:fs`
- `http(s)` URLs → fetched into `Uint8Array`
- already-buffered inputs pass through unchanged

## Verification
- Added regression test `loads local model paths into buffers before reaching the WASM backend` that compiles a standalone probe with the same build plugin + `SYNERGY_STANDALONE: true`, then calls `InferenceSession.create(localPath)`. Before the fix it failed with exactly `fetch() URL is invalid`; after the fix it reaches ONNX protobuf parsing as expected.
- All 3 tests in `test/vector/embedding-standalone.test.ts` pass.
- `tsc --noEmit`, oxlint, and prettier all clean.

### DIFF
--- a/packages/synergy/script/embedding-onnxruntime-web.ts
+++ b/packages/synergy/script/embedding-onnxruntime-web.ts
@@ -1,2 +1,35 @@
+import fs from "node:fs/promises"
+import { fileURLToPath } from "node:url"
+import * as onnx from "onnxruntime-web/wasm"
+
+type CreateOptions = Parameters<typeof onnx.InferenceSession.create>[1]
+
+const originalCreate = onnx.InferenceSession.create.bind(onnx.InferenceSession)
+
+// transformers.js runs in Node mode under Bun and passes local model file
+// paths (strings) to InferenceSession.create(). The WASM backend only accepts
+// buffers — its node:fs branch is compiled out — and rejects disk paths with
+// "fetch() URL is invalid". Resolve any string input (disk path, file:// URL,
+// or http(s) URL) to a buffer before handing it to the backend.
+async function resolveModel(model: Uint8Array | string): Promise<Uint8Array> {
+  if (typeof model !== "string") return model
+  if (/^https?:\/\//i.test(model)) {
+    const response = await fetch(model)
+    if (!response.ok) {
+      throw new Error(`failed to load model from ${model}: ${response.status} ${response.statusText}`)
+    }
+    return new Uint8Array(await response.arrayBuffer())
+  }
+  const filePath = model.startsWith("file://") ? fileURLToPath(model) : model
+  return new Uint8Array(await fs.readFile(filePath))
+}
+
+export const InferenceSession = {
+  ...onnx.InferenceSession,
+  async create(model: Uint8Array | string, options?: CreateOptions) {
+    return originalCreate(await resolveModel(model), options)
+  },
+}
+
 export * from "onnxruntime-web/wasm"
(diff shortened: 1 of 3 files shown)


synergy-agent (commented):
## Review — fix(embedding): resolve local model paths to buffers in standalone wasm shim

### Verdict: APPROVE — sound, minimal, well-tested fix

**What I verified** (via GitHub API, head b5eecae):
- `script/embedding-onnxruntime-web.ts` wraps `InferenceSession.create` to resolve string inputs (`file://` via `fileURLToPath`, bare paths via `node:fs`, `http(s)` via `fetch`) into `Uint8Array` before reaching the ORT-web backend. Correct — the WASM backend has no disk branch, so buffering is the only reliable route.
- Wiring confirmed: `script/embedding-runtime-assets.ts:33-39` maps the `onnxruntime-node` import to this shim via `build.onResolve`, so the shim is actually inside the standalone bundle.
- Regression test compiles a standalone probe with `conditions: [browser]` + `SYNERGY_STANDALONE: true` and asserts the failure mode becomes ONNX protobuf parsing (proving the path reached the backend as a buffer). Good test quality.

### Findings

1. **[Suggestion] `http(s)` → `fetch` buffers the whole model into memory** — `embedding-onnxruntime-web.ts:16-20`. This doubles peak memory for remote models vs streaming, but transformers.js passes local cached paths in practice, so impact is low. Acceptable for now; worth a comment or follow-up if remote models are a supported path.

2. **[Nit] Error message on bare-path read failure** — `fs.readFile` throws a raw ENOENT without context; a wrapped message ("failed to load local model from <path>") would aid standalone debugging, where paths are non-obvious. Non-blocking.

3. **[Info] Scope limited to the standalone build path** — the shim only enters the bundle via `build.onResolve`; dev-mode `bun dev` (non-standalone) does not go through it, so this PR alone does not fix dev mode. #1160 covers that layer.

### Notes
- Both #1159 and #1160 modify `test/vector/embedding-standalone.test.ts` — they append disjoint tests (no overlapping hunks), so the "files do not overlap" claim in #1160's body is slightly inaccurate (the **files** do overlap; the **changes** do not). Merge order is safe either way (see #1160 review for relationship analysis).

**Synergy**: review complete — approve, 2 suggestions, 1 info

PR #1160, opened 2026-08-15
fix(embedding): serve local model files through fetch in standalone builds

yzxoi:
## Summary

Fixes local embedding models failing to load in standalone binaries: the ONNX model is downloaded successfully but the extractor fails at pipeline initialization, leaving the memory system unusable.

## Root Cause

- The standalone build uses `conditions: [browser]`, so onnxruntime-web\u0027s `node:fs` branch is compiled out and model files are read through the global `fetch()` with raw local paths or `file://` URLs.
- Bun\u0027s `fetch()` rejects bare file paths (`Unable to connect`), so model loading fails even though the download completed.
- transformers.js also resolves local lookups against the bundled module directory (a virtual drive in standalone builds) because `env.localModelPath` was never set.

## Fix

- `embedding-runtime.ts`: install a local-file fetch shim (global `fetch` + transformers `env.fetch`) before any ONNX session is created, serving raw paths and `file://` URLs from disk. Applied in both standalone and dev modes (Bun always runs the web backend because `process.release.name === node` forces ORT-web).
- `embedding.ts`: set `env.localModelPath = cacheDir` so local lookups resolve from the configured cache directory.

## Verification

- New regression test compiles a standalone probe and verifies local model files are served through fetch inside the compiled artifact.
- `test/vector/embedding-standalone.test.ts`: 3/3 pass (staging, WASM init, local-file fetch).
- Isolated `bun dev` instance: POST /library/embedding/download reached `asset=cached, runtime=ready, progress=100%`; /library/experience/search returned 200 (real 384-dim inference).
- Standalone probe with production build flags + real cached model: `standalone embedding inference ready dims=384 norm=1.000000`.
- Package typecheck, oxlint, prettier clean.

## Related

Complements #1159 (build-plugin shim wrapping `InferenceSession.create`); this fix works at the fetch layer and additionally covers dev mode and the `localModelPath` drift. Files do not overlap.

### DIFF
--- a/packages/synergy/src/vector/embedding-runtime.ts
+++ b/packages/synergy/src/vector/embedding-runtime.ts
@@ -1,5 +1,5 @@
 import path from "node:path"
-import { pathToFileURL } from "node:url"
+import { fileURLToPath, pathToFileURL } from "node:url"
 
 declare const SYNERGY_STANDALONE: boolean | undefined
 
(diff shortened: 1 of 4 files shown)


synergy-agent (commented):
## Review — fix(embedding): serve local model files through fetch in standalone builds

### Verdict: APPROVE with 2 non-blocking suggestions

**What I verified** (via GitHub API, head efcebd4):
- `embedding-runtime.ts` installs a local-file fetch shim on both `globalThis.fetch` and `runtime.env.fetch` (`installLocalFileFetch`, lines 69-79), with `REMOTE_PROTOCOL = /^(https?|blob|data):/i` (line 13) whitelisting remote protocols — normal http(s) model downloads and blob/data URLs pass through to the native fetch untouched.
- Applied in both standalone and dev branches of `loadEmbeddingTransformersRuntime()` (lines 87, 92).
- `embedding.ts:186` sets `env.localModelPath = cacheDir` inside `configure()`; `cacheDir` is the configured value from `resolveCacheDir(config)` (embedding.ts:250).
- New regression test compiles a standalone probe and verifies both `file://` and raw-path fetch return file content. Good test quality.

### Relationship with #1159 (asked in thread)

- **Complementary — yes, with an overlap caveat**: #1159 fixes the ORT-web *session* layer (buffers for `InferenceSession.create`), #1160 fixes the *fetch* layer (paths served from disk). They fix different failure points of the same chain and both touch `test/vector/embedding-standalone.test.ts` (disjoint appended tests, no hunk conflicts — so the "files do not overlap" claim in the body is inaccurate: the files overlap, the changes do not).
- **Merge order**: either works; **#1160 first, then #1159** is slightly safer, since #1159's standalone-scoped shim would otherwise leave dev mode broken between the two merges.
- **Which is the root fix**: #1160 is closer to the root (fetch layer + `localModelPath` drift). **#1159 remains necessary**, not redundant: transformers.js in Node mode passes *paths* to `InferenceSession.create`, and ORT-web's WASM backend internally uses its own captured fetch/env (not the global one) — the shim is the only guarantee the session layer receives a buffer. Together they are defense-in-depth.

### Finding on `localModelPath` (question 3 in thread)

- **The drift fix is correct**: transformers.js `env.js:167` defaults `localModelPath` to `path.join(dirname__, DEFAULT_LOCAL_MODEL_PATH)` — the bundled module directory, which is a virtual drive in standalone builds. `hub.js:137` then computes `localPath = validModelId ? pathJoin(env.localModelPath, requestURL) : requestURL`. Setting `localModelPath = cacheDir` redirects local lookups to the real cache directory. This addresses the fix-plan v2 drift directly.
- **S1 (suggestion) — relative `cacheDir` and the FileCache path**: transformers.js has two lookup paths — `localModelPath` (path-based, fixed by this PR) and the FileCache (`env.cacheDir` / `env.useBrowserCache`). For non-HF repo_ids (`validModelId=false`, e.g. a custom local repo_id), `localPath = requestURL` and the FileCache governs; this PR does not set `env.useBrowserCache = false`. If `cacheDir` is config-relative, cache-keyed lookups resolve against the process working directory — the same class of virtual-drive drift. Recommend: assert/document `cacheDir` as absolute, and consider `env.useBrowserCache = false` for the local-source flow. Non-blocking — the HF-model-id path (the actual supported flow) is fully covered.
- **S2 (suggestion) — global fetch interception scope**: the shim intercepts *every* non-remote-protocol fetch in the process, not just model files (e.g. any future relative-path fetch would hit `Bun.file`). Low risk today (the app's own fetches are http(s) and pass through), but scoping the interception to paths under `cacheDir`/model paths would make the global patch more surgical.

**Synergy**: review complete — approve with 2 suggestions, relationship analysis above

PR #1161, opened 2026-08-15
TTAWDTT: Add viewport-safe UI zoom

TTAWDTT:
## Summary

- add persistent UI zoom controls from 50% to 200% through keyboard shortcuts, Ctrl/Cmd + mouse wheel, and the command palette
- keep the root layout, workbench size limits, and resize-handle pointer deltas correct under CSS zoom
- preserve the viewport-filling app surface at every supported scale

## Testing

- `bun test test/context/zoom-layout.test.ts test/context/layout/workspace.test.ts`
- `bunx oxlint -c oxlintrc.json --deny-warnings packages/app/src/context/zoom.ts`
- `bun run build` (in `packages/app`)

### DIFF
--- a/packages/app/src/app-i18n.ts
+++ b/packages/app/src/app-i18n.ts
@@ -170,6 +170,9 @@ export const AP = {
     message: "Search sessions across all projects",
   },
   layoutColorSchemeUse: { id: "app.layout.colorScheme.use", message: "Use color scheme: {scheme}" },
+  layoutZoomIn: { id: "app.layout.zoom.in", message: "Zoom in" },
+  layoutZoomOut: { id: "app.layout.zoom.out", message: "Zoom out" },
+  layoutZoomReset: { id: "app.layout.zoom.reset", message: "Reset zoom" },
   layoutLeftWorktree: { id: "app.layout.worktree.left", message: "Left worktree" },
   layoutMovedToWorktree: { id: "app.layout.worktree.moved", message: "Moved to worktree" },
   layoutWorktreeDesc: {
--- a/packages/app/src/app.tsx
+++ b/packages/app/src/app.tsx
@@ -25,6 +25,7 @@ import { AuthProvider } from "@/context/auth"
 import { HolosProvider } from "@/context/holos"
 import { InputProvider } from "@/context/input"
 import { FontPreferenceProvider } from "@/context/font-preference"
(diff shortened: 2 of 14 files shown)


yzxoi:
@synergy-agent review

yzxoi:
## Blocking: zoom never fires on Windows

Verified on Windows: **Ctrl+wheel, Ctrl+=, Ctrl+-, Ctrl+0 all do nothing.** The feature is effectively unusable there, and it's not an implementation bug — the events never reach the renderer:

**1. Ctrl+wheel** — Electron/Chromium on Windows consumes ctrl+wheel as a webContents-level zoom gesture; the `wheel` event is **not dispatched to the page** (see electron/electron#12436: on Windows, holding Ctrl means the `wheel` event never fires in the renderer). The only way an app can observe this gesture is the `webContents` `zoom-changed` event, which per electron/electron#17747 is explicitly "the only way the Electron app will be able to handle these events."

**2. Ctrl+=/-/0** — Even with `Menu.setApplicationMenu(null)` on non-macOS (`packages/desktop/src/menu.ts:9-12`), Chromium's page-zoom accelerators are consumed at the browser layer before any `keydown` reaches the DOM. (electron/electron#40674 also shows the role accelerators are inconsistent on Windows/Linux anyway: `ctrl+=` doesn't map, `ctrl+shift+=` does.)

So `ZoomController`'s capture-phase `wheel` listener (`packages/app/src/context/zoom.ts`) and the command-palette keybinds (`packages/app/src/pages/layout.tsx`) are dead on Windows. Only palette _clicks_ work, since that path is pure renderer.

**Suggested fix** — move zoom interception to the main process and forward it to the renderer:

- `before-input-event`: intercept `input.control` + `Equal` / `Minus` / `Digit0` → forward to the renderer (the existing `installWindowInputShortcuts` in `packages/desktop/src/main.ts` is the right pattern)
- `webContents` `zoom-changed` (or a `before-input-event` wheel interception) → forward ctrl+wheel
- `setVisualZoomLevelLimits(1, 1)` so the native zoom layer can't stack with the CSS zoom
- add a Windows runtime verification to the PR checklist

This also means the earlier macOS concern stands (View-menu zoom roles steal the same keybinds), so the desktop shell needs a single, main-process-owned zoom path across platforms.

PR #1162, opened 2026-08-15
fix: disable file watcher without binding; keep scoped dashboard resource coverage

yzxoi:
## Summary

Fixes two related issues surfaced by a production desktop instance where the control plane appeared unavailable and the server was saturated:

1. **File watcher binding missing → unbounded retry/reload loop**
   When the `@parcel/watcher` platform binding package is absent (e.g. a broken desktop bundle), every `subscribe()` attempt throws, and the recovery layer retries every second while each failure triggers a full config/agent/command/skill/tool-registry reload. This loop saturated CPU (~99%), flooded the log (12万+ ERROR lines in one window), and caused policy-worker timeouts that made the control plane genuinely unresponsive.
   - `FileWatcherBinding.resolvable()` detects the platform binding once without executing it.
   - `FileWatcher.init()` / `FileWatcher.reload()` now disable file watching with a single log line instead of retrying forever.

2. **Scoped performance dashboard reported control plane as unavailable**
   Process resource samples are global and carry no scope ID. `PerformanceDashboard.summary()` filtered them by the requested `scopeID`, which dropped every server row and produced `completeness: "unavailable"` for the control plane owner. Resource rows are now read unscoped; metrics and issues keep their existing scoped semantics.

## Verification

- `bun test test/file/watcher-binding.test.ts test/file/watcher-events.test.ts test/performance/store.test.ts` — 45 pass
- `bun run typecheck` (packages/synergy) — pass
- Prettier check on changed files — pass
- oxlint on changed files — 0 warnings / 0 errors
- New regression tests:
  - `FileWatcherBinding.resolvable()` detects a missing binding package
  - scoped dashboard summary keeps `control_plane` coverage (`completeness: "full"`)

Note: `test/observability/sqlite-maintenance-budget.test.ts` and `test/observability/store-worker-mode.test.ts` fail on Windows with `EBUSY` temp-dir cleanup in both this branch and baseline (pre-existing platform issue, unrelated to this change).


yzxoi:
@synergy-agent review

PR #1170, opened 2026-08-16
fix(app): remove global press-scale animation

TTAWDTT:
## Summary

- Remove the global active-state scale transform applied to every button, link, and element with `role="button"`.
- Preserve intentional component motion, including the height transition used when tool output and file changes expand or collapse.

## Why

The global `transform: scale(0.97)` caused the text and background of session rows and tool-card triggers to shrink and grow on every click. In a workbench UI this reads as visual jitter rather than useful feedback, especially when selecting sessions or expanding command and file-change details. Selection, hover, and disclosure state already have their own visual treatment, so the global press-scale effect is redundant.

## Testing

- `bun .\\node_modules\\vite\\bin\\vite.js build` in `packages/app`
- Started an isolated local server and Vite preview; verified `GET /global/health` and the app root both return `200`.


yzxoi:
Thanks for this fix, @TTAWDTT! Removing the global press-scale animation eliminates the jitter on session rows and tool-card triggers while keeping intentional component motion (e.g. tool output height transitions) intact. I checked against the latest dev branch: no tests depend on the removed rule and local component animations are unaffected. LGTM, approved.

yzxoi (approved):

PR #1171, opened 2026-08-16
refactor(ui): refine icon hierarchy

TTAWDTT:
## Summary

Refines icon hierarchy across the application shell and product navigation.

- Use single edge-oriented arrows for sidebar expand and collapse controls.
- Replace the agenda, performance, and plugins navigation icons with `alarm-clock`, `chart-no-axes-combined`, and `puzzle`.
- Use a downward chevron with rotation for collapsible controls.
- Reduce secondary sidebar action icon size and register every new semantic glyph in the built-in icon registry.

## Why

The sidebar is a dense, frequently scanned surface. The previous calendar-with-clock, radar, package, and double-chevron glyphs contained overlapping or enclosed shapes that made adjacent navigation items feel visually heavier than the surrounding interface.

The revised mappings retain the same semantics with simpler, distinct silhouettes. The sidebar arrows use an edge line plus a single direction arrow, so they describe opening or closing the sidebar without competing with generic navigation chevrons. Each semantic token uses a unique registered glyph, preserving the project's icon contract.

## Validation

- `bun test test/semantic-icon.test.ts` in `packages/ui`
- `bun run typecheck` in `packages/ui`
- `bunx prettier --check` for all changed files
- Ran `bun script/test.ts` in `packages/ui`: all non-browser tests passed; three existing browser-layout suites could not launch locally because Playwright Chromium is not installed on this machine.


yzxoi:
Thanks for this contribution — the icon hierarchy refinement is thoughtful, and we appreciate the careful registry updates and validation work.

After reviewing, we've decided not to merge this PR. The current icon set already reads well in the sidebar, and we feel the existing squarer, more structured glyphs align better with our current design language than the rounder single-stroke replacements.

Thanks again for the time and effort!

PR #1172, opened 2026-08-16
fix(provider): keep health and startup responsive when model catalogs are unreachable

### LINKED ISSUE #1169: Startup/health stalls when a provider's model catalog is unreachable (proxy off) — single provider outage blocks the whole service
## What happened?

Configured codex/gpt models through a proxy. After turning the proxy off, Synergy fails to start properly and the health check is broken. Synergy supports multiple models, so one provider being unreachable (its model-catalog endpoint requires the proxy) should not stall or fail the whole service — but currently a single-provider outage appears to drag down startup.

Relevant logs:

```
WARN  2026-08-16T06:32:21 +9597ms service=models.dev url=https://models.dev/api.json error=The operation timed out. failed to fetch models catalog
INFO  2026-08-16T06:32:21 +589ms service=provider reloading provider state
INFO  2026-08-16T06:32:21 +0ms service=provider provider state reloaded
INFO  2026-08-16T06:32:21 +1ms service=agent reloading agent state
INFO  2026-08-16T06:32:21 +0ms service=agent agent state reloaded
INFO  2026-08-16T06:32:22 +818ms service=provider init
INFO  2026-08-16T06:32:28 +5937ms service=server mode=delta global event ws connected
INFO  2026-08-16T06:32:28 +2ms service=server rid=7ae5e6e6 method=GET route=/global/event/ws status=200 duration=5 request
WARN  2026-08-16T06:32:31 +3072ms service=provider.catalog providerID=openai-codex profileID=openai-codex failure=invalid_response error=provider returned an empty model catalog failed to refresh provider model catalog
```

## Expected behavior

- A single provider's model-catalog refresh failure (timeout / network / `invalid_response`) should not block startup or stall the health check.
- On failure, fall back to cached / bundled model lists and degrade only that provider — other configured models keep working.
- A failed catalog refresh should not remove the provider from the usable list (which surfaces as `modelReady=false` / "no model configured").

## Steps to reproduce

1. Configure codex / gpt models through a proxy; `synergy start` runs normally.
2. Turn the proxy off.
3. Start Synergy again: startup stalls / health check fails, logs show the models.dev timeout and openai-codex `invalid_response` above.

## Environment

- OS: Linux
- Synergy version: to be filled in during reproduction (`synergy --version`)
- Bun version: to be filled in (`bun --version`)

## Initial investigation (static code read, not yet reproduced)

- `packages/synergy/src/provider/models.ts` `doRefresh()`: the models.dev fetch has a 10s `AbortSignal.timeout` and only warns on failure with cached fallback — the first WARN is expected degradation and does not itself block.
- `packages/synergy/src/provider/provider.ts:492`: `Provider.state()` awaits `ProviderCatalog.resolve({ includeLive: true })` synchronously on first build; `/global/health` → `Provider.list()` depends on `state()`, so health is held up by catalog resolution (network steps up to ~10s each).
- The startup-side health window is much shorter: daemon health polling aborts after 1.2s (`src/daemon/server-process-lock.ts`) and `cli/cmd/start.ts` `modelReadinessNotes` uses 3s — startup is judged failed when catalog resolution exceeds that window.
- `provider.ts:725`: a provider with an empty model list is `delete`d; if the only configured provider's catalog refresh fails with no cached snapshot, `modelReady` becomes false and the service looks like it "failed to start / has no model".

Suggested direction: health should not be blocked by provider catalog resolution (avoid awaiting network steps, or give it a much more lenient timeout); catalog refresh failures should keep bundled/cached models and mark the provider degraded instead of emptying it.

yzxoi:
## Summary

Fixes a startup/health-check failure when the model provider registry is unreachable (e.g. the configured proxy is turned off): the daemon/Desktop readiness probes were timing out even though the runtime actually came up, making Synergy look like it failed to start.

**Root cause:** the first `GET /global/health` triggered a provider-state build that awaited the signed remote catalog fetch with no cache-first path. On a black-hole network that meant two 10s fetch timeouts, while health probes give up after 1.2–3s. Per-provider model discovery was already isolated (snapshots + background retry), so a single unreachable model provider never blocked anything — the blocking fetch was the shared provider registry.

## Changes

- **`packages/synergy/src/provider/catalog.ts`**: remote catalog cache is now consulted first. A fresh cache short-circuits the network entirely; a stale cache is served immediately while a background refresh runs (deduplicated, invalidates the in-memory projection on success). Only a cold cache (no file) blocks on the network.
- **`packages/synergy/src/server/server.ts`**: `/global/health` waits at most 1s for the provider state build, then answers optimistically while it is still in flight, so the readiness signal never blocks on catalog discovery.

## Tests

New behavioral coverage:

- `test/provider/catalog-remote-cache.test.ts` — fresh cache short-circuits the fetch; stale cache serves immediately and refreshes in the background.
- `test/server/health.test.ts` — health answers within budget when the provider build hangs; `modelReady` semantics preserved when the build resolves quickly.

Verified locally:

- `bun test test/provider/catalog-remote-cache.test.ts test/server/health.test.ts test/provider/catalog-stability.test.ts test/server/shutdown-admission.test.ts` — 28 pass
- `bun run typecheck` — clean
- `prettier --check` on changed files — clean
- `oxlint` on changed files — no errors


yzxoi:
Closing this PR in favor of #1173, which covers the same root cause more completely:

- **Same fix, better placement**: both PRs make the signed registry fetch cache-first (stale cache served immediately + single-flight background refresh). #1173 keeps this entirely in `catalog.ts`; #1172 also touched the health endpoint.
- **#1173 additionally fixes the root cause that matches the reported logs**: a refresh that fails *before any successful verification* persisted an empty `failure` snapshot, which `applyCachedDiscovery()` applied as authoritative — wiping the bundled fallback models and making the provider look unconfigured (`modelReady=false`). This is exactly the `providerID=openai-codex ... failure=invalid_response error=provider returned an empty model catalog` path in the user's log.
- **#1173 keeps `forceRefresh` semantics** (`synergy models --refresh` still waits on the network); #1172 accidentally short-circuited it for fresh caches.
- **#1173 does not block on the network on a cold first run**; #1172 still awaited the fetch when no cache file existed.

The one part of #1172 worth keeping is the bounded wait on `/global/health` as deep defense, but with the true state on timeout rather than an optimistic `modelReady=true`. I've left that as a suggestion on #1173.

PR #1175, opened 2026-08-16
fix(app): avoid project list bounce while loading sessions

TTAWDTT:
## Summary

- defer opening a local project disclosure until its real session navigation data is available
- show an inline loading indicator in the project toggle instead of expanding six placeholder rows
- retain the existing supplemental-project disclosure so its explicit **Load sessions** action remains reachable

## Problem

Expanding an unloaded local project previously animated six skeleton rows into the sidebar. Once the actual session list arrived, projects with fewer than six sessions were pulled back upward. This made projects below the expanded one appear to overshoot and rebound.

## Testing

- `bun test test/components/sidebar/project-disclosure.test.ts`
- `bunx prettier --check packages/app/src/components/sidebar/sidebar.tsx packages/app/src/components/sidebar/project-disclosure.ts packages/app/test/components/sidebar/project-disclosure.test.ts`
- `bun run build` (in `packages/app`)

`bun run typecheck` was not runnable locally because this Windows checkout materializes the existing `src/custom-elements.d.ts` symlink as a plain path file. CI remains the full validation environment.

### DIFF
--- a/packages/app/src/components/sidebar/project-disclosure.ts
+++ b/packages/app/src/components/sidebar/project-disclosure.ts
@@ -1,0 +1,8 @@
+export function shouldOpenProjectDisclosure(input: {
+  expanded: boolean
+  isSupplemental: boolean
+  navLoaded: boolean
+}): boolean {
+  if (!input.expanded) return false
+  return input.isSupplemental || input.navLoaded
+}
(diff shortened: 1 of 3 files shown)


yzxoi (approved):
Approved — nice, clean fix. The disclosure now waits for real session data instead of bouncing through skeleton placeholders, and the extracted helper plus the four-state unit tests make the policy easy to verify. Thanks for the effort and the thorough testing!

PR #1176, opened 2026-08-16
feat(ui): add built-in color themes

TTAWDTT:
## Summary
- Add seven built-in theme families alongside the existing Synergy theme.
- Replace the General settings theme menu with an accessible visual picker.
- Resolve built-in and plugin themes through one catalog while retaining saved theme selection and boot skin snapshots.
- Adapt Catppuccin, Tokyo Night, Ayu, Rose Pine, Kanagawa, Everforest, and Solarized palettes to Synergy semantic tokens with accessible text and focus roles.

## Testing
- packages/ui targeted theme tests
- packages/ui typecheck
- Manual verification in the local Synergy app: Settings > General lists all eight themes and each selection is operable.

## Palette Sources
The source links, licenses, and selected light/dark variants are documented beside each built-in theme definition.


yzxoi:
Thanks for the contribution — the unified theme catalog and the visual picker are genuinely nice touches, and it's good to see all eight themes passing the built-in contrast validation.

Two suggestions from my side:

**1. Consider shipping these as a plugin theme (recommended)**

The canonical theming docs (`docs/reference/frontend-theming.md`, "Creating a New Selectable Theme") recommend implementing new selectable themes as structured plugin themes, and explicitly advise against a parallel built-in registry ("Do not add a parallel built-in theme registry or theme-specific CSS bundle"). As the earlier review noted, the current PR sits at odds with that guidance.

The good news is the work here is already factored exactly the way a theme plugin would want it — palette seeds plus a small set of overrides — so converting these seven themes into an official theme plugin would be mostly mechanical (the `createTheme` palette definitions could be reused nearly as-is), while gaining independent distribution, the validated manifest contract, and no new registry machinery in the core package. Happy to help scaffold that if useful.

If a curated set of built-in themes is the intended product direction instead, that's a legitimate choice too — in that case just updating the docs / `AGENTS.md` in the same PR to describe the built-in catalog would close the gap.

**2. Light variants could use another polish pass**

The dark sides of these palettes look great, but the light sides feel comparatively less developed: in a couple of palettes the weak text color is identical to the base text (e.g. Everforest and Solarized light), which flattens the type hierarchy, and several light hover/inset steps sit very close to their surfaces, making interactive states a bit hard to perceive. A light-mode pass (a clearer weak-text step, a touch more separation between surface steps) would make the picker previews and the actual light UI look as polished as the dark mode already does.

Overall a strong direction — happy to help iterate on either point.

yzxoi:
Follow-up on the theme architecture — one concrete suggestion for how the built-in themes could align with the shared pipeline (regardless of whether they ship built-in or as a plugin).

**The overrides are currently re-implementing what the resolver already does.**

`themedOverrides` in `default-themes.ts` remaps ~100 tokens (all the surface/input/border/button steps) to palette-derived values. The resolver (`resolveThemeVariant`) already generates all of those from the 13 seeds — with consistent step spacing, layer hierarchy, and the WCAG-AA contrast assertions applied on top. Overriding them back to hand-picked hex values means each theme pays for ~500 lines of maintenance to reproduce (and occasionally fight) behavior the shared code already provides.

Concretely, I'd suggest converging each theme down to:

1. **The 13 seeds only** — the `Palette` fields map quite naturally: `neutral` ← text/textWeak, `primary`/`interactive`/`info` ← accent, `success`/`warning`/`error` ← same-named, `diffAdd`/`diffDelete` ← success/error.
2. **A small syntax/markdown whitelist** (~10–20 tokens) — this is the one real gap: `syntax-string/keyword/type/...` and `markdown-heading/link/...` are hard-coded in the resolver rather than seed-derived, so a theme identity needs those as intentional exceptions (the PR's `syntaxString/syntaxKeyword/syntaxType` palette fields already acknowledge this).

Removing the surface/input/border/button overrides would also **resolve the dark polarity finding from the earlier review** — the darker-than-surface `inset`/`input` steps are exactly what happens when the resolver's dark hierarchy (surfaces step *brighter* than their containers) is overridden by hand-picked palette values. With seeds-only, every dark variant gets the correct step direction for free, plus all the contrast guarantees.

The nice outcome is a clean boundary: **built-in curated themes = seeds-only, quality machine-guaranteed by the shared resolver; plugin themes = full freedom to override every token.** Same schema → resolver → contrast pipeline either way, and the built-in catalog shrinks from ~500 lines of overrides per theme to ~30 lines of seeds plus a small identity whitelist.

Happy to draft the seed mapping for one theme as a reference if that would help.

PR #1177, opened 2026-08-16
feat(ui): add built-in color themes

TTAWDTT:
## Summary
- Replaces #1176, which GitHub cannot reopen after its head branch was rewritten.
- Add seven built-in theme families alongside the existing Synergy theme.
- Use one shared resolver pipeline for built-in and plugin themes.
- Define every built-in theme with 13 seeds only: nine semantic seeds plus string, keyword, type, and property syntax seeds.
- Generate surface, input, border, button, and contrast-safe roles through the shared resolver; plugins can still supply token overrides.
- Preserve compatibility for existing nine-seed plugin themes by deterministically deriving the four new syntax seeds at parse time.

## Testing
- `bun run --cwd packages/plugin typecheck`
- `bun test --cwd packages/ui test/builtin-themes.test.ts test/plugin-theme-registry.test.ts test/theme.test.ts` (66 passing)
- Manual verification in the local Synergy app: selecting and saving the Synergy theme applied resolver-generated surface, input, border, and syntax colors. The preview was restored to Everforest afterward.

## Palette Sources
The source links, licenses, and selected light/dark variants are documented beside each built-in theme definition.


yzxoi:
Thanks for iterating on this — the seed-only direction is exactly right, and I've taken the liberty of building on top of it with a follow-up refactor. Here's a summary of what changed and one honest caveat.

## What we changed (uncommitted in the PR head, pushed as `5d3e087ef9` on the fork branch)

1. **Every built-in skin is now seeds-only.** All 8 skins (Synergy + 7 new ones) go through the shared resolver with 13 seeds and zero overrides. The contract test `built-in themes use the complete seed pipeline without general overrides` enforces this going forward, so the resolver (and its WCAG-AA assertions) is what guarantees quality — not hand-picked hex values.

2. **No parallel registry.** The built-in lookup path is gone. Built-in skins pre-register into the *single* shared registry at module load, plugin themes register into the same table, and a plugin may not shadow a built-in skin id (covered by a registry test). `getBuiltinTheme` and the parallel path in `context.tsx` were removed.

3. **16 variants covered explicitly.** New test `provides 16 variants across 8 skins and both color schemes` verifies every skin resolves in both light and dark, with distinct colors per mode.

4. **Fixed tests the seed-only default skin had broken** (these were failing on the PR head): `packages/desktop/test/startup-page.test.ts`, `packages/app/test/testing/color-token-contract.test.ts` (theme-authoring boundary exemption for `default-themes.ts`), and `packages/ui/test/visual-token-contract.test.ts` (hard-coded old surface values → resolved values).

5. **Docs synced**: `docs/reference/frontend-theming.md` now describes the single shared registry, the 9→13 seed derivation rule, and the built-in skins stay seeds-only; `packages/ui/AGENTS.md` updated to match.

## The honest caveat: light mode still needs a polish pass

I tested all 16 variants locally, and while the dark sides look great, the light sides are still the weakest part of this PR. Two things stand out:

- **Flat text hierarchy in light mode.** With seeds-only resolution, `text-weak` collapses to the same color as `text-base` in several skins (e.g. Everforest and Solarized light), and `text-weaker` ends up *closer* to the background than `text-weak`. The resolver's readable-color picker skips mid-steps to hold a 4.5:1 floor, and without overrides there's no longer anything compensating for that on light neutrals. The old Synergy overrides masked this; the seeds-only route exposes it.
- **Light surface steps sit very close together**, which makes hover/selected states and layered cards harder to perceive than in dark mode.

This is a resolver-level characteristic more than a per-skin seed problem, so the fix probably belongs in the shared pipeline (e.g. a dedicated weak-text step for light neutrals) rather than in each skin's seeds. Happy to dig into that as a follow-up if it would help.

Issue #309, opened 2026-07-05
Sidebar session switch can crash in TipTap BubbleMenu getComputedStyle during session refresh


EricSanchezok:
## What happened?

Clicking a session in the frontend sidebar can throw a runtime error while the target session is refreshing/loading:

```text
TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
TypeError: parameter 1 is not of type 'Element'.
    at $e (http://127.0.0.1:60929/assets/vendor-tiptap-w0ztr0jR.js:1:13504)
    at Zn (http://127.0.0.1:60929/assets/vendor-tiptap-w0ztr0jR.js:1:12599)
    at Fn (http://127.0.0.1:60929/assets/vendor-tiptap-w0ztr0jR.js:1:14065)
    at Zw (http://127.0.0.1:60929/assets/vendor-tiptap-w0ztr0jR.js:1:20685)
    at Object.fn (http://127.0.0.1:60929/assets/index-DV-lk4_G.js:1244:60452)
    at at (http://127.0.0.1:60929/assets/vendor-solid-sEvrnwfb.js:1:5970)
    at ue (http://127.0.0.1:60929/assets/vendor-solid-sEvrnwfb.js:1:5744)
    at Re (http://127.0.0.1:60929/assets/vendor-solid-sEvrnwfb.js:1:7103)
    at un (http://127.0.0.1:60929/assets/vendor-solid-sEvrnwfb.js:1:8243)
    at http://127.0.0.1:60929/assets/vendor-solid-sEvrnwfb.js:1:7888
```

The user-facing trigger observed so far is: click a session row in the frontend sidebar while that session appears to be refreshing/loading. The stack is in the `vendor-tiptap` chunk, not the sidebar chunk.

## Expected behavior

Switching sessions from the sidebar should not crash the frontend. If a Notes/Blueprint editor panel is being mounted, reused, hidden, or destroyed during session refresh, TipTap/Floating UI menu position updates should not receive a detached/non-Element node.

## Steps to reproduce

The exact minimal reproduction is still being narrowed down, but the strongest current reproduction hypothesis is:

1. Open the Web/Electron frontend.
2. Use a session that has the side Workbench Notes panel open, or a session that can auto-open Notes via a Blueprint note write event.
3. Ensure a TipTap note editor / BubbleMenu instance has been mounted recently, preferably with an active or recently active text selection.
[...]

PR #1188, opened 2026-08-17
fix(ui): prevent stale timeline item reads during session switch

EricSanchezok:
## Summary

Fixes a stable crash when switching sessions in the sidebar: `TypeError: k(...) is not a function` thrown from the Solid runtime's `Switch` guard (`untrack(switchFunc)()?.[0]`), surfaced through the session-turn timeline renderer. The crash reproduced reliably on sessions whose last assistant turn ended with `Agent worker exited (SIGTERM)` (an interrupted turn leaves a special reasoning/timeline shape that hits the disposed-memo path).

This is the same bug class as the earlier `Stale read from <Show>` fixes (issue #309, PR #1070): a reactive guard whose owner was disposed during session switch is re-evaluated by a still-mounted child.

## Root cause

`TimelineDisplay` (in `packages/ui/src/components/session-turn.tsx`) rendered every timeline item through **unkeyed** `<Switch>`/`<Match>` branches:

```tsx
<Match when={timelineItem()}>
  {(item) => <TimelineItemDisplay item={item()} ... />}
</Match>
```

For unkeyed `Match`, the child render function receives an accessor; Solid keeps the branch mounted while the guard memo is alive. During a sidebar session switch the turn's reactive owner is disposed, and Solid's `Switch` re-evaluates the guard inside `untrack(switchFunc)()`. With the memo already disposed the call returns `undefined`, and the runtime's next step calls it → `k(...) is not a function` (minified). The `get fullText` / `get item` frames in the user-visible stack correspond to the `CompactReasoningLine` and timeline item prop getters that were still subscribed to that disposed guard.

## Fix

Two layers, mirroring the approach already used for message parts in PR #1070:

1. **Key the `Switch`/`Match` branches in `TimelineDisplay`.** Keyed branches pass the resolved `when` value directly to the child render function, so children never receive a guard accessor that can be re-read after disposal — eliminating the crash path at its source. Child components now read plain values (`item.message`, `item.parts`, …) instead of accessor calls (`item().message`, …).

1. **Wrap each timeline item in an `ErrorBoundary`.** `TimelineDisplay` is now exported and self-contained: any residual renderer failure inside a single timeline item degrades to a local `ErrorCard` (`data-slot="session-turn-timeline-item-error"`) instead of taking down the whole turn or session view. This matches the Part-level error boundary pattern added in PR #1070 and gives the same isolation guarantee for the timeline row surface.

## Behavior notes

- The keyed change does **not** remount turns on incremental `message.updated` reconciliation: `TimelineDisplay` is keyed by the stable timeline item (`displayItemStableKey`, part/message identity) inside the existing `For`, so streaming, tool-card expansion, scroll position, and Markdown state are preserved. The existing activity-switch DOM test (minimal → balanced → full without remounting) still passes.
- The error fallback renders the error message only (same redaction discipline as PR #1070: no message content, tool payloads, or workspace paths).

## Tests

- Added `packages/ui/test/components/session-turn-timeline-boundary.test.ts`: builds a real Solid bundle (Vite fixture) and renders `TimelineDisplay` with (a) a stable reasoning item and (b) an item whose part accessor throws — asserting the stable row renders, the failing item shows a local error card, and the error message is contained.
- Registered the new test in the UI runner's isolated set (`packages/ui/script/test.ts`).

Verified:

```bash
bun test test/components/session-turn-timeline-boundary.test.ts
bun test test/components/session-turn-activity-switch.dom.test.ts test/components/message-part-error-boundary.test.ts test/components/compact-reasoning-settlement.dom.test.ts test/components/compact-reasoning.dom.test.ts test/components/session-turn-activity.test.ts test/components/session-turn-projection.test.ts test/components/session-turn-timeline.test.ts   # 107 pass
bun run typecheck
bunx prettier --check src/components/session-turn.tsx test/components/session-turn-timeline-boundary.test.ts script/test.ts
bunx oxlint src/components/session-turn.tsx test/components/session-turn-timeline-boundary.test.ts
```


yzxoi:
Addressed both review findings in `57f8da8451` (pushed to this branch).

**Blocking — keyed `Match` remount regression:** restored the non-keyed `<Match>` branches in `TimelineDisplayInner` (children receive the accessor `item()` again) while keeping the `ErrorBoundary` wrapper. This matches the decision already validated in #1070 (`c82c18d141`, "keep turn boundary non-keyed and preserve part renderer memo"): `message.updated` reconcile replaces display item references on every streaming tick, so keyed branches rebuilt inner nodes several times per reply, resetting `CompactReasoningLine.userScrolled`, tool-card `Collapsible` state, and Markdown state. With non-keyed branches, the stale disposed-guard read still cannot reach the runtime's `untrack(switchFunc)` crash path — it now surfaces through the `ErrorBoundary` as a local error card.

**Suggestion — test does not cover the fix:** rewrote `session-turn-timeline-boundary.test.ts` (real Solid bundle fixture) with two regression scenarios:

1. **Streaming tick identity** — the display item reference changes on every tick (mirroring `reconcile`); asserts the inner `[data-component="compact-reasoning"]` node keeps its DOM identity and the text streams in. This test fails on the keyed code (reproduced) and passes on the non-keyed restore.
2. **Stale read after mount** — the part's `text` getter starts throwing after the item is mounted, with a throw counter proving the error actually fires; asserts the local error card renders, the error message is contained, and the sibling stable row keeps rendering. This covers the disposed-guard crash scenario the original test did not exercise.

Verification: boundary test 3 pass; related suite (activity-switch DOM, message-part error boundary, compact-reasoning DOM + settlement, session-turn activity/projection/timeline) 109 pass, 0 fail; `tsgo` typecheck, `oxlint`, `prettier --check` clean.

PR #1189, opened 2026-08-18
feat(tool): auto-expand deferred tools on direct model call

EricSanchezok:
## Summary

- When the model directly calls a tool that exists in the current session's tool registry but is hidden because its group is not expanded (`expandedGroups`) or it is a search-only tool not yet activated (`activatedTools`), the runtime now **auto-expands it and executes the call in the same turn** — no wasted `expand_tools` → retry round-trip.
- Works for built-in groups (`browser`, `agenda`, `session`, `note`, `memory`, `email`, `worktree`), search-only tools, and deferred MCP server groups (≥ 100 tools). Always on; no new config flag.
- **Visibility-only, never authorization**: eligibility re-checks agent/session permission, `userTools`, workflow mode policy, and the `expand_tools` permission before persisting the expansion through the canonical `session.toolState` (same state and `Session.update` mutation lock as `expand_tools`). Denied, user-disabled, mode-blocked, `internal`/ephemeral, schema-failed, and truly unknown tools keep their existing diagnostics unchanged.
- The real tool still executes through the existing scheduler, executor classification, enforcement gate, timeout, and settlement pipeline. Auto-expanded calls are observable: part metadata `autoExpanded: true`, a `tool.auto_expanded` observability event, and a muted "Auto-loaded" tag on the tool card in the Web UI.
- Agent prompts (`synergy`, `synergy-max`, `scout`, `plan`) document the behavior.

## Implementation

- `packages/synergy/src/session/tool-resolver.ts` — `applyAvailability` computes `autoExpandable`; new `runtimeToolFor` and `autoExpandTool` resolve the real runtime tool after persisting the expansion against a fresh session snapshot (fail closed).
- `packages/synergy/src/session/processor.ts` — suppresses `tool-error` failure record/settlement for auto-expandable running calls; swaps in the real tool at deferred dispatch; merges `autoExpanded` metadata; emits the event (per-turn cache).
- `packages/synergy/src/session/invoke.ts` — threads `autoExpandable` + resolver input into `streamInput`.
- `packages/synergy/src/tool/exposure.ts` — pure `expansionForTool` / `expandState` helpers.
- `packages/ui/src/components/basic-tool.tsx` — "Auto-loaded" tag from `metadata.autoExpanded`.
- Tests: `packages/synergy/test/tool/auto-expand.test.ts` (15 tests: eligibility, fail-closed cases, resolver, processor interception incl. `tool-error` ordering, MCP groups).
- Docs: `docs/architecture/execution-boundaries.md` invariant + decision record `docs/decisions/implemented/feature/2026-08-18-auto-expand-deferred-tool-calls.md`.

## Verification

- `bun test test/tool/auto-expand.test.ts` — 15/15 pass
- Existing regression suites (`exposure`, `observability`, `tool-call-execution`, `processor`) — 91/91 pass
- `bun run --cwd packages/ui test` — pass
- `bun run test:changed` — 4160/4166 pass; 5 timeouts (daemon/migrations under parallel load) re-ran green in isolation
- `bun run quality:quick` — all 14 gates pass (format, lint, skill, package-guide, test-layout, localization, typecheck, monorepo, package, doc, decision, deadcode, secrets, workflow)


yzxoi (changes requested):
## Review: blocking issues found

The feature is well-structured (fail-closed eligibility, canonical `toolState` reuse, observability), and the 15 new tests pass. However, I found **two blocking bugs** with runtime reproductions, both missed by the test suite because `test/preload.ts` registers the in-process stream, bypassing the production agent-worker pool path.

### P0 — `autoExpandable` / `resolverInput` leak into the worker envelope (breaks every production turn)

`packages/synergy/src/session/processor.ts` destructures `streamInput` keeping only `executionTools`/`executorKinds`/`memoryTurn`:

```ts
const {
  executionTools: _executionTools,
  executorKinds: _executorKinds,
  memoryTurn: _memoryTurn,
  ...agentTurnInput
} = streamInput
const stream = await AgentTurn.stream(agentTurnInput)
```

The new `autoExpandable` (a `Set`) and `resolverInput` (an object containing functions, the full session, and ephemeral tools) flow into `agentTurnInput` → `AgentTurn.stream` → `AgentWorkerPool.run` → `AgentTurnProtocol.serializeTurn`. `TurnInputSchema` is `.strict()`, so the envelope is rejected. Runtime check against `TurnInputSchema.parse`:

```
P0 issues: [{"code":"unrecognized_keys","keys":["autoExpandable","resolverInput"],"path":[],...}]
```

Since `invoke.ts` now sets both fields on **every** turn, every turn through the worker pool would fail serialization. Tests are green only because they run in-process.

**Fix**: strip `autoExpandable` and `resolverInput` in the processor destructure — they are processor-internal and must never enter the worker protocol (the schema is strict and cannot carry functions anyway).

### P1 — lost update on concurrent auto-expands

`ToolResolver.autoExpandTool` reads `session.toolState` **outside** the `SessionMutation` lock, computes `updatedState` from the stale snapshot, then `Session.update` overwrites `draft.toolState` wholesale:

```ts
const updatedState = ToolExposure.expandState(session.toolState, ...)
await Session.update(session.id, (draft) => { draft.toolState = updatedState })
```

Two concurrent auto-expands of different groups race: one wins, the other's expansion is lost and its tool call fails. Reproduced with `Promise.all([autoExpandTool(A), autoExpandTool(B)])` on two different deferred groups:

```
P1 — resolved A: false | resolved B: true | final expandedGroups: ["verify-race-b"]
```

(the loser flips between runs — a genuine race, not a deterministic failure). Same-turn multi-tool calls are exactly what a streaming model turn produces, so this is not a corner case.

**Fix**: merge inside the `Session.update` editor so the read-modify-write happens under the mutation lock, e.g. `draft.toolState = ToolExposure.expandState(draft.toolState, ...)`, then re-resolve against a fresh snapshot.

### Minor

- `runtimeToolFor`/`autoExpandTool` re-run a full catalog collection (including MCP, which can be ≥100 tools) per unique tool per turn; consider a targeted resolve or a shared catalog.
- Test gaps: no concurrency coverage (P1) and no pool-path/serialization coverage (P0). A pool-path integration test (unregistering the in-process stream) or an explicit protocol-layer unknown-key assertion would have caught both.

### Verification performed

- `bun test test/tool/auto-expand.test.ts` — 15/15 pass (in-process path).
- Targeted runtime repros: strict schema rejection of the two new keys; concurrent two-group auto-expand losing one expansion.

yzxoi:
taking this PR

yzxoi:
## Fixes pushed (88af03d) for review findings

Both blocking issues from the review are fixed, with regression tests:

### P0 — internal fields no longer leak into the agent-turn envelope

`processor.ts` now strips `autoExpandable` / `resolverInput` in the destructure before forwarding to `AgentTurn.stream`, so the strict worker `TurnInputSchema` no longer rejects every turn. Previously tests passed only because `test/preload.ts` registers the in-process stream, bypassing serialization.

New test: "does not leak processor-internal fields into the AgentTurn stream input" — asserts the mock receives neither field (plus `executionTools`/`executorKinds`, which were already stripped).

### P1 — concurrent auto-expands no longer lose updates

`autoExpandTool` now merges the expansion into `draft.toolState` **inside** the `Session.update` editor (under the `SessionMutation` lock) instead of computing `updatedState` from a pre-lock snapshot and overwriting. Concurrent same-turn expansions of different groups both persist.

New test: "concurrent auto-expands of different groups keep both in toolState" — `Promise.all` over two different deferred groups; both resolve and both groups land in `toolState`. (Reproduced failing pre-fix: one expansion was dropped, non-deterministically.)

### Verification

- `bun test test/tool/auto-expand.test.ts` — 17/17 pass
- Related suites: `exposure`, `processor`, `tool-call-execution`, `observability` — pass
- Pre-push hooks: prettier, oxlint, turbo typecheck (12 packages), sherif, doc/decision checks — all pass

yzxoi:
@codex review

yzxoi:
## Addressed Codex inline review (944eee14) — validate auto-expanded args before dispatch

The P1 inline comment (processor.ts:1560) was valid: deferred tools are absent from `toolDefinitions` (the diagnostic stub carries an open schema), so the AI SDK never validated model arguments, and MCP/plugin execution paths do not revalidate — malformed arguments reached external implementations.

**Fix** (TDD: failing test first, then green):
- `ToolResolver.validateToolInput()` — Ajv 2020 validation against the freshly resolved `inputSchema`; uncompilable schemas fall back to prior behavior.
- `runtimeToolFor`/`autoExpandTool` now expose the resolved definition's `inputSchema` in `AutoExpandedTool`.
- Processor validates `call.input` after auto-expand and before `ToolScheduler.dispatch`; failures settle the part with the standard `invalid_arguments` diagnostic and skip dispatch (real `execute` never runs).

**Regression test**: "auto-expanded MCP tools validate model arguments before dispatch" — malformed args (`value: "not-a-number"` vs `type: number`) → part settles `error` with "invalid arguments", `executeCalls === 0`, expansion still persists.

**Verification**: `auto-expand.test.ts` 18/18, coverage-run + auto-expand 24/24, tsgo clean, all pre-push hooks pass.

Also merged latest `dev` (7fad093c, #1190) into the PR branch and resolved the coverage-run isolated-list conflicts; the pinned-list test order was corrected after the merge (cd61934a).

yzxoi (approved):

yzxoi:
@synergy-agent review

PR #1192, opened 2026-08-18
feat(app): add multi-session Kanban board view

yzxoi:
## Summary

- Add a **Kanban** board page (`/kanban`) reachable from a fixed sidebar entry (order 25, between Agenda and Library) and the mobile drawer, streaming up to **6 sessions' live messages** on one screen.
- **Three layouts** (grid, focus+rail, waterfall) switchable from the toolbar; layout, per-pane follow toggles, and the pinned pane set persist via localStorage (`Persist.global`).
- **Mixed pane policy**: pinned sessions always occupy slots; leftover pinned keys whose session vanished become removable "unavailable" placeholders; running + waiting sessions auto-fill remaining slots by last activity.
- **Pure frontend view-layer change**: reuses the existing global event stream, per-Scope stores, layout nav lists, `messagePage` loader pattern, and shared UI message components (`SessionTurn`/`MailboxMessage`/`CommandResultOutput`) — no server/SDK/protocol change.

## Sync-layer generalizations (minimal, backward compatible)

- **Eviction protection set**: `global-sync.tsx` now unions a `protectedBucketKeys` set into `planBucketEviction` alongside the single `activeBucketKey`; the board calls `protectMessageBucket` on pane mount and `unprotectMessageBucket` on unmount, so live panes survive LRU eviction while the single-active-session semantics stay untouched.
- **Volatile resync set**: `session-volatile-resync.ts` takes `activeBucketKeys: string[]` and returns `activeSessionIDs`; `refreshVolatileAfterResync` batch-refreshes inbox/todo/dag for the active session plus every board pane after reconnect.

## Files

- New: `packages/app/src/components/kanban/` (panel, model, pane, layout), `packages/app/test/components/kanban/`, `docs/decisions/implemented/feature/2026-08-18-session-kanban.md`
- Changed: `app.tsx` (route + boot gate), `builtin-navigation.tsx`, `mobile-drawer.tsx`, `global-sync.tsx`, `session-volatile-resync.ts`, i18n catalogs (en/zh-CN/pseudo), `semantic-icon.tsx` + `icon.tsx` + `builtin-icons.ts` (new `kanban.main` token), `frontend-data-sync.md`

## Verification

- Narrow tests green: `pane-selection` (6), `message-eviction` + `session-volatile-resync` (10), semantic-icon contract (5)
- `bun run --cwd packages/app test` — full suite green incl. browserOnly/extraSerial groups
- `bun run --cwd packages/app typecheck` ✅ · `i18n:check` ✅ (zh-CN translated) · `build` ✅ · `decision:check` ✅ · `test-layout:check` ✅
- `quality:quick`: 13/14 gates pass; `secrets:check` skipped locally (gitleaks not installed — environment, not code)


yzxoi:
## Review fixes applied (commit `0c087d6b33`)

**Blocking items**

1. **Sidebar position fixed** — kanban entry `order` 25 → 15, so it renders between Agenda (10) and Library (20). Verified in the browser DOM: sidebar order is now Agenda → Kanban → Library.
2. **Waterfall layout aligned to spec** — `.kanban-waterfall` is now a fixed 3-column grid (`repeat(3, minmax(0, 1fr))`). Time alignment: each pane renders per-message timestamps (`.kanban-msg-time`, `toLocaleTimeString` HH:MM) above every message, so the three streams are chronologically aligned for at-a-glance pace comparison. Verified computed styles in a real browser: 3 columns at 1280px, 1 column at 375px (`<768px` degradation).
3. **End-to-end verification** — ran the isolated second-runtime browser smoke on the `/kanban` route (dev stack on ports 4097/3001):
   - Cold boot loads `/kanban` without waiting for a session surface (HTTP 200, UI renders).
   - Empty state renders ("No sessions on the board" + hint) with the Grid/Focus/Waterfall layout switcher.
   - Layout switch to Waterfall persists across reload (`synergy.global.dat:kanban` = `{"layout":"waterfall",...}` in localStorage; `data-active="true"` restored).
   - Full multi-session concurrent-streaming acceptance (2+ live sessions, follow independence, LRU pressure, pane→session navigation) could not be executed: the sandbox's restricted network blocks LLM provider calls (`DNSException: getaddrinfo ENOTFOUND`), so no live session can be created in the isolated home. This is an environment limitation, not a code one. In its place, the board-loader behavior layer now has injected-fake-loader unit tests (5 tests): protect-on-load lifecycle, syncPanes dedupe, reconnect-version force reload, unprotect-on-dispose, and load cap.

**Non-blocking items**

4. **Reduced-motion handled** — `@media (prefers-reduced-motion: reduce) { .kanban-dot[data-pulse] { animation: none } }`. Verified in browser with `emulateMedia({ reducedMotion: 'reduce' })`: `animation-name: none`.
5. **No nested `<button>` in focus rail** — promote control is now `<div role="button" tabindex="0">` with Enter/Space key handling and `:focus-visible` outline; the inner pane's real buttons (title/follow/pin/remove) are no longer nested in an outer button.
6. **Test fixture residue** — 3 leftover `.menu-field-fixture-*` / `.boss-mode-panel-fixture-*` temp dirs remain untracked in the worktree; they are not part of any commit. Cleanup requires shell-destructive permission, which the autonomous profile denies; they are harmless (prettier-formatted, gitignored from commits).

**Gates re-run after fixes (all green)**
- `bun run test` (official app suite, packages/app): exit 0
- Narrow: board-loader (5), pane-selection (6), message-eviction (7), session-volatile-resync (3): 21 pass
- `tsgo -b` typecheck, `i18n:check`, `decision:check`, `test-layout:check`, `doc:check`, `format:check` (root prettier): all pass

yzxoi:
## Codex review fixes applied (commit `d800268304`)

The PR's Codex review contained 19 inline comments. **2 were already fixed in the previous review round** (nested `<button>` in the focus rail → `div role="button"`; reduced-motion pulse). The remaining **17 were all valid and are now fixed**:

### P1 fixes

1. **Route panes via resolved scope key** (`index.tsx`) — `openSession` now navigates with `pane.scopeKey` (the worktree directory resolved by `scopeKeyForNavEntry`), matching sidebar/mobile-drawer route construction instead of using the raw `entry.scopeID`.
2. **No duplicate turn rendering** (`pane.tsx`) — the pane now builds a timeline like the normal session surface: root user messages render as `SessionTurn` rows (with `buildSessionTurnProjection`), mailbox/action-command assistants render as standalone rows, and ordinary turn members render only inside their turn.
3. **Auto-scroll bound to the viewport** (`pane.tsx`) — `ConversationViewport` now receives `setScrollRef={(el) => autoScroll.scrollRef(el)}`; streaming follows and the scroll-to-bottom control works, so the Follow toggle is functional.
4. **Release protection when panes leave the board** (`board-loader.ts`) — `syncPanes` diffs previous vs. wanted keys and unprotects removed ones; a `dirty` set reloads re-joined panes that may have been evicted.
5. **Volatile freshness captured before the request** (`global-sync.tsx`) — inbox/todo/DAG request tokens are captured before `volatileBatch` starts, so events arriving in flight supersede the older snapshot.
6. **Latest root passed to SessionTurn** (`pane.tsx`) — `lastUserMessageID={lastRoot()?.id}` enables working-state presentation for the active turn.
7. **Delete parts for dropped messages** (`board-loader.ts`) — `plan.droppedIds` part buckets are deleted in the batched update, mirroring the session loader.
8. **Bound rendered turns per pane** (`pane.tsx`) — latest-mode trim to 40 turns with `messagesFrom` canonicalization, per pane.

### P2 fixes

9. **No unconditional reload on navigation updates** (`board-loader.ts`) — `syncPanes` skips panes already in a non-idle phase; only reconnect version bumps force reloads.
10. **Propagate message-page failures** (`board-loader.ts` + `pane.tsx`) — the `.catch(() => undefined)` swallow was removed; load errors surface an error state with a Retry button (`boardLoader.load(force)`).
11. **Unavailable panes render without a Scope store** (`index.tsx`) — unavailable placeholders render from a static empty data object; they never enter the loader.
12. **Add-session chooser** (`index.tsx`) — the Add button opens a Popover listing every unpinned source (including idle sessions) with title + scope, so users pick the session to pin.
13. **Pane identity preserved across source updates** (`grid.tsx`/`focus.tsx`/`waterfall.tsx` + `pane-selection.ts`) — layouts render through `buildPaneSnapshot` keyed by the stable `pane.key`, so status/navigation updates never unmount and rebuild the message tree.
14. **Focus rail shows every pane** (`focus.tsx`) — removed the `.slice(0, 3)`; the rail renders all non-active panes and relies on its overflow scrolling.
15. **Localized relative activity label** (`pane.tsx`) — replaced the hard-coded `m`/`h`/`d` label with `fmt.relative` from the locale context, refreshed on a 1-minute cadence.
16. **Waterfall vertically scrollable** (`kanban.css`) — the grid now has `overflow-y: auto` like the grid layout.
17. **i18n** — added `app.kanban.loadError` / `app.kanban.retry` descriptors with zh-CN translations.

### Verification

- `tsgo -b` typecheck: exit 0
- Narrow tests: 32 pass (board-loader 7 incl. new skip-ready + unprotect-on-leave, pane-selection 6, message-eviction 7, volatile-resync 3, message-loader 4, part-snapshot 5)
- Official app suite `bun run test`: exit 0
- `bun run build`: ✓ built
- `i18n:check`, `decision:check`, `test-layout:check`, `format:check` (root prettier): all pass

yzxoi:
## Waterfall removed + per-pane composer (commit `3680b71040`)

Two changes based on hands-on testing feedback:

### 1. Waterfall layout removed

The waterfall view was dropped entirely:
- `layout/waterfall.tsx` deleted (component + `kanban-waterfall` CSS)
- `KanbanLayout` narrowed to `"grid" | "focus"`; persisted `migrateKanbanPreferences` maps any stored waterfall value back to grid
- `app.kanban.layout.waterfall` i18n descriptor removed from en/zh-CN/pseudo catalogs
- The pane `timeAlign` timestamp variant was removed with it

The board now has two layouts: **Grid** (adaptive multi-pane cards) and **Focus** (main pane + scrollable rail).

### 2. Per-pane composer (send messages from the board)

Each live pane now has a minimal composer at the bottom:
- Text input + send button (Enter submits; button disabled while empty/sending)
- Posts through `client.session.input({ sessionID, parts: [{ type: "text", text }] })` using the pane's resolved scope credentials (`directory` for project scopes, `scopeID` for HOME)
- `agent`/`model` are omitted — the server falls back to the session's last used values, so the board stays a thin send surface
- Send failures surface a toast (`app.kanban.sendFailed` + zh-CN "发送消息失败")
- New i18n keys: `app.kanban.sendPlaceholder` ("Message this session…" / "向此会话发送消息…"), `app.kanban.sendFailed`

The decision record was updated: two layouts, board is now a live control surface (plain-text input) rather than read-only.

### Verification

- `tsgo -b` typecheck: exit 0
- Narrow tests: 23 pass (board-loader 7, pane-selection 6, message-eviction 7, volatile-resync 3)
- Official app suite `bun run test`: exit 0
- `bun run build`: ✓
- `i18n:check`, `format:check` (root prettier), `decision:check`, `doc:check`: all pass

yzxoi:
## Grid customization + full pane composer

Implements the two follow-ups requested on the board:

**Configurable grid layout**
- Grid mode now has Columns (1–4) and Rows (1–4) selects in the toolbar; `gridCols`/`gridRows` are persisted and clamped on migrate.
- New **Free layout** toggle: panes auto-place on a dense 4-column grid and each pane gets a size menu (1×1 / 2×1 / 1×2 / 2×2) that sets `grid-column/row: span N`. Per-pane spans persist in `paneSpans`.
- Pure module `model/preferences.ts` holds the persisted schema, migration, and span helpers with unit tests (`test/components/kanban/model/preferences.test.ts`).

**Full-featured pane composer** (replaces the thin send box)
- Agent picker (filters hidden/subagent agents; applies to the next send via `session.input({ agent })`).
- Permission selector (Guarded / Autonomous / Full Access) via `session.update({ controlProfile })`, resolving pending permissions on Full Access.
- Orchestration mode menu (Default / Plan / Lattice / Boss) via `workflow.session.set` (lattice packaged as `{ kind, mode: "auto" }`).
- Status bar under the input: live session status dot/text plus agent · permission · mode metadata.
- All writes go through scope-scoped `createSynergyClient` calls injected per pane (no nested SDK/Sync providers).

**Validation**: `typecheck`, `i18n:check` (strict localization contract, incl. zh-CN), full `packages/app` test suite (1585+ tests), production build, `format:check`, `decision:check`, `test-layout:check` all pass. Dynamic descriptors route through `MessageDescriptor`-typed helpers per the repo's localization contract.

yzxoi:
## Free layout removed + sidebar drag-to-pin

Per follow-up feedback:

**Free layout removed**
- Dropped the "Free layout" toggle, the per-pane size (span) menu, and the `freeLayout`/`paneSpans` preference fields; `migrateKanbanPreferences` now drops the retired fields from stored values.
- Grid remains the user-set columns × rows matrix (rows split the viewport height evenly, overflow panes land on fixed-height implicit rows).

**Sidebar drag-to-pin**
- Sidebar session rows are now `draggable`; `dragstart` writes a shared `application/x-synergy-session` JSON payload (scope key + session ID) via `packages/app/src/utils/session-drag.ts`.
- The kanban panel accepts the drop (`dragover` + `drop` on the panel root), parses the payload, and pins the session — with a dashed outline while a session is dragged over the board.
- The empty-state hint now says "Drag a session from the sidebar to pin it"; the Add-session picker still works as an alternative.

**Validation**: `typecheck`, `i18n:check` (strict, catalogs re-extracted), 18 kanban + 93 sidebar/prompt + 5 preferences tests, `prettier`, `decision:check` all pass.

yzxoi:
## Fix: pinned panes no longer disappear after browsing other sessions

**Root cause**: the board protected every pane's message bucket against LRU eviction, but the protection was released when the board unmounted. While the user browsed other sessions, the bucket was evicted by the 15-bucket LRU; returning to the board rendered the pane empty (nothing refetched it) until a manual retry.

**Fix (per review direction — no special kanban LRU protection)**:
- Removed the `protectedBucketKeys` set and `protectMessageBucket`/`unprotectMessageBucket` from `global-sync.tsx`; only the actively viewed session is exempt from eviction.
- Board panes now enter the **normal load path when the board mounts** — loading touches their bucket, keeping visible panes near the LRU head.
- `board-loader.syncPanes` marks panes that left the board as dirty and **refetches them when they rejoin**, so a pane can never silently show a blanked timeline after eviction.
- Docs updated (`frontend-data-sync.md` eviction section, kanban decision record); `board-loader` tests updated to cover the leave-and-rejoin refetch.

**Validation**: `typecheck`, 27 kanban/eviction/resync tests (incl. new rejoin-refetch test), `prettier`, `decision:check` all pass.

yzxoi:
## Kanban mirrors session display settings + sidebar recent order

Two follow-up fixes:

**1. Shared display settings (no more hardcoded minimal activity display)**
- Panes previously rendered `SessionTurn` with `activityDisplay="minimal"` hardcoded; now `activityDisplay` and `compactReasoning` resolve from the global config (`resolveActivityDisplay(globalSync.data.config.activityDisplay)` / `config.compactReasoning === true`), exactly like the session page. Toggling Activity display or Compact reasoning in Settings now affects the board too.

**2. Unread/idle sessions appear on the board, following sidebar recent order**
- Pane selection changed from "pinned + running/waiting only" to: **pinned sessions occupy the leading slots** (pinned order), and **every remaining slot fills with the other visible sessions in sidebar recent order** (most recently active first) — including idle and unread sessions, mirroring the sidebar.
- Board capacity follows the active layout: grid shows exactly `columns × rows`; focus keeps the board cap.
- Empty-state hint updated ("Sessions appear here in your recent order…") in en + zh-CN.

**Validation**: `typecheck`, 90 kanban/sidebar tests (pane-selection tests updated for the new ordering), `i18n:check` (strict), `prettier`, `decision:check` all pass.

yzxoi:
## Kanban pane interactions: distinct pin icon, FLIP animation, drag-reorder

Four interaction improvements on the board:

**1. Pin/unpin icons are now distinct**
- Pinned panes show `pin-off` (via the new `action.unpin` semantic token), unpinned show `pin` — no more identical icon for both states.

**2. FLIP position animations**
- New `FlipPanes` 2D FLIP container (`packages/app/src/components/kanban/flip.tsx`) animates panes to their new positions whenever the pane order changes: reorder swaps, focus promotion, overflow reflow, grid re-split. Mirrors the sidebar `FlipList` timing, respects `prefers-reduced-motion`.

**3. Pinned panes drag to reorder**
- Each live pane gets a **grip handle**; dragging it onto another pinned pane reorders the pinned list (new `application/x-synergy-kanban-reorder` drag MIME, `reorderPinnedKeys` helper, persisted order). Works in grid and focus layouts; reorder-triggered FLIP animates the swap.

**4. Unpinned panes can't be dragged — with a hint**
- The grip renders locked (`data-locked`, dimmed, `not-allowed` cursor) for unpinned panes; attempting to drag it shows a toast: "Pin this session to reorder it" instead of moving anything.

New i18n keys (`app.kanban.dragReorder`, `app.kanban.pinToReorderHint`) with zh-CN translations; `reorderPinnedKeys` covered by unit tests.

**Validation**: `typecheck`, 19 kanban tests (incl. reorder helpers), `i18n:check` (strict), `prettier`, `decision:check` all pass.

yzxoi:
## Fix: kanban grip drag-reorder now actually drags

**Root cause**: SolidJS treated `draggable` as a boolean attribute and rendered `draggable={true}` (and bare `draggable`) as `draggable=""`. Per the HTML spec `draggable` is an *enumerable* attribute — the empty string is invalid, so the element was never actually draggable and `dragstart` never fired. Verified in the browser: zero drag events, no reorder.

**Fix**: pass explicit strings — `draggable="true"` on the kanban grip, `draggable={props.flyout ? "false" : "true"}` on sidebar session rows. The sidebar rows had the same latent bug, so sidebar drag-to-pin is also repaired.

**Verified in browser**: full drag sequence (`dragstart` → `dragover` → `drop` → `dragend`) with the `application/x-synergy-kanban-reorder` payload; pinned order swapped on drop (FLIP animation included). `typecheck`, 19 kanban tests, prettier all pass.

yzxoi:
## Fix: consistent FLIP animation + sidebar drops pin to the front

**1. FLIP animation was intermittent**
Root cause: the FLIP container measured pane positions synchronously inside a render effect, but Solid's keyed `<For>` below commits its DOM moves in a *later* effect. When the reorder effect ran first, the measurement read pre-move positions → zero delta → animation skipped (timing race explains the "sometimes works, sometimes not"). Fix: defer measurement to a `requestAnimationFrame` (after the DOM commit) and cancel in-flight animations before re-measuring, mirroring `FlipList`.

**2. Sidebar drop didn't look like it pinned**
Root cause: drops appended to the **end** of the pinned list. With the board at capacity (e.g. 1×3 grid = 3 panes, all pinned), the newly dropped session landed beyond the visible cap — nothing appeared to change, so it seemed the drop didn't pin. Fix: new pins (sidebar drag + Add-session) go to the **front**, immediately occupying the first slot even when the board is full.

**Validation**: `typecheck`, 19 kanban tests, prettier, decision check all pass.

yzxoi:
@synergy-agent review

yzxoi:
## Review fixes applied (commit `3cc1496752`)

All four open items from the latest review are fixed:

**1. Blocking — undefined CSS tokens in `kanban.css`**
The board referenced 10 CSS variables that are never defined anywhere (`--status-*`, `--font-size-lg/md/sm/xs`, `--text-strong-hover`), so status dots rendered transparent and the type scale silently fell back. Now mapped to canonical theme tokens:
- Font sizes → `--font-size-large/small/base/x-small`
- Status dots → `--surface-success-base` / `--surface-warning-base` / `--surface-info-base` / `--surface-brand-base` / `--text-weaker`
- Title hover → `--text-stronger`

`kanban.css` is now included in the `css-token-integrity` test's app-side coverage, so this class of regression is caught in CI.

**2. High — first-paint blank panes**
`renderPane` returned `null` when the scope store didn't exist yet (created lazily by the loader). It now falls back to `EMPTY_BOARD_PANE_DATA`, so the pane shell (header, loading state, actions) renders unconditionally and fills in when the first message page lands.

**3. Medium — follow toggle didn't affect autoscroll**
`createAutoScroll` is now keyed on `follow && working`, so "Paused" actually stops the stream from scrolling; the viewport's manual scroll-to-bottom button still works.

**4. Nit — vestigial `activeBucketKeys` plural API**
Simplified `session-volatile-resync` to a single `activeBucketKey` (the board dropped its per-pane protection with the LRU change); tests updated.

**Bonus — semantic-icon glyph collision fixed**
The dev merge exposed `notes.pin` and `action.unpin` both mapping to `pin-off` (unique-glyph contract violation), and the kanban grip used a raw Lucide literal. Registered `bookmark` (`notes.pin` now uses it), kept `action.unpin` on `pin-off`, and added `action.grip`.

**Validation**: `typecheck` ✅, 24 ui tests (semantic-icon + css-token-integrity incl. new kanban.css coverage) ✅, 24 app tests (volatile-resync + kanban) ✅, `i18n:check` strict ✅, prettier ✅, decision validation ✅.

yzxoi:
## Fix: pin/unpin now visibly distinct and reactive in grid panes

Two-part fix for "pin and unpin look the same":

**1. Reactive bug (the real culprit)** — `KanbanGrid` rendered panes through a keyed `<For>` whose callback captured `const current = pane()` once. Toggling pin/unpin changes only the `pinned` flag (key set and order are unchanged), so the closure never re-ran and the button stayed stale — icon, title, and `data-active` kept showing the old state even though the store and localStorage updated. Verified in the browser: before the fix, clicking Pin updated localStorage but the DOM didn't move; after, the pane re-renders immediately. Rendering now goes through a reactive `<Show>` so pane-object changes propagate.

**2. Visual distinction** — pinned action buttons now get a brand-filled background (`--surface-brand-base` / `--text-on-brand-base`, `--surface-brand-hover` on hover) instead of only a subtle text-color change, so pinned vs unpinned is obvious at a glance alongside the `pin` vs `pin-off` glyphs.

**Validation**: `typecheck`, 21 kanban tests, 24 ui tests (semantic-icon + css-token-integrity with the new brand tokens), prettier, decision validation all pass.

yzxoi:
## Fix: sending from a pane can no longer drop the previous assistant reply

**Symptom**: after sending a message from a kanban pane, the previous assistant message sometimes disappeared (until reload).

**Root cause** (compared against the session page's submit path in `prompt-input/submit.ts`): the session page **invalidates the message resource before `session.input` lands** (in `addOptimisticMessage`). The board composer called `session.input` directly without that invalidation. When a message-page snapshot was already in flight (captured before the send), SSE could merge the new user + assistant messages (bumping the freshness revision), and then the stale snapshot returned — its revision no longer matched, but with a version header it could still be accepted — replacing the whole window with the older contents. The just-generated assistant turn was dropped until a reload refetched.

**Fix**: `sendToPane` now calls `globalSync.invalidateResource(pane.scopeKey, pane.sessionID, "message")` before `session.input`, matching the session page. Any in-flight snapshot is rejected as superseded; the loader retries and applies the fresh page.

**Validation**: `typecheck`, 55 kanban/freshness/message-page tests, prettier, staged-lint, decision validation all pass.

yzxoi:
## Review findings resolved

All findings from the four review reports are fixed in `af859c403b`:

**Grid cap**
- Grid rows capped at 3 → 4×3 = 12 panes max; stored `gridRows` above 3 clamp through the existing preferences migration.

**Reliability**
- New global eviction-version signal: a visible pane whose message bucket is LRU-evicted now refetches instead of sitting on a stale loading shell.
- Composer profile switching guards against in-flight double updates.

**Performance**
- `FlipPanes` gates FLIP on the pane-key sequence (memo), so nav-store churn no longer triggers `getBoundingClientRect` sweeps on unchanged layouts; positions refresh via `ResizeObserver` on container resize.
- Offscreen rows get `content-visibility: auto` + `contain-intrinsic-size`; `.kanban-pane-body` gets `contain: layout style`.

**Maintainability**
- Removed dead `removePane` descriptor + three catalog entries; catalogs re-extracted.
- Renamed vestigial `kanban-pane-span-menu/item` → `kanban-composer-menu/item` (CSS moved to the composer section).
- `isActionCommandMessage` shared from `session-message-order.ts`; dropped unused import in `session.tsx`/`index.tsx`.
- Fixed stale decision-record text, misleading test names/comments, and coverage-exempt reasons.

**Tests**
- Added `parseSessionDragPayload` unit tests (canonical, legacy, malformed payloads); drag-drop pinning validates payloads against visible sources.

Verification: app typecheck, `i18n:check`, kanban/eviction/drag/volatile-resync tests, CSS contract test, ui semantic-icon tests, prettier, `decision:check` — all pass.

PR #1193, opened 2026-08-18
fix(workspace-files): resolve workspace root before realpath containment check

yzxoi:
## Summary
- **Root cause fix**: `WorkspaceFileService.assertRealpathInside()` compared each target's `fs.realpath()` result against the **lexical** workspace root (`path.resolve(ScopeContext.current.directory)`, never realpath-resolved). When the scope directory is reached through a symlink (e.g. `~/workspace` itself is a link, or the project path contains a symlinked component), every file's real path falls outside the lexical root, so `children` / `read` / `stat` returned `Access denied: real path escapes workspace` (500) for the whole workspace — surfacing immediately when opening a file in the Files panel.
- The fix resolves **both** sides of the containment check: the target path and the workspace root are both passed through `fs.realpath()` before `isPathContained()` compares them. If the root cannot be resolved, the lexical root is used as a fallback, so access is never silently widened. Symlinks inside the workspace whose target escapes the resolved root are still rejected — the security property is unchanged.
- **Error contract alignment**: `children`, `read`, and `stat` now catch `AccessDeniedError` and return a structured 403 (same `{name, data}` shape as the existing `write` route) instead of leaking the error into the generic 500 handler.

## Why
Opening a freshly created file under `docs/` in the workspace Files sidebar 500'd for users whose workspace directory is a symlink (or contains one in a path component). The check was asymmetric: real path vs. lexical root. This is a false positive in the access check, not a real escape.

## Verification
- New tests (fail on unpatched code, pass with the fix):
  - `test/workspace-file/service.test.ts` — "accepts files when the workspace root itself is a symlink" (node/children/read all work through a symlinked root); "still rejects directory symlinks whose target escapes the workspace" (children/node/read all still throw).
  - `test/server/workspace-files.test.ts` — HTTP-level: symlinked root serves children/read with 200; escaping directory symlink returns **403** (was 500) with `WorkspaceFileAccessDeniedError` for children/read/stat.
- Confirmed the new root-symlink test fails against the unpatched `assertRealpathInside` (observed `WorkspaceFileAccessDeniedError`) and passes with the fix.
- Test runs: `packages/synergy` — `test/workspace-file/service.test.ts` 28 pass, `test/server/workspace-files.test.ts` 19 pass, 0 fail.
- Gates: `format:check`, `lint` (0 warnings), `decision:check`, `test-layout:check`, `typecheck` (11/11) all green; SDK regenerated via `script/generate.ts` (openapi.json + types.gen.ts include the new 403 responses).
- Decision record added: `docs/decisions/implemented/bug-fix/2026-08-18-workspace-files-realpath-root.md`.

## Files changed
- `packages/synergy/src/workspace-file/service.ts` — realpath-resolve the root in `assertRealpathInside`
- `packages/synergy/src/server/workspace-files.ts` — 403 handling for children/read/stat + OpenAPI 403 responses
- `packages/synergy/test/workspace-file/service.test.ts`, `packages/synergy/test/server/workspace-files.test.ts` — new coverage
- `packages/sdk/openapi.json`, `packages/sdk/js/src/gen/types.gen.ts` — regenerated
- `docs/decisions/implemented/bug-fix/2026-08-18-workspace-files-realpath-root.md` — decision record


yzxoi:
## Semantics note: escape-symlink visibility is relative to the workspace root (verified)

I reproduced this PR's scenario in an isolated dev instance (symlinked workspace root) and want to record the containment semantics the fix preserves, since the same filesystem setup can look contradictory depending on which root the scope uses.

Setup:
- `real-workspace/` contains `escape -> ../outside-dir` (target sits inside the **parent** directory)
- `workspace-link -> real-workspace` (the root symlink)

Same symlink, two different workspace roots:

| Workspace root | Request | Result |
|---|---|---|
| Parent dir (`.../pr1193`) | `children?path=workspace-link/escape` | **200** — escape visible and traversable |
| `workspace-link` (real root `real-workspace`) | `children?path=escape` | **403** `WorkspaceFileAccessDeniedError`; also omitted from the root `children` listing |

So the containment check is correctly **root-relative**: the same `escape` symlink is "inside" the parent workspace (target under the root) and "escaping" the link workspace (target outside the real root). This PR only removes the false positive when the root itself is reached through a symlink; it does not change the escape semantics — which matches the security property stated in the summary.

One non-blocking UX observation from the same session: when a directory symlink **does** escape, `children()` silently drops the entry (`node()` throws → `.catch(() => undefined)` → filtered out), so the Files panel shows the folder as "missing" rather than "denied". Direct access to the path returns the new structured 403. Silently hiding is defensible from a security standpoint (don't leak that a path exists outside the root), but if the UI should ever distinguish "not found" from "no access", `children` would need a denied marker on the entry. Purely optional; not a blocker for this PR.

PR #1194, opened 2026-08-18
fix(app): seed note editor content when TipTap becomes ready

EricSanchezok:
## Summary

- Fix Notes/Blueprint open editor showing only the slash placeholder while the list card already has `previewHtml`.
- Seed the saved TipTap document when the editor becomes ready, so `applySnapshot()` no longer races ahead of mount and no-ops.
- Restore the open-note flex height chain so long Blueprints scroll inside TipTap instead of collapsing under a parent `overflow-y-auto` column.
- Add focused App regressions plus an implemented decision record for the frontend load path.

## Problem

Users saw Blueprints such as “剧本格式规格…” with a populated list thumbnail, but opening the note showed an empty editor body (only `输入 / 查看命令...`), while title/tags/Saved remained visible.

Runtime inspection showed the backend note was intact (full ProseMirror doc, dozens of top-level nodes). List cards render metadata `previewHtml`; the open editor uses a separate TipTap load path. Two frontend defects stacked:

1. Race: `NoteEditor.applySnapshot()` called `replaceEditorContent()` before `DocumentEditorCore` finished `onMount` / `onEditorReady`. With no editor instance yet, replacement no-oped and nothing re-seeded content after TipTap mounted.
2. Layout: title + editor + tags lived in one scrolling column. `DocumentEditorCore` depends on `flex-1 min-h-0` + inner `h-full`; nesting that chain under parent `overflow-y-auto` collapsed the editor viewport for long documents.

## Changes

- `packages/app/src/components/note/note-sync.ts`
- Add `isEmptyEditorDoc()` so blank starter TipTap docs can be distinguished from real content.
- `packages/app/src/components/note/panel.tsx`
- Seed the loaded snapshot from `onEditorReady` via `handleEditorReady` / `seedEditorFromSnapshot` (`setContent(..., { emitUpdate: false })`, no dirty/autosave).
- Split the open-note shell into fixed title, `min-h-0 flex-1` editor host, and fixed tags; keep scrolling inside TipTap.
- Tests
- Empty-doc detection coverage in `note-sync.test.ts`.
- TipTap acceptance of blockId-bearing note JSON in `document-editor-core.test.ts`.
- Decision record
- `docs/decisions/implemented/bug-fix/2026-08-18-note-editor-seed-on-ready.md`

## Test plan

- [x] `cd packages/app && bun test test/components/note/note-sync.test.ts`
- [x] `cd packages/app && bun test --conditions=browser --timeout 30000 test/components/note/document-editor-core.test.ts`
- [x] `bun run decision:check`
- [ ] Manual: open a Blueprint whose list card already shows preview content and confirm the editor body matches the thumbnail after refresh.

## Notes

- Backend storage / `note_read` were not empty; this is a frontend open-path fix only.
- Ready-time seed intentionally does not mark dirty or emit autosave updates.
- Local dirty content is not overwritten by seed.


yzxoi:
## Review follow-up: pushed a one-line layout fix

Reviewed the PR and found the editor height chain still broken: the new editor host wrapper (`min-h-0 flex-1 px-4 py-4`) is a **block** container, so `DocumentEditorCore`'s `flex-1` root is ignored and grows to the full document height. Browser-measured with the exact class strings on a 100-paragraph doc (765px panel):

| Layout | editor root height | internal scroll |
|---|---|---|
| Before PR (single `overflow-y-auto` column) | 646px | yes |
| PR as merged (block host) | **18314px (full content)** | no — long docs overflow, tags pushed out of view |
| Host as flex column (`flex min-h-0 flex-1 flex-col px-4 py-4`) | 647px | yes |

Pushed commit `a03138fde` to `synergy/session-feccf0` applying the fix:

```tsx
<div class="flex min-h-0 flex-1 flex-col px-4 py-4">
```

Verified: `note-sync.test.ts` 9/9, `document-editor-core.test.ts` 4/4, `tsgo -b` typecheck, `decision:check`, and pre-push hooks all pass.

Minor nits (non-blocking, can address in follow-up):
- `document-editor-core.test.ts` new test title says "includes blockId attrs" but `blockId` is not a registered schema attribute — ProseMirror drops it on load, and assertions only check acceptance/text. Suggest renaming to "accepts note JSON with unknown blockId attrs" or registering the attribute if round-trip is intended.
- The decision record's Consequences claims the flex chain is restored; with the fix commit it now matches reality. Worth re-reading the wording once merged.

yzxoi (approved):
Approved ✅ Thanks for the effort on this one! The ready-time seed cleanly fixes the empty-editor race, and the flex host change keeps long notes/Blueprints scrolling inside TipTap. Focused tests and typecheck all pass. Great work!

PR #1196, opened 2026-08-18
fix(cortex): inherit parent session model when agent has none

yzxoi:
## Summary

`Cortex.runTask()` was the only model-resolution point in the codebase without a fallback chain: explicit task model → agent model → hard failure (`No model configured for agent <name>`). A custom subagent without a `model`/`modelRole` (e.g. a user-configured BlueprintLoop `auditAgent` such as a `reviewer` agent) could be selected and launched successfully, then fail at run time. The Blueprint continuation policy retried the same unconfigured reviewer and the loop ended `failed` with `review_terminal_tool_missing`, surfacing as a stuck `[Cortex] [Review] Audit BlueprintLoop …` child session.

Fix: fall back to the parent session's last model, aligning with the other delegation paths (`session/input.ts`, `tool/task.ts`, `plugin/host-services.ts`).

## Changes

- `packages/synergy/src/cortex/manager.ts`: model resolution is now `model ?? Agent.getAvailableModel(agent) ?? lastModel(task.parentSessionID)`.
- `packages/synergy/test/cortex/manager.test.ts`: regression test — forces `Agent.getAvailableModel` to return `undefined` and asserts the task receives the parent session model.
- `docs/decisions/implemented/bug-fix/2026-08-18-cortex-parent-model-inheritance.md`: decision record.

## Verification

- `bun test test/cortex/ test/session/blueprint-continuation.test.ts test/session/light-loop-continuation.test.ts` — 165 pass / 0 fail
- `bun run typecheck` — pass
- Prettier on changed files — pass
- `bun run decision:check` — pass
- Note: `quality:quick` reports 2 environment-only failures unrelated to this change: `secrets:check` (gitleaks not installed locally) and `workflow:check` (actionlint download timed out).


yzxoi:
Review feedback addressed. Thanks for the rigorous controlled experiment — it confirmed a real regression I had misattributed as pre-existing flakiness.

## Verification of your finding (reproduced locally)

| State | `dag-result.test.ts` |
|---|---|
| Fallback removed (your base) | 4.36s, 11/11 pass |
| Fallback present (PR code, before fix) | ~5.7s, 327/399 tests attempt real LLM network calls; timed out at 5s in combined runs |

Your diagnosis was correct: with the fallback, `lastModel` bottoms out at `Provider.defaultModel()`, which resolves against the seeded catalog fixture, and the task proceeds to a real `LLM.stream` with no API keys — network-restricted CI hits the per-test timeout.

## Fix (commit 5f6cbf6988)

1. **`test/cortex/dag-result.test.ts`** — the four delegated-subagent tests that launch `Cortex.launch` without an explicit model now pass `model: { providerID: "test-provider", modelID: "test-model" }` and mock `SessionInvoke.invokeInternal` (matching the pattern already used by the sibling tests at :230-238 / :304-313). Assertions unchanged.
2. **`src/cortex/manager.ts`** — adopted your minor suggestion: `lastModel(...)` is now wrapped in `.catch(() => undefined)`, keeping the explicit `No model configured for agent` failure mode if the parent session is missing/deleted instead of an unhelpful error.

## Verification after fix

- `bun test test/cortex/dag-result.test.ts` — 11/11 pass in 4.23s (deterministic, no network path)
- `bun test test/cortex/ test/session/blueprint-continuation.test.ts test/session/light-loop-continuation.test.ts` — 165 pass / 0 fail
- `bun run typecheck` (tsgo) — pass

PR #1197, opened 2026-08-18
feat(tools): expose connected MCP group directory in expand_tools description

EricSanchezok:
## Summary

When the total connected MCP tool count reaches the defer threshold (100), every MCP tool is folded behind a `mcp:<server>` group with **zero model-visible presence** — the model cannot tell that a configured MCP server exists, so agents keep calling the only visible search tool (`websearch`) even when it returns nothing.

This PR makes deferred MCP servers discoverable through the same mechanism built-in groups (note/session/agenda/…) already use:

- **`MCP.deferredGroupCatalog()`** (`packages/synergy/src/mcp/index.ts`) — cheap snapshot of connected MCP servers and their tool names, read directly from the supervisor's cached `toolDefs` (no `dynamicTool` conversion, no network calls).
- **`ToolExposure.mcpGroupTable()`** (`packages/synergy/src/tool/exposure.ts`) — renders the same three-column markdown table shape as the built-in group table, one row per server with group id, tool count, up to 6 tool names, and an `expand_tools({groups:["mcp:<server>"]})` hint. Bounded at 10 servers / 6 tool names.
- **`expand_tools` description** now appends a "Connected MCP groups" section whenever the defer threshold is active and servers are connected. The description refreshes on every tool init, so main sessions and Cortex subagent sessions both see the directory.

## Verification

- New `packages/synergy/test/tool/mcp-group-catalog.test.ts` — 10 tests covering the pure table formatter (rows, dedup/sort, caps, sanitized ids) and the `expand_tools` wiring (threshold on/off, empty catalog, no-throw init).
- Existing `exposure.test.ts`, `observability.test.ts`, `tool-call-execution.test.ts`, `mcp/supervisor.test.ts` all green unmodified (54 tests).
- `bun run quality:quick` — all 14 gates passed (lint, typecheck, doc:check, decision:check, deadcode, secrets, workflow, …).
- `bun run doc:gen:check` — generated reference docs unchanged (the dynamic section is not a static literal).
- `bun run test:changed` — 5868 pass; the only failures were 8 pre-existing session-migration timeouts under 510-file parallel load, which pass green when run standalone (`migration.test.ts` 20/20).

## Notes

- Discovery-only change: no permission, exposure-mode, `search_tools`, `session.toolState`, config, persisted-state, SDK, or plugin-API semantics changed. `expand_tools` execution still fails closed on permission-denied groups.
- No config keys, no migrations, no new per-turn resident tokens (the section costs tokens only when the model reads the `expand_tools` description).
- Decision record: `docs/decisions/implemented/feature/2026-08-18-mcp-group-catalog-in-expand-tools.md`.


yzxoi (approved):
This closes a real discoverability gap — deferred MCP servers were effectively invisible to the model, so agents kept falling back to the only visible search tool. The directory approach is clean and cheap.

Verified locally:
- The threshold gate matches the defer condition in `discovery.ts`, and the printed `mcp:<server>` ids come from the same `mcpGroupID()` the runtime uses, so they expand correctly.
- Reading the supervisor's cached `toolDefs` avoids both the `ToolDiscovery.collect()` recursion and the dynamicTool conversion cost — good call.
- Catalog failure degrades to an empty section; combined with the registry's `Promise.allSettled`, tool init stays safe.
- New tests pass (10/10) and existing exposure tests stay green (29/29).

Non-blocking nit: `unique(server.toolNames)` is computed up to three times per row in `mcpGroupTable()`; hoisting it to a local would tidy that path.

Thanks for the thorough decision record and for documenting the recursion trap in the alternatives — much appreciated. Nice work!

PR #1198, opened 2026-08-18
fix: host-selected Blueprint reviewers and agent worker SIGTERM misfires

EricSanchezok:
## Summary

Fixes two production BlueprintLoop failure modes:

1. **Host-selected Blueprint reviewers** — `note_write` still exposed an `auditAgent` parameter after the execution-agent parameter was removed, so a Plan/Lattice agent could write an unconfigured or non-host reviewer onto a Blueprint note. The reviewer task then failed at launch and was misclassified as a missing terminal review tool, burning two doomed recovery turns and reporting a misleading `review_terminal_tool_missing`.
2. **Agent worker SIGTERM misfires** — a worker receiving a message that referenced a recently released turn (legal late messages, e.g. `collect-memory` responses) was killed as a protocol violation, taking down the next turn already assigned to it. The crashed turn was not retried because worker exits were not classified as retryable.

## Changes

- `note_write` no longer exposes `auditAgent`; stored metadata is preserved as inert legacy data (same strategy as `defaultAgent`). `note_read` reports `Audit Agent: supervisor`.
- `BlueprintLoopService.resolveBlueprintAuditAgent` always resolves `supervisor`; plugin Protocol 5 `blueprint.start` keeps its explicit validated audit-agent request path.
- Cortex marks tasks that fail before their first model turn with a durable `launchFailure` flag; BlueprintLoop and Light Loop continuation fail such reviewers immediately with `reviewer_launch_failed: <original error>` without consuming the terminal-tool recovery budget.
- Agent worker pool keeps a bounded ring of the two most recently released request IDs (30s TTL): late messages referencing them are dropped with an `agent.worker.late_message` metric instead of killing the worker; never-owned request IDs still kill.
- Protocol-violation kills now record the shared `agent.worker.recycle` metric (`reason: protocol_violation`).
- `SessionRetry.retryable` classifies `Agent worker exited (…)` as retryable so the session retry loop covers worker crashes.

## Verification

- Focused suites: `packages/synergy/test/session/{retry,blueprint-continuation,light-loop-continuation,agent-worker-pool}.test.ts`, `test/tool/note-write.test.ts`, `test/server/blueprint-route.test.ts` — 103 pass / 0 fail, including new regression cases for each fix.
- `bun run quality:quick` — all 14 gates pass (format, lint, typecheck, doc:check, decision:check, deadcode, …).
- `bun run test-layout:check` and `bun run decision:check` pass.
- Full `test:changed` run: 5879 pass / 11 fail; 10 failures were 5s timeout flakes that pass in isolation, and 1 (`ServerProcessLock` pid-reuse case) fails identically on the base commit — pre-existing, unrelated to this change.
- `docs/reference/tools.md`, SDK types, and `openapi.json` regenerated; two implemented decision records added under `docs/decisions/implemented/bug-fix/`.

## Notes

- The previously failing Blueprint notes need no migration: their `auditAgent` metadata is inert and now resolves to `supervisor` automatically.


yzxoi (approved):
Approved — thank you for the sharp insight on both fixes. The released-request grace window keeps the protocol-violation defense for genuinely unowned request IDs while tolerating benign IPC timing, and making worker crashes retryable closes the blast radius on innocent next turns. The `launchFailure` flag is exactly the right boundary: fail a pre-first-turn reviewer immediately with the real error instead of burning the terminal-tool recovery budget. Focused suites pass (103/0) and the decision records are excellent. Great work!

PR #1199, opened 2026-08-18
test: cover generic-branch late-message tolerance for agent workers

EricSanchezok:
## Summary

Follow-up to #1198 addressing review feedback on the late-message tolerance fix. The production SIGTERM incident hit the **generic ownership branch** in `AgentWorkerPool` (`message referenced an unowned turn`), but the original change only had drop-path regression coverage for `heartbeat`-type messages. This PR closes that gap and adds the observability assertions the review requested.

## Changes

- `packages/synergy/test/session/agent-worker-pool.test.ts`
- New test: a non-heartbeat late message (duplicate `complete` for a released turn) arriving while the worker runs its next turn is dropped without killing the worker, records `agent.worker.late_message` with the message type, and does not record a recycle — then the next turn still completes normally.
- New test: a non-heartbeat message referencing a never-owned request still kills the worker and records `agent.worker.recycle` with `reason: protocol_violation`.
- `packages/synergy/src/session/agent-turn/worker-pool.ts`
- `terminateForProtocol` accepts optional message context and includes `messageType`/`requestId` in the protocol-violation warn log, per the planned observability enhancement.

## Verification

- Focused suites (worker-pool, blueprint-continuation, light-loop-continuation, note-write, exposure, blueprint-route): 114 pass / 0 fail.
- `bun run quality:quick`: all 14 gates pass.
- `bun run test:changed`: 4191 pass / 7 fail — all seven are 5s timeout flakes (session migrations and one hook); `test/session/migration.test.ts` passes 20/20 in isolation and the touched surface is limited to the worker pool and its test file.

### DIFF
--- a/packages/synergy/src/session/agent-turn/worker-pool.ts
+++ b/packages/synergy/src/session/agent-turn/worker-pool.ts
@@ -562,7 +562,10 @@ export class AgentWorkerPool {
           this.dropLateMessage(worker, message.type, message.requestId)
           return
         }
-        this.terminateForProtocol(worker, "heartbeat referenced an unowned turn")
+        this.terminateForProtocol(worker, "heartbeat referenced an unowned turn", {
+          messageType: message.type,
+          requestId: message.requestId,
+        })
         return
       }
       this.recordWorkerMemory(
@@ -608,7 +611,10 @@ export class AgentWorkerPool {
         this.dropLateMessage(worker, message.type, lateRequestId)
         return
       }
-      this.terminateForProtocol(worker, "message referenced an unowned turn")
+      this.terminateForProtocol(worker, "message referenced an unowned turn", {
+        messageType: message.type,
+        ...(lateRequestId !== undefined ? { requestId: lateRequestId } : {}),
+      })
       return
     }
     if (message.type === "run-ready") {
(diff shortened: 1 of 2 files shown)


yzxoi (approved):
## Review: Approve ✅

This follow-up closes exactly the gap flagged in #1198's review: the production SIGTERM incident hit the **generic ownership branch** in `AgentWorkerPool`, while the original change only had drop-path coverage for `heartbeat`-type messages.

### What I verified
- Ran the focused suite (`packages/synergy/test/session/agent-worker-pool.test.ts`): **41 pass / 0 fail**, including both new tests.
- Test 1 (duplicate `complete` for a released turn arriving while the worker runs its next turn) genuinely exercises the generic branch: the message fails the `requestId !== worker.task?.requestId` check, matches the recently-released ring, and is dropped — worker survives, `agent.worker.late_message` is recorded once with `{messageType: "complete"}`, no recycle is recorded, and the next turn still completes.
- Test 2 (never-owned request) confirms the kill path still works and now asserts the `agent.worker.recycle` metric with `reason: "protocol_violation"` — the observability assertion the review asked for.
- The `terminateForProtocol` context param is a narrow, well-scoped change: the two call sites pass `messageType`/`requestId` through, and the metric shape stays identical to #1198's definition.

### Nits (non-blocking)
1. `agent.worker.late_message` records only `messageType` in labels — the `requestId` is in the debug log but not the metric. Fine as-is (matches the #1198 metric contract), but worth considering if late-message → turn correlation is ever needed in dashboards.
2. The new `messageType`/`requestId` warn-log fields aren't directly asserted (only metrics are). Acceptable — the metric assertion covers the observable behavior.

No overlapping coverage with the #1198 tests (heartbeat variants, never-owned heartbeat, idle case, post-grace-window case). No new decision record needed; this extends the existing late-message-tolerance change.

Thanks for the quick follow-up — great to see the incident path getting direct regression coverage, and the observability additions will make future protocol violations much easier to diagnose. Nice work!

PR #1200, opened 2026-08-18
fix(channel): deduplicate and filter attachments in outbound projection

yzxoi:
## What

Fix the Feishu channel outbound projection so (a) incidental bash-probe artifacts stop riding along on every message, (b) already-delivered attachments are not re-sent when the same task tree is woken again, and (c) every terminal a foreground turn completes actually reaches the user. All are solved by moving ownership to the right layer instead of patching the projection.

### 1. Deliverability is classified once, at creation

Attachment deliverability is a first-class verdict written when the attachment is created:

- `AttachmentDiscovery.discover` (`packages/synergy/src/tool/attachment-discovery.ts`) records `metadata.attachment.deliverable`: `true` for explicit references (`detectedFrom: "markdown" | "file_url"`), `false` for file paths that merely appeared in tool output (`"line" | "path"`).
- The `attach` tool (`packages/synergy/src/tool/attach.ts`) records `deliverable: true`.
- `MessageV2.isDeliverableAttachment()` (`packages/synergy/src/session/message-v2.ts`) is the single predicate: prefers the canonical verdict, falls back to the legacy `detectedFrom` heuristic for attachments persisted before the verdict existed.

### 2. Delivery dedup is root-scoped, not terminal-scoped

- After a successful channel send, the attachment urls actually delivered are recorded on the task's **root user message** (`metadata.channelOutboundAttachmentUrls` via `markChannelTaskAttachmentsDelivered` in `packages/synergy/src/channel/outbound-parts.ts`).
- `projectChannelTaskParts` skips urls already recorded, so a later terminal reply in the same root tree (steer wake-ups, boss reports, agenda continuations) no longer re-sends attachments that were already delivered.
- A new task = a new root = empty record, so a file can be deliberately re-delivered in a later task. Failed sends record nothing, so retries still work.
- Both delivery points (the async bridge in `outbound.ts` and the foreground `replyChannelTaskAttachments`) share the same projection and recording path.

The previous terminal-scoped `channelOutboundSent` marker could not prevent re-sends: it only gates the bridge trigger for one terminal message, while the projection walks the whole root tree.

### 3. Foreground turns deliver every completed terminal, not just the loop result

The foreground channel path (`channel/index.ts`) snapshots the session's terminal replies (`collectChannelTaskTerminals`) before invoking the inbox loop. The loop drains queued tasks and returns only the LAST terminal, so a terminal the foreground lane's root completed mid-drain would be skipped by the bridge (foreground-registered) yet never delivered by the foreground path — silently dropping its text and attachments when a follow-up message arrived mid-generation.

After the invoke, `deliverForegroundTaskTerminal` delivers such a terminal (text + attachments) if it is neither the loop result nor pre-existing, records it on the root, and marks it `channelOutboundSent`.

## Why

In boss-session flows, a debug `find` probe once matched a Flutter pub-cache test image (a 120x120 grinning-face emoji PNG). `AttachmentDiscovery` attached it (`originTool: "process"`, `detectedFrom: "line"`), and because the outbound bridge projects the task tree's tool attachments on every terminal reply, that emoji image was re-uploaded to Feishu on every subsequent channel message.

A second, separate bug surfaced while testing: when a follow-up channel message arrived while a task was still generating, the foreground invoke drained both tasks but only delivered the final terminal — the first terminal's text and attachment were lost entirely (skipped by the bridge, not delivered by the foreground path).

Filtering in the projection layer alone was rejected as the fix point: it duplicated discovery's classification vocabulary in a downstream consumer and left the model context and UI consuming the same incidental attachments.

## Verification

- `packages/synergy/test/channel/outbound-parts.test.ts`:
  - `filters attachments classified as non-deliverables`
  - `filters legacy incidental attachments via the detectedFrom fallback`
  - `keeps attachments referenced explicitly via markdown and file urls`
  - `deduplicates attachments that share the same asset url`
  - `skips attachments already recorded as delivered on the root message`
- `packages/synergy/test/channel/outbound.test.ts`:
  - `does not re-deliver already-delivered task attachments on a later terminal reply in the same root`
  - `delivers a foreground-completed terminal that the loop did not return` — text + attachment delivered, root marked, terminal marked sent
  - `does not re-deliver a terminal already present before the invoke` — pre-existing and excluded terminals are skipped
  - existing suite (bridge delivery, foreground skip, cards, anchors) still passes
- `packages/synergy/test/tool/attachment-discovery.test.ts` / `attach.test.ts` assert `deliverable` classification.
- `bun test test/channel/outbound-parts.test.ts test/channel/outbound.test.ts test/tool/attachment-discovery.test.ts test/tool/attach.test.ts` — 38 pass, 0 fail.
- `bun run format:check`, `bun script/decision-check.ts`, staged lint — all green. `tsc --noEmit` on packages/synergy: no errors in changed files (pre-existing unrelated errors only).

## Notes

- Narrowing the projection to the terminal task segment was considered and rejected: the existing contract/tests require projecting tool attachments from earlier assistant steps in the same task.
- Content-fingerprint dedup was considered and rejected: content-addressed asset ids already encode the bytes, legacy named ids (`asset://n.png`) are rejected earlier by `resolveAttachmentSource`, and url dedup + root scoping already guarantee uniqueness.
- Decision record: `docs/decisions/implemented/bug-fix/2026-08-18-channel-outbound-attachment-dedup.md`.


yzxoi:
@synergy-agent review

PR #1201, opened 2026-08-18
fix(channel): keep pinned model, persist image paths for look_at fallback

yzxoi:
## Summary

This PR resolves the Feishu channel image-message gap by **aligning channel model selection with chat**, in three parts.

### Part 1: Reject auto-switching to the vision model

**Original problem**: Feishu channel messages with image attachments materialize the image correctly (download → temp file → data URL → `file` model part), but when the pinned model lacks image input capability (e.g. default `deepseek-v4-flash`), `ProviderTransform.unsupportedParts` degrades the image to a text placeholder — the model only sees `[Image]` and a "does not support image input" notice.

**Proposed fix (rejected)**: `Channel.handleMessage` would detect image attachments and switch the delivery to the configured `vision_model` role.

**Why rejected**: channel model selection must align with chat semantics — the explicit session `modelOverride` (and account model) always wins (`input.model ?? session?.modelOverride ?? Agent.getAvailableModel(agent) ?? lastModel` in `createUserMessage`); chat never switches models based on message content. The auto-switch silently overrode explicit selections, breaking that invariant. The decision record is preserved under `docs/decisions/rejected/bug-fix/`.

### Part 2: Persist image attachments locally so the look_at fallback works

Reviewing the rejection surfaced a real gap in the existing fallback: the degradation text tells the model to "use look_at with the file's local path", but channel attachments were materialized as data URLs **without** a `localPath` — the path never appeared in model context, so `look_at(file_path=...)` had nothing to point at. (Chat avoided this because pasted/uploads run `saveDataPartLocally` / `asset://` resolution.)

`ChannelBusyHandoff.buildDurablePromptParts` now persists each image attachment via `Attachment.saveDataPartLocally` at the durable-parts seam (before the inbound temp file is cleaned up) and sets the resulting path on the part, matching chat's media-dir behavior. `MessageV2` already emits `[The user attached a file: ... Local path: ...]` when `includeLocalPath` is set, so the model now sees the exact path for `look_at`. Best-effort: save failures keep the previous data-URL-only part and are logged at warn.

### Part 3: Content-hash media filenames (collision-safe persistence)

Reviewing Part 2 exposed two defects in the shared `Attachment.saveDataPartLocally`:

1. **Same-name overwrite**: two attachments with the same filename but different content wrote to the same path, silently clobbering the first (chat pasted images and channel images both share this function).
2. **Path escape**: filenames come from external metadata (user-controlled), and `path.join(mediaDir, filename)` honored separators — `../../evil.png` escaped the media directory.

The persisted name is now `<basename>-<sha256-12hex><ext>` (or `<sha256-12hex><ext>` without a filename): distinct content with the same name keeps separate files, identical content reuses one path (idempotent under replay/re-send), and separators can no longer escape the media directory.

## Verification

- TDD: each fix started with a failing test (channel `localPath: undefined`, same-name overwrite, path escape) and passes after
- `bun test test/attachment/ test/channel/` → all pass (one pre-existing dev flake: `feishu-acceptance-lane > replaying a materialized direct delivery` times out on dev merge-base too — unrelated to this PR)
- `bun run typecheck`, `bun run lint` (0 warnings), prettier, `decision:check` → all green
- Net source diff vs `dev`: `busy-handoff.ts` seam + `attachment/index.ts` persistence naming

## Files changed

- `packages/synergy/src/channel/busy-handoff.ts` — persist image `localPath` at the durable-parts seam
- `packages/synergy/src/attachment/index.ts` — content-hash media filenames + basename sanitization
- `packages/synergy/test/channel/feishu-busy-inbox.test.ts` — persisted-path coverage
- `packages/synergy/test/attachment/save-data-locally.test.ts` — collision/idempotency/escape coverage
- `docs/decisions/rejected/bug-fix/2026-08-18-channel-image-vision-model.md` — rejected auto-switch
- `docs/decisions/implemented/bug-fix/2026-08-19-channel-image-local-path.md` — path persistence
- `docs/decisions/implemented/bug-fix/2026-08-19-media-filename-content-hash.md` — hash naming


yzxoi:
@synergy-agent review

PR #1202, opened 2026-08-18
perf(ui): memoize session-turn timeline projection per message

yzxoi:
## Summary

Streaming replies kept the desktop renderer main thread saturated even after the row-keying and non-keyed-timeline fixes removed row remounts. Runtime telemetry showed long-task p95 ~600ms and token paint p95 ~584ms while streaming; sampling showed the busy thread executing JIT code with heavy allocation.

The cause: `SessionTurn` re-projected the **entire turn** on every part delta. `projectAssistantMessage` ran for every assistant message in the turn, allocating fresh display-item objects even for settled messages whose parts never changed, so the element-wise `same` guard could never short-circuit the snapshot/boundary/slot-index chain. `markdownText` additionally re-joined the accumulated reply text per delta.

## Changes

- **Per-message projection memos** (`mapArray` over display messages): a part delta re-projects only the owning message; other accessors return cached references, so the downstream `same` guard chain stays dormant between deltas.
- **Explicit `working()` subscription** in `timelineItems`: settlement re-projects every message exactly once (preserves full-mode reasoning promotion/hiding).
- **`latestAssistantTimelineItems`** reuses the per-message accessor instead of projecting the latest assistant twice per tick.
- **`markdownText`** returns early while working — Copy Markdown is only presented after settlement.
- **Balanced-mode live reasoning scan** only builds its map when compact reasoning is active and the turn is working.

## Tests

- New `session-turn-projection-memoization.dom.test.ts` (Vite+Solid+JSDOM fixture): mounts one settled + one streaming assistant, counts tool-renderer lookups on the settled message, asserts streaming deltas do not re-project it and settlement re-projects once and reveals Copy Markdown.
- Regression suites run: `session-turn-timeline`, `session-turn-activity`, `session-turn-projection` (99 pass), plus `session-turn-activity-switch`, `compact-reasoning-settlement`, `session-turn-timeline-boundary` DOM suites (5 pass).
- `bun run typecheck` (tsgo) clean; `bun run decision:check` clean; Prettier clean.

## Decision record

- `docs/decisions/implemented/bug-fix/2026-08-19-session-turn-projection-memoization.md`


yzxoi:
@synergy-agent review

PR #1205, opened 2026-08-19
feat(settings): apply theme and color scheme instantly without Save Changes

TTAWDTT:
## Summary

- Theme and color scheme selections now apply **immediately on click** instead of requiring the user to press "Save Changes"
- Color scheme is purely client-side (localStorage) — no server call needed
- Theme ID is persisted to the server via an independent background domain update, with error toast on failure
- "Save Changes" button and dirty-state tracking no longer include theme/colorScheme
- Discarding other changes restores the instant-applied theme from the live theme provider to prevent stale server config from overwriting it

## Test plan

- [x] Open Settings → General, switch theme → applies instantly
- [x] Switch color scheme (light/dark) → applies instantly
- [x] Modify other settings then Discard → theme/colorScheme not reverted
- [x] Refresh page after theme change → theme persists (server-side)

### DIFF
--- a/packages/app/src/components/settings/SettingsPanel.tsx
+++ b/packages/app/src/components/settings/SettingsPanel.tsx
@@ -61,7 +61,6 @@ import {
   retainDraftAfterSave,
   saveExplicitSettingsChanges,
   snapshotSettingsDraft,
-  themeIdToApplyAfterSave,
 } from "./settings-explicit-save"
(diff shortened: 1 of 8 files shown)


yzxoi:
LGTM — theme and color scheme now apply instantly while the explicit save flow stays intact, and the follow-up fix keeps the theme picker in sync with the applied appearance after Save Changes. Thanks for the contribution @TTAWDTT!

yzxoi (approved):

PR #1207, opened 2026-08-19
fix(session): env block probes live git state via GitHealth

EricSanchezok:
## Summary

The `<env>` system prompt block rendered `Is directory a git repo: no` for the entire lifetime of any session whose scope was created **before** the directory became a git repository.

- Root cause: `SystemPrompt.environment()` derived the line from `scope.vcs`, a snapshot field resolved once when the scope was created and frozen into the turn envelope. A session that starts in a non-git directory and later runs `git init` showed `no` forever, while the `<git-health>` block (a live probe) and every real git command said the directory is a repo.
- Fix: `GitHealth.isGitRepo(cwd)` added to `project/git-health.ts`, wrapping the existing `resolveRepo()` probe (same `git rev-parse` semantics, same 2s timeout). `system.ts` now computes the env line from that live probe for project scopes; home scope keeps rendering `no`.
- The env block and git-health now share one probe, so they structurally cannot disagree. Sessions created before `git init` self-heal on their next turn.

## Changes

- `packages/synergy/src/project/git-health.ts` — export `isGitRepo(cwd?)` (bypasses the scan cache deliberately; env needs the live answer).
- `packages/synergy/src/session/system.ts` — env block uses `GitHealth.isGitRepo(ScopeContext.current.directory)` instead of the `scope.vcs` snapshot.
- Tests: `git-health.test.ts` (4 new `isGitRepo` cases) + `prompt-roots.test.ts` (2 new regression cases: same stale scope snapshot, `no` before `git init`, `yes` after).
- Decision record: `docs/decisions/implemented/bug-fix/2026-08-19-env-block-live-git-detection.md`.

## Verification

- Red-first: new regression case fails on old code (`git repo: no` after `git init`).
- `bun test test/project/git-health.test.ts test/session/prompt-roots.test.ts` → 42 pass / 0 fail.
- `bun run quality:quick` → all 14 gates pass (format, lint, typecheck, doc, decision, deadcode, secrets, workflow, etc.).
- `bun run decision:check` → passed.


yzxoi (approved):
## Review: Approve ✅

Reviewed the full change set (original commit `677f5998df` + follow-up `81cc848174`).

### What was verified

- **Root cause fix at the right layer**: `SystemPrompt.environment()` no longer trusts the frozen `scope.vcs` snapshot; it probes live via the new `GitHealth.isGitRepo()`, sharing the same `resolveRepo()` probe (same `git rev-parse --is-inside-work-tree` semantics, same 2s timeout) as the `<git-health>` block — structurally eliminating the two-sources-of-truth contradiction.
- **Deliberate cache bypass is correct**: the env block needs the current answer; a cached probe would re-create the exact staleness bug this fixes.
- **Tests**: `bun test test/project/git-health.test.ts test/session/prompt-roots.test.ts` → 42 pass / 0 fail (locally re-run on this worktree).
- **Gates**: `typecheck` (tsgo), `decision:check`, oxlint, prettier all pass on the changed files.
- **Decision record**: format-compliant, alternatives genuinely considered (snapshot refresh, git-health block inference, `.git` existsSync, deleting the line, second Vcs implementation all reasonably rejected).

### Follow-up fixes included in `81cc848174` (from review)

1. Restored `refresh()` in the `git-health.test.ts` contract comment (it was accidentally dropped when `isGitRepo` was added; `refresh` still exists and is used by `check()`/`injectCached()`).
2. Corrected the decision record's spawn count: a repo probe is **three** spawns (one `rev-parse --is-inside-work-tree`, then two parallel for `--show-toplevel`/`--git-dir`), not two.
3. Removed the never-rendered `<files>` block (`scope.vcs === "git" && false` dead code) and its now-unused `Ripgrep` import from `system.ts` — pure dead-code cleanup, no behavior change.

### Notes (non-blocking, pre-existing)

- `system.ts` had `&& false` since the initial open-source release; the cleanup above was opportunistic and safe.
- Per-turn cost of up to 3 uncached git spawns for project scopes is acceptable; a turn-level cache could be considered later if it ever matters.

No blockers. Thanks for the clean, well-tested fix!

PR #1208, opened 2026-08-19
feat(app): show active project name in session top bar

EricSanchezok:
## Summary

- The desktop session top bar now shows the **active project name** in place of the bare folder icon, so users can tell which project a session belongs to without consulting the sidebar.
- The label reuses the existing `getScopeLabel` rules (custom scope name → directory basename → "Project"), long names ellipsize, and the full path is exposed through a tooltip. Home sessions keep no prefix; the mobile layout is unchanged.
- Adds `resolveProjectScope` in `packages/app/src/utils/scope.ts` (pure, unit-tested): prefers the active scope resolved by the sync store, falls back to a known scope by worktree or sandbox directory so sub-directory routes name their parent project.
- Includes a small process fix: `doc-check --staged` now checks the same Markdown scope as the full `doc:check` gate, so out-of-scope product docs (which keep one principle per line) no longer fail pre-commit on pre-existing lines.

## Verification

- `bun test test/utils/scope.test.ts test/workbench-surface.test.ts` (app) — 26 pass
- `bun run --cwd packages/app test` — full app suite pass (incl. build CSS contract)
- `bun run --cwd packages/app typecheck` — pass
- `bun run localization:check` — pass
- `bun test --config /dev/null test/script/doc-check.test.ts` — 10 pass
- `bun run quality:quick` — all 14 gates pass
- `bun run doc:check` / `bun run decision:check` — pass

## Files

- `packages/app/src/components/top-bar/session-top-bar.tsx` / `.css`
- `packages/app/src/utils/scope.ts` + `packages/app/test/utils/scope.test.ts`
- `packages/app/test/workbench-surface.test.ts`
- `packages/app/PRODUCT.md`
- `script/doc-check.ts` + `test/script/doc-check.test.ts`
- `docs/decisions/implemented/feature/2026-08-19-session-topbar-project-name.md`
- `docs/decisions/implemented/process/2026-08-19-doc-check-staged-scope.md`


yzxoi:
## Review findings

Thanks for the PR — the direction is right (project identity in session chrome, plus aligning staged `doc-check` with the full gate). I traced the complete resolution path (bootstrap → `Scope.fromDirectory` → `sync.scope` → `resolveProjectScope`) and found one functional bug plus a few smaller issues.

### 1. (Bug, medium) Sandbox sub-directories resolve to their own scope, never the parent project — the `sandboxes` fallback is effectively dead code

The PR summary says *"sub-directory routes name their parent project"*, but the actual path is:

- Server-side, `ScopeContext.current.scope` comes from `Scope.fromDirectory(directory)` (`server.ts`), which never maps a registered sandbox back to its parent: a non-git sub-directory yields `{ id: dirHash(dir), worktree: dir, sandbox: dir }` — no upward traversal by design (see the scope-boundary comment in `packages/synergy/src/scope/index.ts`).
- The frontend `sync.scope` getter returns exactly that bootstrap scope (binary search by `store.scopeID` in `context/sync.tsx`), so `resolveProjectScope`'s first branch `if (activeScope?.worktree) return activeScope` always wins → the label becomes the sandbox basename (e.g. `app`), not the parent project name.
- The `scopes.find(s => s.sandboxes?.includes(directory))` branch only runs in the pre-bootstrap window (`scopeID === ""`) or when the scopeID is absent from the global list — i.e. never in normal operation.

This also diverges from the existing frontend mapping (`context/layout` `roots()` and `mobile-drawer.tsx`), which both match sandbox → parent **before** trusting the resolved scope.

Suggested fix — check the sandbox reverse-map first, then only trust the active scope when it actually covers the route directory:

```ts
if (!directory || isHomeScope(directory)) return undefined
const bySandbox = scopes.find((scope) => scope.sandboxes?.includes(directory))
if (bySandbox) return bySandbox
if (activeScope && (activeScope.worktree === directory || activeScope.sandboxes?.includes(directory))) return activeScope
return scopes.find((scope) => scope.worktree === directory)
```

### 2. (Minor) Empty/absent `params.dir` renders a fake "Project" label

`isGlobal()` returns `false` when `params.dir` is falsy, so `<Show when={!isGlobal()}>` renders, and `getScopeLabel(undefined, "")` falls back to `"Project"`. The route always passes `:dir` today, but the guard should be symmetric (`true` when `params.dir` is missing).

### 3. (Nit) Tooltip shows the worktree, not the directory the user actually opened

`projectPath = projectScope()?.worktree || directory()` — for a sandbox session the tooltip points at the parent root while the user is inside the sub-directory. Consider showing `directory()` instead.

### 4. doc-check

- The scope alignment is the right fix, but out-of-scope staged files now also lose their link checks (staged mode previously checked them). That's a deliberate behavior change worth stating explicitly — the decision record currently reads ambiguously ("still receiving link checks").
- **Subdirectory execution silently empties the check**: `stagedFiles()` runs `git diff --cached --name-only`, and git always prints paths relative to the **repository root**, but `filterStagedFiles` resolves them with `path.resolve(cwd, file)` where `cwd` may be a subdirectory. Running `doc-check --staged` from inside `packages/` therefore filters every staged file out and the gate passes without checking anything. Fix: resolve against the repo root (or pass `root` as the resolution base), and add a warning when staged markdown exists but the filtered set is empty.

### 5. (Nit) Test style

The new `workbench-surface.test.ts` case asserts source text (`toContain('resolveProjectScope(directory()')`, `not.toContain('getSemanticIcon("workspace.main")')`) rather than behavior. I understand it matches the file's existing surface-polarity style, but the sandbox behavior in #1 is exactly the kind of thing only a behavioral test catches — the pure `scope.test.ts` unit tests are good; consider a component-level case for the sandbox route.

yzxoi:
## Fixes pushed

The issues above are addressed in two commits on this branch:

- **`fix(app): resolve parent project for sandbox session routes`** — `resolveProjectScope` now matches a known scope by sandbox directory first (so a sandbox session names its parent project even when the bootstrap-resolved active scope covers the sub-directory), only trusts the active scope when its worktree/sandboxes actually cover the route directory, treats a missing `params.dir` as global (no fake "Project" label), and the tooltip now shows the actual route directory instead of the parent worktree.
- **`fix(script): resolve staged doc-check paths from repo root`** — `filterStagedFiles` resolves staged paths against the repository root (git prints them root-relative even when the hook runs from a subdirectory), and a staged run with markdown that is entirely outside the document scope now reports a warning instead of passing silently.

Verification:

- `bun test test/utils/scope.test.ts` (app) — 12 pass, including the new sandbox-over-active-scope cases
- `bun test test/workbench-surface.test.ts` (app) — 16 pass
- `bun run --cwd packages/app typecheck` — pass
- `bun test --config /dev/null test/script/doc-check.test.ts` — 12 pass
- `bun run doc:check` / `bun run decision:check` — pass
- `bun run quality:quick` — 13/14 gates pass; the only failure is `secrets:check` on this machine because `gitleaks` is not installed (CI runs the secret scan)

I left the source-text assertions in `workbench-surface.test.ts` untouched — they match the file's existing surface-polarity style, and the behavioral coverage for the sandbox case now lives in the `scope.test.ts` unit tests.

yzxoi:
@synergy-agent review

yzxoi:
@synergy-agent review

yzxoi:
## Follow-up: keep the folder icon beside the project name

Based on visual feedback, the bare project name read as an unanchored model selector, so the folder icon is back: `fix(app): keep folder icon beside project name in session top bar` (`23ed043b7e`) restores `workspace.main` next to the name, keeping the established project-session anchor while retaining the new identity, tooltip, and sandbox resolution. The workbench surface test and decision record were updated to match (icon + name is now the accepted layout).

yzxoi (approved):
## Approved — thanks!

Approving this PR. The session top bar now names the active project while keeping the folder icon and separator rhythm, which makes project identity readable at a glance without losing the established visual anchor.

The review feedback was addressed cleanly:

- Sandbox sub-directory sessions resolve to their **parent project** (sandbox reverse-map checked before the bootstrap-resolved active scope), matching the existing sidebar/mobile-drawer behavior.
- `doc-check --staged` now resolves staged paths against the repo root and warns instead of silently passing when nothing is in scope.
- Guards and tooltip behavior are symmetric and show the actual route directory.

Unit coverage for the resolution logic and the doc-check helper is in place, and the decision records were kept in sync.

Nice work — thanks for the contribution, and for incorporating the review fixes quickly!

PR #1209, opened 2026-08-19
feat(skills): per-source compatibility toggles and honest load diagnostics

EricSanchezok:
## Summary

- **Settings → Skills compatibility toggles**: new Skills section under Library with per-source on/off controls for Agent Skills (`.agents/skills`), Claude Code (`.claude/skills`), Codex (`.codex/skills`), and OpenClaw (`.openclaw/skills` + workspace `skills/`). Each row shows that source's discovered skill count (from `skill.list`). All default **on** (missing config key = enabled, zero behavior change for existing users). Disabling a source hides its skills from the library, skill tool, and session catalog — files on disk are never touched.
- **Live apply, no restart (R7)**: toggles take effect immediately after save through the existing config-reload pipeline (`skills → skill` cascade + `CONFIG_LIVE_APPLIED` entry). Hand-editing `55-skills.jsonc` applies live too via the watcher. Locked in by a reload test.
- **Honest load diagnostics**: the Library → Skills banner no longer says "N skill(s) skipped during load". It now shows grouped, collapsed-by-default diagnostics — failed / shadowed / compatibility notices — with correct `en` and `zh-CN` pluralization. Warnings on loaded skills are never counted as "skipped".

## What changed

- New `skills` config domain (`55-skills.jsonc`) + `SkillsCompatibility`/`SkillsConfig` schema (`config/domain.ts`, `config/schema.ts`) — `assertRegistryComplete` pair, importable/exportable.
- `Skill.state` reads config at init, records per-source candidate counts, and excludes disabled sources from materialization/grouping/winner competition; env flag `SYNERGY_DISABLE_CLAUDE_CODE_SKILLS` stays a hard AND override. `Skill.sourceCounts()` added; `skill.list` now returns `sources` with counts (route + SDK/OpenAPI regenerated).
- Settings frontend: `SkillsPanel` (SettingRow + Switch, existing tokens), catalog/types/hydration/patch wiring, count resource + refetch on `skills` config change.
- Library frontend: `partitionSkillDiagnostics` (failed / shadowed / compat, deduped by path+code) + grouped banner.
- Docs: implemented decision record, regenerated config reference, `synergy-config` reference + `base.txt` notes.

## Verification

- Backend: `test/skill` + config domain + reload + skill-route suites (122 tests green), including `compatibility.test.ts` (defaults, per-source disable, winner semantics, claude flag AND, live reload) and the R7 reload-cascade guard.
- Frontend: settings hydration/patch tests, library diagnostics partition tests, full app suite green; `i18n:check` + `localization:check` green (en/zh-CN authored, no `{plural}` hack).
- Gates: `quality:quick` 14/14, `doc:check`, `decision:check`, `./script/generate.ts` fresh.


yzxoi:
@synergy-agent review

yzxoi (approved):
## Review: Approved ✅

Reviewed the full diff against `dev`, plus the three follow-up fix commits from review. The feature is well-scoped and well-tested; isolated-runtime verification confirmed the fixes end to end.

### Fixed during review
1. **Parallel watcher reload race** (`a2301fc136`) — `reloadTargets: ["config", "skill"]` executed both targets in parallel, so `Skill.state` init could read a stale `Config.current()` mid-reload and resurrect hidden sources. Now `["config"]` only; the serial `config → skill` cascade rebuilds skill state on both the settings-save and watcher paths.
2. **Project-level overrides baked into global config** (`a2301fc136`) — the settings patch wrote all four compatibility booleans whenever one diverged, so merged project-level overrides got re-written into the global `55-skills.jsonc`. `buildSkillsPatch` now emits only the fields the user actually changed.
3. **Strict sources rejected standard vendor fields** (`4e0e3f7650`) — `argument-hint`, `model`, `effort`, `context`, `agent`, `hooks` are standard Agent Skills / Claude Code fields; the strict schema treated them as manifest errors and refused to load the skill entirely. Declared as ignored `unknown` fields — ecosystem-authored skills now load with zero diagnostics.
4. **Nested `references/` entries scanned as skills** (`4e0e3f7650`) — nested `SKILL.md` under a skill's `references/` directory surfaced spurious load errors (e.g. `references/numen-official/*`). Excluded from candidate enumeration.

### Verification
- Isolated runtime: loaded skills 40 → 102; `Unrecognized key: "argument-hint"` errors 63 → 0. Remaining diagnostics are all legitimate (missing frontmatter, directory-name mismatch).
- Tests: 101 pass / 0 fail across the skill suite (3 new regression tests, red → green), frontend view-model tests pass, Prettier / staged-lint / decision:check / doc:check clean.
- Decision record updated to document both fixes.

### Nits (non-blocking, fine to follow up)
- `partitionSkillDiagnostics` labels any non-error/non-shadowed diagnostic as "Compatibility notices", slightly narrow for `reference_read_failed` / shim diagnostics.
- Panel counts are pre-dedup candidate entry counts; the copy could clarify "candidates" to avoid implying unique skills.

Nice work — clear design, honest diagnostics, solid test coverage.

PR #1211, opened 2026-08-19
fix(ui): route session store reads through null-safe session data view

EricSanchezok:
## Summary

根治 session 切换时的 undefined 崩溃类别（`Cannot read properties of undefined (reading 'find'/'filter'/...)`）。此类崩溃自 4 月以来已反复出现十余次（`b0fa7fcf0`、`6cdd685be`、`170cfae13`、`014be8f7b`、`57f8da845`、`6968eb569` 等），根因是渲染链直接读取共享 scope store 的会话字段，而 session 切换瞬间 store 处于中间态（新 session 桶未加载、旧 scope store 已释放、`createMemo` 默认值首次计算后失效），每次都在不同调用点爆出。

本次改为架构级根治：

- 新增 `createSessionDataView()` 空安全视图层（`packages/ui/src/context/session-data-view.ts`）：所有访问器在函数体内做 `?? EMPTY` 兜底，缺失桶返回共享空数组常量（保住 `same()` 相等守卫，无性能回退）。
- `useData()` 新增 `view` getter（现有形状不变，13 个消费方零破坏）；app 侧新增 `useSessionDataView()` hook。
- 迁移全部会话字段读取：app 20 文件、ui 6 文件（含 `session-turn.tsx` 12 处、permission-dock、status-bar、prompt-input 等）。
- 恢复投影链最后防线：`displayItemProjections()[index]?.() ?? []`（`6968eb569` 删除的兜底）、`turnMessagesFor(undefined)` 返回 `[]`。
- 保留依赖 undefined 判空的 loading 语义（handoff/inbox/session_diff 门槛），不做行为改变。

## Verification

- 新增测试：
- `session-data-view.test.ts` — 空安全矩阵 + 共享引用 `===` 断言（6 用例）
- `session-switch-stress.dom.test.ts` — 快速交替桶替换/清空/整体 store 恢复，断言零渲染错误（4 用例）
- `turnMessagesFor(undefined)` 投影用例
- 回归：ui 全量测试、app 全量测试、两包 typecheck 全绿；`bun run quality:quick` 14 项 gate 全部通过。
- 已合并最新 dev（19 commits），合并后重新跑 ui/app 全量测试与 typecheck 全绿。

## Notes

- 新增决策记录 `docs/decisions/implemented/architecture/2026-08-19-session-data-view-layer.md`
- `develop-frontend` skill 增加强制约定：渲染层必须经视图层读取会话字段

PR #1212, opened 2026-08-19
fix(ui): snapshot timeline guard values to stop k(...) is not a function

EricSanchezok:
## Summary

Eliminates the `k(...) is not a function` crash class that recurs during session switches, complementing #1211 (session data view layer).

- **TimelineDisplayInner** now resolves its item once through `parseTimelineDisplayItem` into a discriminated kind snapshot. Each `<Match>` is keyed on the **stable kind string**, so streaming ticks (which replace item references via reconcile) keep the subtree mounted — preserving the non-keyed streaming behavior restored in `57f8da845` — while a genuine kind change (reasoning promotion/collapse, activity mode switch) remounts correctly.
- Branch children short-circuit on `undefined` and never re-evaluate a destroyed accessor: a disposed owner degrades to no-match instead of Solid's `untrack(switchFunc)()` calling a dead memo and throwing `k(...) is not a function`.
- **SessionTurn's root render is wrapped in an ErrorBoundary** (`data-slot="session-turn-error"`, localized `session-turn.render-error`), so previously uncovered guards (`shellModePart`, `specialUserMessageRenderer`, the `item()`/`boundary()` Show chain) degrade to a localized turn-level error card. The existing TimelineDisplay item-level boundary stays as the inner defense layer.
- Records the decision (`2026-08-19-timeline-guard-value-snapshot.md`) and adds the guard-value invariant to the `develop-frontend` skill.

## Verification

- New `session-turn-dispose-guard.dom.test.ts` (3 cases): streaming identity preserved with content updates, kind flips switch branches without uncaught errors, disposing the owner then ticking surfaces no window error.
- Existing boundary/activity/compact-reasoning/projection DOM tests stay green (identity assertion included).
- `bun run --cwd packages/ui test` 542 pass; `bun run --cwd packages/app test` 1581 pass; both packages typecheck; `bun run quality:quick` all 14 gates pass (incl. `decision:check`, `doc:check`, `localization:check` with extracted catalogs).

## Relation to #1211

#1211 fixes the store-intermediate-state `reading 'find'/'filter'` crash class via a null-safe session data view layer. This PR fixes the **disposed-owner guard re-evaluation** class, which is data-independent and not reachable by the view layer. The two are complementary: together they close the session-switch crash family.


yzxoi:
Thanks for the detailed write-up — I read the decision record and the three DOM tests against the solid-js 1.9.10 runtime. Two things need to change before this can land:

**1. The turn-level boundary should log + propagate, not degrade to a card**

Swallowing the error into an error card makes this crash family invisible: `ErrorBoundary`/`catchError` intercepts the throw, so nothing reaches `window.onerror` or telemetry. The card also always shows the raw minified message — `err?.message || _(SESSION_TURN_DESC.renderError)` evaluates `renderError` only for errors without a message, and this class is `TypeError`s with messages, so the localized copy is dead on the main path.

The correct behavior is the one the Part boundary (and skill rule 7) already establishes: log with stable IDs only — no message content, tool payloads, or paths — then let the error propagate to the app-level boundary and render `FatalErrorPage`, which already has copyable error details, an issue link, and a recovery action. A renderer crash that survives SessionTurn should be fatal-visible, not silently replaced by a card.

solid-js 1.9.10 notes: `ErrorBoundary` has no `onError` hook, but throwing from the fallback escapes the boundary (the fallback isn't evaluated inside its own `catchError`) and reaches the parent boundary; equivalently, use a bare `catchError` whose handler logs and rethrows.

With log+propagate, `session-turn.render-error` becomes dead and should be removed, and the decision record + skill rule 8 should document log-then-propagate instead of degrade-to-card.

**2. The crash still has no verified reproduction**

The decision record explains the mechanism but has no captured stack, and "the disposed memo accessor returns undefined" is inferred rather than evidenced. In solid 1.9.10 a normally disposed memo keeps its cached value and `cleanNode` detaches it from its sources' observers, so a post-dispose tick notifies nothing — the dispose+tick test cannot exercise the `untrack(switchFunc)()` path, and by my reading of the runtime all three new cases would also pass against current dev code.

Please attach the field stack (minified + sourcemapped), or add a case that actually fails on current code — e.g. an update enqueued and the owner disposed before flush, or a live observer reading a destroyed guard accessor. Without a failing case we can't tell whether this PR fixes the reported crash or rearranges the same mechanism.

The kind-snapshot core (stable kind key + single parser memo) looks right for keeping streaming remount-free while keying the branches — the concerns are the boundary policy and the missing repro evidence.

yzxoi:
CI 全绿但 mergeable=CONFLICTING + DIRTY。dev 近期已合入 #1225/#1226/#1230 等改动，分支需要同步到最新 dev 解决冲突后推送，CI 会重新跑。timeline guard 修复本身与近期 CI 红无关，冲突解决后即可合并。

PR #1216, opened 2026-08-20
feat(provider): live Grok model discovery via xAI language-models API

EricSanchezok:
## Summary

- **Live Grok model discovery**: register `fetchModelCatalog` for the `grok` provider profile so the model list is discovered from the xAI `GET /v1/language-models` API using the stored subscription OAuth credential. New models (e.g. `grok-4.6`) now appear automatically within the catalog refresh TTL (≤1h) or via `synergy models --refresh`, with no code change required.
- **Robust fallback**: non-2xx responses, malformed envelopes, or missing credentials return an empty catalog, so the existing `ProviderCatalog` failure path keeps the bundled list and retries — credentials are never marked dead on 403 (entitlement gate, not credential failure).
- **Bundled fallback updated**: `DEFAULT_MODEL_IDS` now `["grok-4.6", "grok-4.5", "grok-4.3", "grok-build-0.1"]` and the Grok recommended default model is `grok-4.6`, so offline / first-run / unauthenticated states still surface the latest flagship.
- Reuses the existing catalog infrastructure unchanged (snapshot cache, 1h TTL refresh, RuntimeReload hot reload, retained-model semantics). No config/schema/route/SDK/persistence changes.

## Files changed

- `packages/synergy/src/provider/grok.ts` — `fetchModelCatalog` + updated `DEFAULT_MODEL_IDS`
- `packages/synergy/src/provider/builtin.ts` — register `fetchModelCatalog`, default model `grok-4.6`
- `packages/synergy/test/provider/grok.test.ts` — unit (parsing/defensive), integration (live vs bundled cache isolation), failure fallback tests
- `docs/decisions/implemented/feature/2026-08-20-grok-live-model-discovery.md` — decision record
- `docs/reference/cli-guide.md`, `docs/reference/configuration-layout.md` — one-line live-discovery note

## Verification

```bash
cd packages/synergy
bun test test/provider/grok.test.ts            # 14 pass (4 new)
bun test test/provider/                        # 370 pass / 0 fail
bun run typecheck                              # clean
bun run lint                                   # 0 warnings / 0 errors
bun run format:check                           # clean
bun run decision:check                         # Decision validation passed
bun run doc:check                              # Document validation passed
```


yzxoi:
## Review: feat(provider): live Grok model discovery via xAI language-models API

Thanks for the change — the direction is right and the failure handling is solid (403 → empty catalog without marking credentials dead matches xAI's tier-gating reality). I verified the implementation against the current codebase and the xAI docs; a few issues need fixing before merge:

### 1. `context_length` is not part of the `/v1/language-models` schema (high)

Per the official xAI REST reference (`docs.x.ai/developers/rest-api-reference/inference/models`), `GET /v1/language-models` documents no `context_length` field — the only length-related field is `long_context_threshold` (a pricing boundary). `context_length` exists only on the OpenAI-compatible `GET /v1/models`, and **no endpoint exposes an output-token limit** (grok-4.6's model page states "No text output limit").

Consequences in the current code:

- The `entry.context_length` read is effectively dead against the real API, so the "live limits" claim in the decision record doesn't hold.
- The hardcoded `output: 32_768` is ungrounded; if xAI ever returns a context_length, this would **overwrite** the correct models.dev limit via `mergeRefresh` → `applySnapshotEntries` (the `patch` deep-merges over source metadata).
- The unit test mocks `context_length: 524_288`, asserting behavior the real API never produces.

Fix: drop the limit construction entirely and let limit metadata inherit from models.dev / the bundled fallback (the decision record already concedes "the context_length gap is covered by models.dev/fallback metadata" — make the code match). Assert in tests that entries carry only `{ id }` (+ `inputImage`) regardless of any `context_length` present in the payload.

### 2. grok-build-0.1 silently disappears for authenticated users (high)

`applyCachedDiscovery` rebuilds `provider.models` entirely from the snapshot (`activeModels + retainedModels`) once a live catalog succeeds. `/v1/language-models` returns only language models; if grok-build-0.1 (a coding-agent surface) is not in that list, it vanishes from an authenticated account's catalog on the **first** successful refresh (no previous snapshot → empty retained set), while offline/unauthenticated users still see it via the bundled fallback. Online/offline inconsistency.

Fix: either verify whether the endpoint actually returns grok-build, or keep bundled `fallbackModels` entries that the live response did not return instead of dropping them.

### 3. Catalog identity hashes the refresh token, which xAI rotates on every refresh (medium)

The grok profile does not register `modelCatalogIdentity`, so snapshots use the default credential identity — which includes the refresh token (`defaultCredentialIdentity` in `catalog.ts`). Multiple independent clients confirm xAI **rotates and revokes the refresh token on every refresh** (e.g. aeon PR #773 live-verified; holon #2153; pi-xai-oauth docs). A rotation therefore changes the identity hash → snapshot key mismatch → live state rebuilds and retained-model semantics reset. Codex solved the same problem with a JWT-claim-based `modelCatalogIdentity`.

Fix: register `modelCatalogIdentity` deriving a stable identity from the access-token JWT `sub` (fall back to `email`, then to `undefined` so the default path is unchanged when no claim exists).

### 4. Timeouts/network errors are swallowed into `invalid_response` (low)

`fetchModelCatalog` `.catch(() => undefined)` turns `AbortSignal.timeout` (TimeoutError) and `TypeError` into `[]`, so `classifyFailure` can only report `invalid_response` instead of `timeout`/`network`. Behavior is similar (1-min retry), but diagnostics and the catalog-state UI are misleading. Mirror codex's `fetchModelPayload`: wrap the request in `ProviderAuthRecovery.execute` (with `reloadOnTransition: false`, `throwOnActionRequired: false`) and let non-2xx map to `[]` while transport errors propagate.

### 5. Test issues

- The integration test mocks a `context_length` field the real API never returns (see #1).
- The ProviderCatalog integration tests write snapshots for the real provider ID and only call `ProviderCatalog.reset()` (memory-only); the on-disk snapshot file is not isolated between tests, so snapshots leak across tests in the same file. Delete `Global.Path.providerModelCatalogCache` in the test reset (catalog-stability tests already do this).

### 6. Docs

`configuration-layout.md` gains a Chinese sentence in an otherwise English document (the same content in `cli-guide.md` is English). Keep the reference docs in one language.

---

Verified as correct: `payload.models` envelope (matches the official schema, not `data`), `input_modalities.includes("image")` for vision, 403 → empty catalog without marking credentials dead, 10s timeout, and the reuse of the existing catalog failure path (`neverVerified` guard keeps the bundled list on first-failure).

I've implemented these fixes locally (branch `synergy/session-fe3137` on the fork, pushed for reference): dropped the limit construction, removed grok-build-0.1 from `DEFAULT_MODEL_IDS`, added a JWT-sub `modelCatalogIdentity` + rotation-stability test, switched the fetch to `ProviderAuthRecovery.execute` with timeout/network classification tests, isolated the snapshot cache file in tests, and made the docs English.

yzxoi:
@synergy-agent review

yzxoi:
@synergy-agent review

yzxoi (approved):

PR #1218, opened 2026-08-20
fix(embedding): auto-fallback to hf-mirror and add download timeout

TTAWDTT:
## Summary

- Add 30s TTFB timeout for embedding model downloads to prevent indefinite hangs when huggingface.co is unreachable (e.g. behind a network that requires a proxy). Without this, the download stays at 0% with no error message.
- Auto-fallback from huggingface.co to hf-mirror.com when the primary download source fails, then to local disk cache as a last resort. This keeps the zero-config default working in networks where huggingface.co is blocked.
- Update existing tests to cover the new 3-tier fallback chain (huggingface → hf-mirror → disk cache).

## Behavior

| Scenario                                         | Behavior                                            |
| ------------------------------------------------ | --------------------------------------------------- |
| huggingface.co reachable                         | Normal download                                     |
| huggingface.co unreachable                       | Auto-switch to hf-mirror.com → success if reachable |
| Both huggingface.co + hf-mirror unreachable      | Try local disk cache → run offline if cached        |
| All sources fail                                 | Report `asset: "failed"` with clear error message   |
| User explicitly configures `source: "hf-mirror"` | No auto-fallback, direct to disk cache on failure   |

## Changes

**`embedding-runtime.ts`:**

- Added `fetchRemoteWithTimeout()` — wraps remote fetch with 30s TTFB timeout, produces a clear error message pointing to proxy/mirror config
- Only affects `runtime.env.fetch` (transformers.js downloads), not `globalThis.fetch` (LLM provider calls unaffected)

**`embedding.ts` — `loadLocalExtractor()`:**

- When `source === "huggingface"` and pipeline fails: switch to `hf-mirror`, call `runtime.configure()`, retry pipeline
- If hf-mirror also fails: fall back to disk cache (`local_files_only: true`)
- Error state preserves the original configured source so `status()` correctly reports the failure

**`embedding-local.test.ts`:**

- Updated "falls back to the disk cache after a network load failure" test: expects 3 pipeline calls (huggingface → hf-mirror → disk cache)
- Updated "allows a failed explicit download to retry in the same process" test: expects 4 pipeline calls (3 fail + 1 retry succeeds)

## Test Results

All 21 embedding tests pass:

- `embedding-local.test.ts`: 14/14 ✅
- `embedding-standalone.test.ts`: 4/4 ✅
- `library-embedding-api.test.ts`: 3/3 ✅


yzxoi:
@synergy-agent review

yzxoi (approved):

Issue #1217, opened 2026-08-20
Add the ability to trigger scheduled tasks with a hook


TTAWDTT:
### Problem or motivation

Synergy has been equipped with hooks to record the end/start of a user request in a session, but they are currently underutilized. 

Given that sessions/agents in synergy are able to perceive and affect other sessions and agents, I highly recommend adding the ability to trigger scheduled tasks with a hook, so that agents(especially boss agent) can make scheduled tasks not only based on time (like "Doing this 30 min later" or "Doing that every 8 a.m."), but also based on session/agent status (like "When session 'research' ends this turn, tell me what he has done and what his conclusion is.").

### Proposed solution

Just use the hooks that has been equipped ...
I have not figure out the details ...

yzxoi:
Implemented in #1219

PR #1219, opened 2026-08-20
feat(agenda): add session turn-event triggers

yzxoi:
## Summary

Add a fourth agenda trigger source driven by session turn events (issue #1217): agents can now schedule work on another session's turn start/end — e.g. a boss session summarizing a research session when it finishes a turn — instead of only time/file/webhook triggers.

- **`SessionEvent.TurnStart` / `SessionEvent.TurnEnd`** Bus events published beside the existing `session.turn.after` plugin hooks (processor normal path + invoke error/abort paths), deduplicated by `messageID` across the two publishers.
- **`TriggerSession` schema** (`{type:"session", sessionID, event:"turn.start"|"turn.end", agent?, finish?, once?}`) added to the `Trigger` union and exposed through `agenda_schedule`.
- **`AgendaSessionTrigger` subscriber**: global Bus subscription (cross-scope), entry-level filter matching (session/event/agent/finish), per-entry `messageID` dedup, and `FiredSignal` emission into the shared agenda handler. Registered/unregistered in `syncItem`/`teardownItem`; one-shot items auto-complete and drop their registration after the first fire, recurring (`once:false`) items stay persistent.
- **`agenda_watch`** gains a mutually exclusive `onSessionEnd` option (`delay` xor `onSessionEnd` enforced by a zod refine) for first-person "wake me when session X ends a turn".
- **Web UI safety fix**: the agenda form now preserves triggers it cannot render (session/webhook/watch) on edit saves instead of silently dropping them, and shows a notice when such triggers exist.
- Wake indicator shows a label for session triggers; SDK/OpenAPI regenerated; docs and a decision record updated.

## Verification

- `bun test test/agenda/ test/tool/agenda-watch.test.ts` — 86/86 pass (including end-to-end: TurnEnd → reactor → execution → auto-done)
- `bun run typecheck` in `packages/synergy` and `packages/app` — clean
- `bun run quality:quick` — all code-related gates pass (only environment-local actionlint download failure and missing gitleaks remain; both run in CI)
- `bun ./script/generate.ts` — SDK/OpenAPI regenerated; `doc:check` and `localization:check` pass

## Notes

- Old agenda items and old clients remain compatible: trigger unions are additive, persisted items read without strict parsing (no migration needed), and unknown Bus event types are ignored by the frontend.
- Session triggers are event-driven and do not depend on clock ticks or host liveness, avoiding the reliability pitfalls of delay/watch-based patterns.
- `sessionID` is required (no wildcard) to prevent an execution session's own TurnEnd from recursively re-triggering unrelated items.


yzxoi:
Review finding confirmed as valid — the edit branch built the patch but the `agenda.update` call was dropped while restoring the trigger-preservation logic, so edit saves closed the panel without persisting anything.

Fixed in d168e96138: `await sdk.client.agenda.update({ id: props.item!.id, directory: props.directory, agendaPatchInput: patch })` is restored after building `patch`, making the preservation logic live again. The non-blocking suggestion (Chinese dedup-key comment in session-trigger.ts) is also translated to English in the same commit.

Verified: packages/app typecheck clean, prettier clean; agenda suite remains 86/86 + 4/4.

PR #1221, opened 2026-08-20
feat(app): make interface zoom instant-apply

TTAWDTT:
## Summary

- Interface zoom is now **instant-apply**: dragging the slider immediately changes the zoom factor via `platform.desktopZoom.set()` (fire-and-forget), instead of staging a draft and waiting for explicit save
- Zoom persistence is decoupled from the explicit save flow — on failure, an error toast is shown using the new `settings.panel.zoom.saveFailed` i18n string (en + zh-CN + pseudo)
- `restoreInstantZoom()` re-syncs the slider with the live zoom value on discard and after explicit save, mirroring the existing `restoreInstantTheme()` pattern

## Test plan

- [x] `GeneralPanel.markup.test.ts` — 6 pass (zoom markup unchanged)
- [x] `settings-explicit-save.test.ts` — 8 pass (no zoom in save coordination)
- [ ] Manual: drag zoom slider → zoom applies immediately
- [ ] Manual: discard changes → slider reverts to persisted zoom
- [ ] Manual: save settings → slider stays at applied zoom


yzxoi:
Thanks — instant-apply matches the theme pattern nicely. A few issues before merge:

1. **Slider desync on failed apply**: `applyDesktopZoomFactor` only updates `desktopZoomSaved` on success; on failure the controlled slider stays at the dragged, never-applied position until an unrelated re-render. Re-read the live value (`bridge.get()`) in the catch.
2. **Toast storm**: every `onInput` tick fires `set()`; a persistently failing bridge spams one error toast per tick. Throttle to one toast per panel session.
3. **No behavioral tests**: the referenced tests only sniff source text / assert zoom is absent from save coordination. Repo policy wants a behavioral test first — cover apply success, apply failure + re-sync, and discard re-sync.
4. **Test-plan wording**: "discard → slider reverts" is inaccurate — zoom applies instantly, so discard re-syncs the slider to the live applied value.

I'll push fixes for 1–3 to this branch.

yzxoi:
Follow-up after manual testing: applying zoom on every `input` tick rescales the settings page under the pointer, which shifts the slider thumb mid-drag and makes the drag unstable. Changed the interaction to **preview locally while dragging, commit once on pointer release** (`onChange`), so the page never rescales mid-drag. The failure re-sync + single-toast throttling from the previous commit still applies on release. Keyboard arrow keys keep native range semantics (input without commit until `change`).

PR #1224, opened 2026-08-20
fix(test): restore ToolScheduler singleton in shard-process tests

yzxoi:
## Summary

Fixes the CI test failures in `SessionProcessor auto-expand interception` (packages/synergy/test/tool/auto-expand.test.ts) that started with the agenda session-trigger commit (`f7b8e68956`) and broke every dev CI run.

**Root cause:** `ToolScheduler` is a module-level singleton whose `stop()` closes admission (`accepting = false`) without restoring it. `test/session/tool-scheduler.test.ts` ends with `await ToolScheduler.stop()`. Bun runs each test shard in one shared process, so the stopped singleton leaked into sibling files in the same shard. The agenda commit added two test files, shifting the shard boundary so `auto-expand.test.ts` landed in shard 4 right after `tool-scheduler.test.ts`; its real `ToolScheduler.dispatch()` then rejected with "Tool scheduler is stopping" and the three success-path tests failed (expected `completed`, received `error`).

**Fix (test-side):**

- `test/session/tool-scheduler.test.ts`: `afterAll(() => ToolScheduler.configure())` restores the singleton after the shutdown suite.
- `test/tool/auto-expand.test.ts`: defensive `afterAll` stops + reconfigures the scheduler so this file leaves the singleton clean regardless of shard order.
- Added decision record `docs/decisions/implemented/testing/2026-08-20-tool-scheduler-singleton-restore.md`.

Production code is untouched — the `accepting` gate is a deliberate shutdown semantic and should not mask test leakage.

## Verification

- Reproduced the exact CI failure locally before the fix: `bun test --shard=1/1 test/session/tool-scheduler.test.ts test/tool/auto-expand.test.ts` → 3 fail (same assertions as CI).
- After the fix: same command → 29 pass / 0 fail.
- `bun run test:ci` (packages/synergy): 1711 pass / 1 skip / 2 fail — the 2 failures are pre-existing environment-dependent suites unrelated to this change (provider proxy network test, embedding model-load hook timeout; both also fail without this change on this machine).
- `bun run --cwd packages/synergy typecheck` passes; `format:check` and `decision:check` pass. `quality:quick` only fails the actionlint gates because the binary cannot be downloaded in this environment (unrelated).

### DIFF
--- a/docs/decisions/implemented/testing/2026-08-20-tool-scheduler-singleton-restore.md
+++ b/docs/decisions/implemented/testing/2026-08-20-tool-scheduler-singleton-restore.md
@@ -4,9 +4,7 @@ Status: implemented
 
 ## Problem
 
-On 2026-08-20, every CI run containing commit `f7b8e68956` (feat(agenda): session turn-event triggers) failed the same three `SessionProcessor auto-expand interception` tests in `packages/synergy/test/tool/auto-expand.test.ts`, expecting `completed` and receiving `error`. `test/workspace/policy.test.ts` was green throughout. The failure started the moment the agenda commit merged into `dev` and reproduced on every later run, on Linux CI only.
-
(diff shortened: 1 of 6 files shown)


yzxoi:
Addressed both P1 review comments in commit `b94e4be06`:

1. **Incident narrative moved to a postmortem** — created `docs/postmortem/0002-tool-scheduler-singleton-leakage.md` with the full incident story (summary, timeline, root cause, why the safety nets missed it, guardrails, lessons) and added it to the postmortem index. The decision record now keeps only the durable cleanup decision, rejected alternatives, and consequences, and links to the postmortem via a relative link.

2. **Bare references replaced with relative links** — the decision record's `## Consequences` now links `[postmortem 0002](../../../postmortem/0002-tool-scheduler-singleton-leakage.md)` and `[test-home isolation guard](./2026-08-18-test-home-isolation-guard.md)` instead of prose references.

Verified: `bun run doc:check` and `bun run decision:check` both pass.

PR #1225, opened 2026-08-20
fix(app): align session top-bar type to one baseline

EricSanchezok:
## Summary

- The desktop session top-bar left cluster (folder icon, project name, `/`, model name, thinking effort) mixed three type treatments on one row: 18px/600 project name, 21px IBM Plex Mono slash, 18px/650 model button, and 15px/520 effort label.
- Unified the cluster on one sans family, one 20px line box, and the semantic size/weight tokens: project name/slash/model use `font-size-base` (slash stays regular), effort stays one step quieter at `font-size-small`/`medium`.
- Selector buttons shrink from 38px to 32px to match the text box; folder icon + project name group as `.stb-project` with consistent 8px gaps; removed the overlapping negative selector margin and local pixel sizes.

## Files

- `packages/app/src/components/top-bar/session-top-bar.css`
- `packages/app/src/components/top-bar/session-top-bar.tsx`
- `packages/app/PRODUCT.md`
- `packages/app/test/workbench-surface.test.ts`
- `docs/decisions/implemented/bug-fix/2026-08-20-session-topbar-type-alignment.md`

## Verification

- `bun test test/workbench-surface.test.ts` (packages/app): 17 pass, 0 fail
- `bun run --cwd packages/app typecheck`: clean
- `bunx prettier --check` on all changed files: clean
- `bun run decision:check`: passed
- Visually verified in an isolated dev runtime: project name, slash, and model now share one baseline; effort label uses the smaller/medium treatment on the same line box.

PR #1231, opened 2026-08-20
chore(coverage): drop duplicate interface-zoom exemption

EricSanchezok:
## Summary

`script/coverage-exempt.json` contained two identical `src/components/settings/panels/interface-zoom.tsx` exemptions (one merged via dev, one via PR #1225), so the app coverage gate failed with "matches multiple exemptions" and the Coverage CI job stayed red on dev.

Remove the duplicate entry, keeping a single exemption with the standard settings-panel reason.

## Verification

- `bun run coverage:check` no longer reports the duplicate-exemption error for `interface-zoom.tsx` (1 matching exemption remains).
- No source or contract changes; only the manifest is deduplicated.

### DIFF
--- a/script/coverage-exempt.json
+++ b/script/coverage-exempt.json
@@ -913,10 +913,6 @@
           "reason": "settings panels render through the app shell; unit suite covers hooks/models"
         },
         {
-          "glob": "src/components/settings/panels/interface-zoom.tsx",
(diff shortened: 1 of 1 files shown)


yzxoi:
Duplicated with #1230

PR #1233, opened 2026-08-20
fix(enforcement): classify cp/install/ln source operands as reads

yzxoi:
## Summary

- The enforcement gate classified every path argument of a write-capable shell segment as a write target, so a plain `cp <external-source> <workspace-dest>` surfaced a non-bypassable `file_external_write` capability for its **read-only source** operand and was hard-denied under the `autonomous` profile — even though the only write landed inside the workspace. The same inversion affected `ln -s`, where the read-only link target was classified as the write.
- The gate now resolves operand roles for plain `cp`/`install`/`ln` segments (`copyOperandSources` in `packages/synergy/src/enforcement/gate.ts`): the final positional operand — or the `-t`/`--target-directory` value — is the write target; every other operand classifies as a read, including sensitive-path read checks.
- The role resolution is closed-world and conservative: pipelines, directory changes, shell-state reuse, wrappers, command substitution, glob write targets, quoted operands, and unrecognized flags fall back to the previous all-write classification. `mv` keeps all-write classification because it deletes its sources.
- Import direction is now allowed under `autonomous` when the source is readable; copying **out** to an external destination, `mv` across the boundary, credential sources (still surfacing `secrets`), and dynamic/pipelined spellings keep the previous hard denial.
- Corrected the pre-existing `ln -s` gate test whose expectation encoded the inverted classification, added a dedicated `copy-operands` test file, updated `docs/architecture/execution-boundaries.md`, and added the implemented decision record.

## Test plan

- `cd packages/synergy && bun test test/enforcement/copy-operands.test.ts` — new behavior, 10/10 pass (written red-first: 5 expected-allow cases failed before the fix)
- `cd packages/synergy && bun test test/enforcement/` — 1189/1189 pass
- `cd packages/synergy && bun test test/control-profile test/sandbox test/permission` — 543/543 pass
- `cd packages/synergy && bun test test/session test/workspace` — 1096/1098; the 2 `SessionInbox` timeouts were verified (via stash) to fail identically **without** this change — pre-existing flakiness, unrelated
- `bun run typecheck` (packages/synergy), `bun run decision:check`, `bun run doc:check` — pass
- `bun run quality:quick` — 13/14 gates; the only failure is `secrets:check` requiring the local `gitleaks` binary (not installed here; covered by CI)


yzxoi:
Addressed all 7 review findings in 021a26d12 (TDD: 8 new red tests first, then the parser rework).

- **P1 quoted operands** — `resolveCopyOperands` now tokenizes with a quote-aware `shellTokenize` that preserves `"/tmp/out dir/file"` as one operand; any residual quote/escape, expansion (`$`, backtick), glob (`*?[`), redirection (`> < | ;`), or comment syntax anywhere in the segment returns `undefined` → conservative all-write. `cp ./src "/tmp/out dir/file"` is denied as an external write.
- **P1 hard links** — `ln` without `-s`/`--symbolic` and `cp -l`/`--link` return `undefined` (all-write fallback): a hard link mutates the source inode's link count and a later write through the link pierces the external file. `ln -s` (symlink) still classifies the target as a read.
- **P1 `-t` destination unclassified** — the copy branch now classifies `copyOperands.target` explicitly when the generic extractors skip flag values, so `cp -t ../outside ../source` denies with an external write on `../outside`.
- **P2 identical source/destination** — role assignment uses `candidate === target` first (`isTarget || !sourceRead`), so `cp --force --backup=numbered /x/file /x/file` keeps the destination write role.
- **P1 glob sources** — `copyOperandPath` rejects `*?[` in **every** operand (source and target), so `.[s]sh` spellings fall back to all-write instead of bypassing credential-root matching.
- **P1 redirections** — `shellTokenize` returns `undefined` on `>`/`<`/`|`/`;`, so `cp ./src /tmp/external-copy > ./log` keeps the conservative classification with `/tmp/external-copy` as an external write.
- **P2 short option values** — optional-value short flags (`install -m 0755 …`) now consume their attached value (`-m0755`) or following value; `install -m 0755 /external/src ./dest` is allowed with the source as a read.

Tests: `copy-operands.test.ts` now 19/19 (8 review-derived cases added red-first); full `test/enforcement/` 1198/1198; `test/control-profile test/sandbox test/permission` 543/543; typecheck, `decision:check`, `doc:check` pass.

PR #1235, opened 2026-08-21
fix(ci): pack publishable manifests from staged copies in package:check

yzxoi:
## Summary
- Post-merge dev push (run 32439413071) failed `format:check` on `packages/plugin/package.json` even though the identical tree passed the PR run — a gate-concurrency race, reproduced locally (Prettier poll hit the rewrite window within 57 iterations).
- `package:check` used to rewrite each tracked `packages/*/package.json` **in place** with the publishable variant, pack, then restore in a `finally`. `ci-static` runs up to 4 gates concurrently, so `format:check`'s repo-wide scan could sample a manifest mid-window.
- New `stagePackablePackage` copies the package to a temp staging dir and packs the publishable manifest from there (same pattern `validateSynergyWrapper` already uses). The working tree is never touched; tarball payloads unchanged (identical shasums before/after).
- `.prettierignore` gains `packages/sdk/js/openapi.json` — the SDK build inside the same gate transiently writes this generated file, closing the remaining window of the same class.

## Test plan
- [x] `bun test --config /dev/null test/script/release/package-check-staging.test.ts` — source tree stays byte-identical, tarball carries publishable manifest + `files`-filtered payloads
- [x] `bun run package:check` green, zero workspace pollution, identical tarball shasum vs pre-fix
- [x] 200 Prettier polls during a full gate run: 0 failures (pre-fix: hit within 57)
- [x] `decision:check`, `test-layout:check`, oxlint, `typecheck`, format all pass
- [ ] CI green on this PR


yzxoi:
Review findings addressed in ded6f770a:

**Blocking — unguarded `await main()`** ✅ Fixed. Entry point now guarded with `if (import.meta.main) await main()` (matching the 17-script repo convention). Verified: the staging test runs in **95ms** (was silently executing the entire gate in ~22s), and `bun run release:test` completes its 139 tests in 5.3s with no gate re-execution. Direct `bun script/package-check.ts` still runs the full gate (exit 0).

**Suggestion — dead `?? tempRoot`** ✅ Fixed. Temp root is now initialized once in `beforeAll`; both tests share the fixture root, no dead fallback.

**P1 — operations doc stale** ✅ Fixed. `docs/operations/open-source-quality.md:101` now describes temp-staged packing and states tracked package.json files are never rewritten; the decision record also notes the `import.meta.main` guard.

Thanks @chatgpt-codex-connector and the synergy-agent review — the import-time execution catch was a real defect in my test (it "passed" locally only because the full gate also passed in this environment).

PR #1237, opened 2026-08-21
perf(provider): extend late-user-context prompt caching to openai-compatible transports and Anthropic

yzxoi:
## Summary

- Extend the `late-user-context` prompt layout to every `@ai-sdk/openai-compatible` transport and the native `anthropic` provider, so volatile advisory context (memory recall, environment block, reminders) is emitted after the append-only history as a final `<runtime-context>` user message instead of between the stable system and history.
- Previously these providers fell into the `system` layout, where per-turn-volatile late system blocks sat **before** history and broke prefix-cache matching every turn — the whole history was re-billed at full input price on each call. Mainstream providers behind the SDK (DeepSeek, Zhipu GLM, Qwen, Kimi, custom gateways) all implement automatic prefix caching priced at 10–30% of input, so the loss was material.
- On the late-user-context layout, `applyCaching` now marks a second `cache_control` breakpoint on the last history message, so Anthropic's explicit-breakpoint caching (0.1× reads) covers the tool-loop transcript incrementally. The `<runtime-context>` message itself stays outside every breakpoint because it changes each turn; a breakpoint ending inside it would only waste 1.25× write cost.
- System-layout providers (Bedrock, Vertex-Claude) intentionally keep the single system breakpoint: their volatile late system blocks sit before history, so a history breakpoint could never be reused across turns.
- Consecutive user messages (tool results followed by the runtime-context message) are valid on the current Anthropic Messages API.

## Decision records

- `docs/decisions/implemented/architecture/2026-08-21-openai-compatible-late-user-context.md`
- `docs/decisions/implemented/architecture/2026-08-21-anthropic-late-user-context-breakpoint.md`
- Architecture doc updated: `docs/architecture/llm-loop.md` (Prompt Assembly section)

## Test plan

- `cd packages/synergy && bun test test/provider/prompt-cache-policy.test.ts test/provider/transform.test.ts test/session/kvcache-measurement.test.ts` — 112 pass, 0 fail (new coverage: openai-compatible routing, Anthropic history-tail breakpoint, runtime-context exclusion, array-content handling; legacy system-layout behavior re-homed on a Vertex-Claude model)
- Broader regression: `bun test test/provider/ test/session/kvcache-measurement.test.ts test/session/llm.test.ts` — 376 pass, 0 fail
- `bun run --cwd packages/synergy typecheck` — clean
- `quality:quick`: 13/14 gates pass locally; the only failure is the `secrets:check` gate because `gitleaks` is not installed on this machine (CI runs the secret scan).


yzxoi:
@synergy-agent review latest

PR #1227, opened 2026-08-20
feat(files): preview workspace PDFs inline via content route

EricSanchezok:
## Summary

The Files workbench previously showed a "Binary files do not have a text preview" placeholder for workspace PDFs. This change adds a real, paginated, zoomable PDF preview by streaming the file bytes through a new scoped route and reusing the attachment workbench's existing pdfjs renderer.

- **Backend**: new `GET /workspace/files/content` (`operationId: workspace.files.content`) streams PDF bytes with `Content-Type: application/pdf` and `Cache-Control: no-store`. It reuses the existing workspace path checks (lexical/symlink escape, Scope required), accepts `.pdf` or `application/pdf` MIME, rejects non-PDF files (`WorkspaceFileUnsupportedPreviewError`, 400) and files over 50 MiB (`WorkspaceFileTooLargeError`, 400), and keeps 403/404 error shapes.
- **Contract**: the JSON `workspace.files.read` union is unchanged — PDFs still read back as `kind: "binary"` metadata. The generated SDK gained `workspace.files.content` (binary/blob response).
- **Frontend**: `context/file` owns a separate PDF byte cache (50 MiB per-file cap, two-buffer LRU protecting open tabs, abort on switch, shared concurrency gate). `file-workbench/content.tsx` classifies PDFs as preview-only and renders the shared `AttachmentPdfPreview` with Files-native loading / too-large / error-retry states. PPTX, DOCX, and other binary files keep the existing placeholder.

## Verification

- Backend: `bun test test/server/workspace-files.test.ts` (28 pass) + `test/workspace-file/service.test.ts` + `test/util/document.test.ts` (45 pass) — covers PDF streaming, uppercase extension, >50 MiB, non-PDF 400, lexical/symlink escape 403, missing 404, missing Scope 400, JSON read stays binary.
- Frontend: `model.test.ts` + `pdf-render.test.ts` (10 pass), `bun run typecheck` clean, `bun run build` succeeds with pdfjs worker emitted as a `'self'`-loadable asset.
- Repo gates: `bun run quality:quick` — all 14 gates pass; `bun run localization:check` passes (zh-CN translations added).
- Docs: architecture doc, file-workbench README, and decision record `docs/decisions/implemented/feature/2026-08-20-workspace-pdf-preview.md` updated; `decision:check` passes.

## Notes

- No persistence, config, or migration changes.

PR #1239, opened 2026-08-21
feat(files): rework PDF preview onto the official pdf.js viewer components

EricSanchezok:
## Summary

Fixes three PDF preview defects in both the Files workbench and the attachment workbench (they share one component, `AttachmentPdfPreview`):

1. **Continuous scroll** — replaced the hand-rolled single-canvas, single-page renderer with the official pdf.js viewer components (`PDFViewer` + `EventBus` + `PDFLinkService` from `pdfjs-dist/web/pdf_viewer.mjs`). All pages now live in one vertical scroll container with lazy per-page rendering and off-screen canvas recycling. The prev/next toolbar buttons remain as page-jump shortcuts.
2. **Fit-width works** — the button is now a real toggle between `page-width` and 100% scale, with correct pressed-state feedback. The host re-applies the `page-width` preset on container resize because the viewer components' own ResizeObserver only tracks height.
3. **Selectable, copyable text** — the pdf.js text layer is enabled, so users can select and copy PDF text natively (the app layout's `select-none` is reverted on the PDF container, same as the markdown preview).

## Key details

- `pdf_viewer.mjs` reads its core API from `globalThis.pdfjsLib` (with a version check), so the component assigns `globalThis.pdfjsLib ??= await import("pdfjs-dist")` before importing the viewer module.
- The vendor stylesheet is imported through a shim (`pdf-viewer-vendor.css`) that re-asserts the app theme's `color-scheme` at `html:root` specificity, preventing the vendor's `:root { color-scheme: light dark }` from leaking onto the document root.
- `pdf-render.ts` (`createPdfRenderCoordinator`) and its test are removed — the component was their only consumer.
- `pdfjs-dist` stays pinned at 5.4.296; no backend, SDK, or data-plane changes.

## Verification

- New Playwright integration test (`pdf-preview.dom.test.ts`, registered in the serial Chromium batch): multi-page continuous scroll DOM, text-layer selection under a `user-select: none` ancestor, fit-width toggle semantics (initial pressed, toggle off/on, zoom clears, fit restores), page-position label, prev/next jump, and vendor CSS color-scheme non-leak.
- `bun run --cwd packages/app test` — full suite passes (1675 tests, 0 fail).
- `bun run --cwd packages/app typecheck` — passes.
- `bun run --cwd packages/app build` — passes; `pdf.worker.min.*.mjs` and `pdf_viewer-*.js` chunks verified in the production bundle.
- `bun run quality:quick` — all 14 gates pass (format, lint, typecheck, doc:check, decision:check, deadcode, etc.).


yzxoi (approved):
**Approve** — verified locally on a clean cherry-pick of both commits, plus an end-to-end manual pass on an isolated dev instance (continuous scroll, native text selection, fit-width across container resize). Excellent rework.

Automated verification:

- New Playwright suite (`pdf-preview.dom.test.ts`): **4/4 green** (continuous scroll + text selection, dark color-scheme, resize re-fit, corrupted-PDF error state)
- `bun run typecheck` (packages/app): pass
- Existing `file-workbench/` + `attachment-workbench/` suites: **25/25 green**
- `pdf-render.ts` has no remaining consumers repo-wide; removal is clean
- `globalThis.pdfjsLib ??=` is the only assignment repo-wide — no conflicts

Spot-checks worth recording:

- The `abortSignal` cast is legitimate: `pdfjs-dist@5.4.296` types omit the option, but the runtime does consume it (`watchScroll`, `web/pdf_viewer.mjs:7673-7681` in the installed package).
- The `pdf-viewer-vendor.css` shim is consistent with the boot script's inline `color-scheme` (which outranks any stylesheet in the real app); the shim protects the un-bootstrapped test fixture and adds defense-in-depth.
- All three follow-ups from the #1227 review are confirmed landed: `WorkspaceFileContentError` schema, the 403 response shape, and the `pdfAccess` LRU rename migration.

Two non-blocking nits, fine as follow-ups:

1. Consider a one-line comment on the `abortSignal` cast noting the runtime-vs-types gap, so it can be dropped once types catch up.
2. The error state renders `1 / 0` in the page-position label (the test even asserts it) — purely cosmetic; could hide the label when `error()` is set.

PR #1241, opened 2026-08-21
fix(sidebar): render project menu via Kobalte popover to escape scroll clipping

TTAWDTT:
## Summary

Fixes the project row's "⋯" (three-dot) menu in the sidebar **项目/Projects** section being clipped by the sidebar's scroll container.

**Root cause:** the menu was `position: absolute` (`top: 100%` / `.sb-project-menu-up` → `bottom: 100%`) anchored inside `.sb-scroll`, which is `overflow-y: auto`. Any absolutely-positioned popover inside a scroll container gets clipped, so with no sessions expanded the menu was cut off entirely, and with sessions expanded only the first item (固定/Pin) was visible.

**Fix:** rewrite the project menu with [Kobalte](https://kobalte.dev) `Popover`. Kobalte renders content through a `Portal` to `document.body` (immune to scroll-clipping ancestors) and automatically flips/positions by the viewport, matching the previous intent ("fixed + viewport measurement") without re-introducing the `position: fixed` ancestor-transform containment trap.

## Changes

- `packages/app/src/components/sidebar/sidebar.tsx` — replace the hand-rolled `menuPlacement`/fixed-position menu with a Kobalte `Popover` (`placement="bottom-end"` + `gutter`), keeping the existing Pin/Unpin, Edit, Archive items and the "+" button.
- `packages/app/src/components/sidebar/sidebar.css` — remove `.sb-project-menu-backdrop` / `.sb-project-menu-up` and the hand-rolled absolute positioning; keep visual styling on `.sb-project-menu` (Kobalte owns positioning).
- `packages/app/src/locales/messages.ts` — add `sidebar.projectMenu` i18n key for the menu's accessible label.
- `packages/app/src/components/sidebar/project-menu-placement.ts` — **deleted** (obsolete).
- `packages/app/test/components/sidebar/project-menu-placement.test.ts` — **deleted** (tested the removed module).

## Testing

- Verified locally: `sidebar.tsx` transforms without JSX/type errors under the dev server.
- Manual check (dev frontend attached to a running backend): menu opens in both collapsed and expanded project states, all three items (固定/编辑/归档) are visible, and the menu is not clipped by the scroll container.


TTAWDTT:
@yzxoi help

yzxoi (changes requested):
# PR #1241 Review — request changes

## Summary

The Portal direction is right and the implementation matches existing patterns (`note/panel.tsx`, `MenuField`), but testing on an isolated dev instance built from this branch surfaced a **must-fix visual regression**: the menu renders with a fully transparent background.

### 🔴 Blocking: menu background / border / item hover are all broken (CSS custom property scope)

The popover content is portaled to `document.body`, but `.sb-project-menu` / `.sb-menu-item` still resolve sidebar-scoped custom properties defined on `.sb-root` (`sidebar.css` lines 1–8): `--sb-popover-bg`, `--sb-border`, `--sb-hover-bg`. Outside `.sb-root` these resolve to the empty string, so with the menu open:

- `background: rgba(0, 0, 0, 0)` — menu text floats over the page content
- `border: 0px none` — no border
- item `:hover` background is also broken (`--sb-hover-bg` resolves empty)

Verified via computed styles on this branch with the menu open: `backgroundColor: rgba(0,0,0,0)`, `--sb-popover-bg: ""`, and the content node is not a descendant of `.sb-root`.

Suggested fix — reference the global tokens directly (same approach as the other portaled menus; `note/panel.tsx` already uses `--surface-raised-stronger-non-alpha` on its portaled content):

```diff
 .sb-project-menu {
   …
-  background: var(--sb-popover-bg);
-  border: 1px solid var(--sb-border);
+  background: var(--surface-raised-stronger-non-alpha);
+  border: 1px solid var(--border-weaker-base);
   …
 }

 .sb-menu-item:hover:not(:disabled) {
-  background: var(--sb-hover-bg);
+  background: var(--surface-hover-base);
 }
```

(Alternatively, re-declare the needed vars on `.sb-project-menu` itself.)

Applied this locally and verified: background `rgb(250,252,255)`, border `rgb(236,238,252)`, hover resolves `#f5f7ff` — identical to the pre-PR rendering, and all three tokens exist in both the light and dark variable tables (`theme.generated.css`).

### 🟡 PR description: the old clipping bug is narrower than described

A/B-tested the `dev` implementation: with a 560px-tall viewport and the project row flush against the bottom of `.sb-scroll`, the initial `down` placement is clipped by ~110px, but the `queueMicrotask` flip then switches to `up` and the menu ends up fully inside the scroll area. The "menu invisible / only first item visible" case reproduces when the scroll content is too short to fit the menu in either direction (e.g. all projects collapsed and no recent sessions). Not a blocker — the Portal fix is still the right call — just useful context for writing a regression repro.

### 🟡 Pre-existing (both old and new): `.sb-menu-item-danger` color token doesn't exist

`.sb-menu-item-danger` uses `var(--text-critical)`, which is not defined anywhere — the family is `--text-on-critical-base` / `--text-on-critical-weak` / … (see `.sidebar-account-menuItem--danger` for the working usage). So the Archive item has never actually rendered red. Happy to see it fixed here (`--text-on-critical-base`) or in a follow-up.

### ✅ Verified working

- Escape closes, outside click closes, focus returns to the trigger
- Positioning/flip correct in both collapsed and expanded project states; no clipping by `.sb-scroll`
- No leftover references to `project-menu-placement` / `sb-project-menu-backdrop` / `sb-project-menu-up`
- i18n key registered in en / zh-CN / pseudo catalogs

### Test suggestion

A lightweight structural test in the style of `sidebar-account-menu.test.ts` (e.g. asserting the menu renders through `KobaltePopover.Portal` and its colors resolve against global tokens rather than `.sb-root`-scoped ones) would guard against regressing back to absolutely-positioned, scope-dependent styling.

yzxoi:
will be some commits followed by.

yzxoi (approved):

PR #1242, opened 2026-08-21
fix(app): key sidebar update notice by object identity

yzxoi:
## Summary
- **Root cause**: The sidebar update card (`SidebarAttentionNotice`) froze at the first observed progress value (e.g. "Downloading Synergy / 0%") while the Settings page showed the live 87%. Both UIs share the same `desktopStatus` signal — the difference is reactivity. The card is rendered through a **non-keyed** Solid `<Show when={props.notice}>` whose child eagerly captures the notice object (`const n = notice()`). Non-keyed `Show` only re-runs its child when the condition toggles truthiness (`equals: (a, b) => !a === !b`); once the notice was truthy (phase `downloading`, percent 0), every later `download-progress` event updated the signal but never re-ran the child, so `n.progress` stayed 0 forever.
- **Fix**: Key the `<Show>` by object identity (`keyed`) and read the keyed child argument as the value directly. `attention()` in `product-update.tsx` is a `createMemo` that builds a fresh `ProductUpdateNotice` on every status change, so each progress event yields a new identity and the child re-runs with the latest progress. Behavior when the notice disappears is unchanged (child torn down).

## Why
User report (with screenshot): at the same moment the sidebar showed "正在下载 Synergy / 正在下载 0% / 进度条 0%" while the Settings page showed "正在下载更新, 87%". Single event source, single renderer state — the divergence was entirely this Solid reactivity pitfall.

## Verification
- New Playwright regression test `packages/app/test/components/sidebar/sidebar-attention-notice.dom.test.tsx`: mounts the component via a Vite fixture, sets the notice to `downloading` at 0%, then updates it to a new object at 87%, and asserts the progress bar style and detail text follow.
- **Fails on pre-fix code**: with the non-keyed `Show` restored, the test observes the progress bar stuck at `--sb-attention-progress: 0%` (exact reported symptom); **passes with the fix** (87% rendered).
- `bun test test/components/sidebar/` → 74 pass, 0 fail (includes the new DOM test).
- Gates: `format:check`, `lint` (0 warnings), `decision:check`, `test-layout:check`, `packages/app` `typecheck` (tsgo -b) all green.
- Decision record: `docs/decisions/implemented/bug-fix/2026-08-21-sidebar-attention-keyed.md`.

## Files changed
- `packages/app/src/components/sidebar/sidebar-attention-notice.tsx` — keyed Show + direct value read (2 lines)
- `packages/app/test/components/sidebar/sidebar-attention-notice.dom.test.tsx` — new regression test
- `packages/app/script/test.ts` — register the Playwright suite in the serial isolated batch
- `docs/decisions/implemented/bug-fix/2026-08-21-sidebar-attention-keyed.md` — decision record

## Notes
A sibling pattern audit found one structurally similar non-keyed `Show` child in `subagent-dock.tsx` (retry status), but its `when` value is a non-reactive snapshot taken inside a tooltip content builder, so it cannot be fixed by keying and is not affected by this bug; left untouched to keep the change minimal.


yzxoi:
@synergy-agent review

PR #1243, opened 2026-08-21
fix(app): exempt mobile drawer header from Electron drag region

TTAWDTT:
## Summary

On frameless Electron windows (Windows `frame: false`), the always-rendered top
titlebar band is a `-webkit-app-region: drag` region. The mobile drawer is a
full-viewport `position: fixed; inset: 0; z-index: 100` overlay that paints over
that drag band. Electron's drag-region hit-test takes precedence, so it swallows
`hover` + `click` on everything beneath it — including the drawer header row whose
top-right is the close X.

The close button sits at y≈8–40px, overlapping ~87% of the 36px drag strip, making
the X effectively unclickable on narrow (<768px) desktop windows where the mobile
drawer is the active sidebar variant.

## Fix

Exempt the drawer header row from the drag region with `WebkitAppRegion: "no-drag"`.
This is the same pattern the desktop titlebar already uses for its interactive
`--brand` / `--controls` elements (`desktop-window-chrome.css`). The logo link and
close button receive pointer events again. The property is inert in non-frameless /
non-Electron contexts.

## Verification

- Strict typecheck of the edited region passes with the project's solid-js settings
  (the non-standard `WebkitAppRegion` key requires an `as any` cast on the style object).
- Existing Electron behavior confirmed against electronjs.org docs: "draggable areas
  ignore all pointer events... setting `app-region: no-drag` reenables pointer events."
- The Electron drag-region swallow is a webContents behavior that cannot be reproduced
  in headless Chrome/jsdom, so a mock-based DOM test cannot assert it; the prior broken
  fixture test was removed rather than left failing.


yzxoi:
@synergy-agent review

yzxoi (approved):

PR #1244, opened 2026-08-21
fix(app): render error-page version from the Show accessor value

yzxoi:
## Summary

- Fix both error-page footers (fatal-error and initialization-error) that rendered the minified source of a Solid `Show` accessor instead of the version string. The `<Show>` callback parameter is an accessor function and must be called before i18n interpolation: `i18n._(…, { version: version() })`. Previously Lingui stringified the function itself, so production builds printed `()=>{if(!k(r))throw hn("Show");return n()}` as the "Version" line.
- Add a Playwright DOM regression fixture `packages/app/test/pages/fatal-error.dom.test.tsx` (registered in the serial Playwright batch in `packages/app/script/test.ts`): asserts the literal version text, the buildLabel-over-version preference, and the same footer path on the initialization-error page. Against the unfixed code the fixture fails by rendering the accessor source.
- Record the `Show`/`Index` accessor rule in the `develop-frontend` skill and add an implemented bug-fix decision record `docs/decisions/implemented/bug-fix/2026-08-21-error-page-version-accessor.md`.

## Root cause

`packages/app/src/pages/fatal-error.tsx` and `packages/app/src/pages/error.tsx` used the callback form of `<Show>` and passed the callback parameter directly as an ICU interpolation value. Lingui stringifies non-string values, so the rendered footer became the accessor's own (minified) source. All other i18n interpolation sites in the app were audited; none passes an uncalled accessor.

## Tests

- `bun test test/pages/fatal-error.dom.test.tsx` (from `packages/app`) — red on unfixed code, 3/3 green after the fix
- `bun test --conditions=browser test/pages/fatal-error.test.tsx`
- `bun run --cwd packages/app test` — full suite incl. registered serial batches, exit 0
- `bun run decision:check`
- `bunx prettier --check` on every changed file
- `bun run quality:quick`: 3/14 gates failed in this sandbox for environment reasons (e.g. `workflow:check` needs an `actionlint` binary that is not installed); not attributed to this change and left to CI.


yzxoi:
Addressed in 368055f9f — good catch. Verified both cited sites (`reward-radar.tsx` chart mounts, `status-bar.tsx` branch label) use `<Show keyed>` and receive raw values, so the rule now scopes the accessor requirement to non-keyed `<Show>` and `<Index>` only, with the keyed/`<For>` raw-value behavior noted alongside. The decision record wording is aligned as well.

PR #1245, opened 2026-08-21
fix(session): recover queued inbox tasks after abort and prune ghost inbox rows

yzxoi:
# fix(session): recover queued inbox tasks after abort and prune ghost inbox rows

## Summary

Two user-visible defects shared one trigger: a message queued while a run was active, followed by an explicit abort.

- **Stranded durable task after abort (regression).** A user message queued during an active run lands in the session inbox as a durable `task` item — the loop's abort path intentionally keeps task items and discards steer/context. But two later changes removed every path that would wake the session to consume it: terminal assistant errors became a thrown `SessionTerminalError` (`7c6ad5619`), and `loop()`'s failed exit suppressed the release-time pending-work drive (`4a97fb40d`, `requestNextWorkOnFailure: false`). An abort is a loop failure, so the surviving task stranded in the inbox until the next user message arrived — which then materialized both messages together. `SessionManager.run()` now captures `lease.signal.aborted` **before** `finish()` (release aborts the controller, erasing the distinction) and treats an explicit abort like a completed run for the one-shot pending-work drive. Abort is a user action, so this restores recovery without reopening the failed-wake hammering that `4a97fb40d` stopped for ordinary errors.
- **Ghost inbox row after consumption.** The client applies a queued `session.input()` acceptance by upserting the returned item into the store, guarded only by inbox-resource freshness. When the backend consumed the item before the HTTP response landed (idle transition, reconnect gap, dropped `session.inbox.updated`), the upsert resurrected a row the missed event would have cleared, and badge/popover surfaces render the raw bucket. Forensic scans of live data confirmed the backend never leaves consumed items on disk (consumption is peek-then-commit; zero materialized-but-undeleted items found). The client now treats a canonical materialized user message as proof of consumption: the acceptance-response upsert skips an already-materialized item (`isInboxItemMaterialized`), and a user `message.updated` prunes the matching ghost from the store bucket (`removeMaterializedInboxItems`).

## Tests

- `cd packages/synergy && SYNERGY_TEST_HOME=1 bun test test/session/invoke.test.ts test/session/manager.test.ts test/session/inbox.test.ts` — 87 pass (includes new red→green abort-recovery test)
- `cd packages/app && bun test test/components/session/session-inbox.test.ts test/context/global-sync.test.ts test/context/session-volatile-resync.test.ts test/context/session-message-window.test.ts` — 72 pass
- `bun run decision:check`, `bun run doc:check`, `bunx prettier --check` on changed files — pass
- `bun run quality:quick` — all local gates pass except `secrets:check` / `workflow:check`, which need `gitleaks` / `actionlint` binaries not installed on this machine (CI covers both)

## Docs

- Decision record: `docs/decisions/implemented/bug-fix/2026-08-21-abort-inbox-stranding.md`
- `docs/architecture/llm-loop.md` — abort release still schedules the pending-work drive
- `docs/architecture/frontend-data-sync.md` — guarded mutation upsert + reactive ghost prune


yzxoi:
Thanks for the review — all three findings are addressed in 17a21713d.

**P1 · Distinguish user aborts from internal cancellations** — valid race, fixed with explicit recovery intent instead of abort-state sniffing. `signalAbort`/`SessionInvoke.cancel`/`SessionAbort.abort` now accept `recoverQueuedTasks`; only the user-facing abort entries (the abort route and the `session_control` abort action) set it. The loop owner records the flag when the **first** abort wins, and `SessionManager.run()`'s release path reads the owner flag before `finish()` (which aborts the controller and clears the owner). Internal cancellations — Boss task cancel, Lattice run cancel/pause, Cortex timeouts, Light Loop cancel — abort before removing their own inbox items and leave the flag unset, so release never races `removeRunInbox`/pending-item cleanup anymore. A user abort arriving after an internal cancellation is a no-op for the drive (first abort wins), trading one recovery opportunity for cancellation correctness; that trade-off is recorded in the decision record. Covered by a new red→green test pair: user abort schedules the release drive, internal cancellation does not.

**P2 · History-mode pending messages** — valid gap. `isInboxItemMaterialized` now also matches the window's `pendingLatestIds`: history mode records unseen canonical arrivals there instead of inserting into `messages`, so a late acceptance response can no longer re-insert an already-materialized item during history mode. The submit guard passes the pending IDs through, and a unit test covers the case.

**P2 · Materialization precedes commit** — valid crash window. The event-side prune in `global-sync.tsx` now schedules one authoritative inbox refresh after pruning (the existing debounced `refreshInbox`). An item still retryable on disk — `materializeItem` succeeded but `commitReady` never ran — comes back from that refresh instead of staying hidden behind an empty local bucket (the idle refresh path only fires when the local bucket is nonempty, so it could not restore it). This keeps the reactive prune for the common ghost case while making the rarer uncommitted case self-correcting.

Docs updated (`llm-loop.md`, `frontend-data-sync.md`, decision record) and tests re-run: backend invoke/manager/abort-hook/inbox suites 90 pass, frontend inbox + sync suites 23 pass.

PR #1246, opened 2026-08-21
fix(ui): replace dead status-color token references with defined theme tokens

yzxoi:
# Summary

- Replace 26 references to status-color CSS custom properties that no theme defines (`--text-critical`, `--text-warning`, `--text-success`, `--text-critical-base`, `--text-warning-base`, `--surface-critical-soft`, `--surface-warning-soft`, `--surface-success-soft`) across 10 app CSS files. Undefined custom properties compute to the initial value, so error text, warning banners, and danger menu items have been silently uncolored in both light and dark themes.
- Each dead reference is swapped to the closest defined family: `--text-on-critical-base` / `--text-on-warning-base` / `--text-on-success-base` for status text, `--surface-*-weak` for tinted backgrounds, and `--icon-critical-base` for the subagent retry indicator ring.
- Extend `packages/ui/test/css-token-integrity.test.ts` with a guard that scans every app/ui css and tsx source file for the eight dead names, so they cannot reappear in any file — the existing per-file enrollment lists kept missing new files, which is how these survived.
- Decision record: `docs/decisions/implemented/bug-fix/2026-08-22-dead-status-color-tokens.md`.

# Affected surfaces

Sidebar archive menu item, settings personalize/account-profile error text and logout action, plugin consent warning/error banners, plugin marketplace state pills and error banners, session transition card icons, session inbox discard action, subagent retry ring, note waiting states, file workbench warning banner, workbench error text.

# Testing

- New guard test fails on `dev` (lists all 26 offenders), passes after the fix: `bun test test/css-token-integrity.test.ts` from `packages/ui` — 20 pass.
- Full ui package suite: `bun run test` from `packages/ui` — 0 fail.
- `bun run decision:check` — passed.
- Note: raw `bun test` with parallel workers shows ~20 pre-existing failures in unrelated domains (Lattice renderers, menu-field, attach gallery) that also fail on a clean `dev` checkout with parallel workers and pass with the documented per-file runner; they predate this change.


yzxoi:
Addressed all three review findings in a313c375b — thanks for the catches.

**1. Scan TypeScript sources in the whole-source token guard** (css-token-integrity.test.ts)

Agreed — `packages/app/src/components/note/extensions.ts` and `packages/ui/src/pierre/index.ts` both embed `var(--…)` references in imperative styles, so a `.ts` blind spot would defeat the guard. The scan now covers `css`, `ts`, and `tsx` (with the Tailwind-bracket extractor applied to both ts and tsx). Verified no current `.ts` source references any blacklisted name, so the widened scan is green; `theme.generated` stays excluded and generator code under `src/theme` contains no `var(--` refs.

**2. Use the standalone error token off critical surfaces** (settings-panel.css)

Agreed — `text-on-critical-base` is only contrast-validated against `surface-critical-weak`, so using it for bare text on ordinary settings surfaces could make errors invisible under a custom theme. Switched the seven unfilled-state usages to the standalone `--text-error` role (resolved from the error ramp for exactly this case, with existing precedent in `checkbox.css`/`switch.css`): settings personalize + account-profile input errors, byte-count error, account logout, sidebar archive item, session-inbox discard action, and workbench error text. Tinted-surface usages (consent/marketplace banners, plugin state pills, note waiting states, file-workbench banner) keep the paired `text-on-*-base` on `surface-*-weak`, since those pairs are covered by the generator's 4.5:1 checks.

**3. Pair transition surfaces with readable icon colors** (session-transition-card.css)

Confirmed your measurements — `icon-success-base` (#60bc76) on `surface-success-weak` (#e8fbeb) is ~2.17:1 and `icon-critical-base` on `surface-critical-weak` is ~2.78:1, both below the 3:1 non-text threshold. The three status-icon rules now use the paired `text-on-success-base` / `text-on-critical-base` foregrounds, which are the tokens the generator contrast-validates against exactly these `surface-*-weak` fills (resolve.ts pair list). The border-color rules keep the icon-ramp color-mix at 26% opacity as decorative accents. The pairs themselves were already covered by the theme contrast checks, so no new pair coverage was needed.

The decision record (`2026-08-22-dead-status-color-tokens.md`) now documents the split-by-rendering-context rule (tinted surface → paired token, bare text on untinted surface → `text-error`, indicator graphics → icon ramp). Full ui suite and the token guard are green locally.

PR #1248, opened 2026-08-22
fix(storage): retry transient EPERM/EACCES in atomic JSON persistence

### LINKED ISSUE #1247: `EPERM: operation not permitted, rename` on Windows when persisting session state — transient file-lock failure crashes subagents; no auto-retry
### What happened?

On Windows, session-state persistence intermittently fails with `EPERM: operation not permitted` while renaming a temp file to the final session state file. This crashes background subagents with `SessionTerminalError: EPERM` and leaves DAG/taskboard nodes in a failed state. Because there is no automatic retry, the failed node requires manual re-dispatch.

Typical error:

```
EPERM: operation not permitted, rename
'C:\Users\zhen.luo\.synergy\data\sessions\home\<session_id>\.tmp-<pid>-<rand>'
    -> 'C:\Users\zhen.luo\.synergy\data\sessions\home\<session_id>\info.json'
```

Surfaced to callers as:

```
SessionTerminalError: EPERM
```

This terminates background subagents (observed on `research-scout`, `code-cartographer`, `inspector`, `implementation-engineer`). Scheduled agenda runs are also affected — a recurring patrol run failed once with the same error on the run itself.

### Expected behavior

- Session-state persistence should not intermittently fail with `EPERM` on Windows.
- If a transient file-sharing race does occur, the write should be retried automatically (a bounded number of attempts with small backoff) rather than terminating the run.
- If persistence still fails after retries, surface a clear retryable error — not a fatal `SessionTerminalError` that kills the subagent and leaves nodes failed for an operator to manually re-dispatch.

### Steps to reproduce

The failure is intermittent and cannot be triggered deterministically on demand, but it recurs frequently under routine load on Windows. To observe it:

1. Run multiple concurrent Synergy sessions/subagents on a Windows machine so session state files are written and `rename`d frequently under `~/.synergy/data/sessions/<scope>/<session>/`.
2. Operate for a period of normal activity (minutes to a few hours); the error appears periodically.
3. Watch for either:
   - the direct error above during state persistence, or
   - a subagent/agenda run terminating with `SessionTerminalError: EPERM`.

Expected frequency: it has recurred on the order of several times over a few days across multiple sessions (verified on Windows 11).

### Environment

```markdown
- OS: Windows (observed on Windows 11)
- Runtime: Synergy on Windows, Node filesystem layer
- Error path: `~/.synergy/data/sessions/<scope>/<session>/` — rename of `.tmp-*` → `info.json`
```

### Relevant logs or screenshots

_No response_

yzxoi:
Fixes #1247.

## Summary

- **Root cause**: every persisted session object (info, messages, parts, dag, todo, inbox, agenda) funnels through `writeJsonAtomic` (`packages/synergy/src/storage/storage.ts`), which performed a single unguarded `Bun.write` + `fs.rename`. On Windows, `rename` maps to `MoveFileEx`; when another process (antivirus, OneDrive, or a cross-process reader) briefly holds a handle without `FILE_SHARE_DELETE`, the call fails with `EPERM`. One transient failure escalated through the invoke loop into a persisted terminal assistant error, killing the subagent via `SessionTerminalError` and leaving DAG nodes failed until manual re-dispatch.
- **Fix**: `writeJsonAtomic` retries the whole write+rename sequence — up to 4 attempts, 50–200 ms exponential backoff, reusing `isRetryableIOError` from `util/io-retry` so only `EPERM`/`EACCES`/`EBUSY` retry while permanent errors (`ENOSPC`, real permission failures) fail on the first attempt. The temp file is unlinked on terminal failure, so no `.tmp-*` residue remains.
- **Read-side hardening**: the diagnostics pending-session scan (`observability/diagnostics.ts`) reads the same `info.json` files cross-process; it now reads through `readFileWithRetry` so a concurrent atomic rename no longer silently drops sessions from the dashboard on Windows.
- Decision record: `docs/decisions/implemented/bug-fix/2026-08-22-storage-atomic-write-transient-retry.md`.

## Test plan

- [x] `cd packages/synergy && bun test test/storage/` — 10 pass (4 new retry tests: transient rename recovery, temp-write recovery, exhaustion propagates the original error code, non-transient fails on first attempt, no temp-file residue on any failure path)
- [x] `cd packages/synergy && bun test test/observability/diagnostics.test.ts test/util/io-retry.test.ts` — 10 pass
- [x] `bun run typecheck` — 11/11 tasks pass
- [x] `bun run decision:check`, `bun run test-layout:check` — pass
- [x] `bun run quality:quick` — 14/15 pass (`secrets:check` requires local gitleaks; CI runs the secret-scan job)
- [x] Full `bun run test:changed` — 5247 pass; 7 failures are pre-existing 5s-timeout flakes (BossRuntime, daemon entry, plugin host-service, embedding API, session migration) — re-verified in isolation and against the unmodified baseline, where they fail identically
- [ ] Windows real-machine confirmation by the reporter (sharing violations cannot be reproduced natively on macOS)

### DIFF
--- a/.synergy/skill/change-persistence/SKILL.md
+++ b/.synergy/skill/change-persistence/SKILL.md
@@ -18,6 +18,7 @@ description: Add or modify Synergy durable state, JSON storage keys, SQLite tabl
 1. Build logical keys through `StoragePath`; use `Storage` for locks, atomic writes, reads, scans, and removal.
 2. Keep independently updated or streamed records independently addressable. Do not rewrite a whole session or collection for one leaf update.
 3. Update derived indexes and events in the same owner transaction/lifecycle as the canonical write.
+4. Preserve the atomic-write transient-retry contract: `Storage` write+rename retries `EPERM`/`EACCES`/`EBUSY` (classified by `isRetryableIOError`) so Windows sharing violations do not fail persistence, permanent errors fail fast, and temp files are removed (with the same transient retry) on the failure path. Do not bypass `Storage` with a bare rename; extend `test/storage/storage-retry.test.ts` when changing write-path failure behavior.
 
 ### SQLite and other domain stores
 
--- a/docs/decisions/implemented/bug-fix/2026-08-22-storage-atomic-write-transient-retry.md
+++ b/docs/decisions/implemented/bug-fix/2026-08-22-storage-atomic-write-transient-retry.md
@@ -1,0 +1,34 @@
+# Decision Record: Retry transient EPERM/EACCES in atomic JSON persistence
+
+Status: implemented
+
+## Problem
+
+On Windows, session-state persistence intermittently failed with `EPERM: operation not permitted, rename '.tmp-<pid>-<rand>' -> 'info.json'`, terminating background subagents and agenda runs through `SessionTerminalError` and leaving DAG/taskboard nodes failed until manually re-dispatched (#1247).
+
+Every persisted session object (info, messages, parts, dag, todo, inbox, agenda) funnels through `Storage.write`/`Storage.update` → `writeJsonAtomic` in `packages/synergy/src/storage/storage.ts`, which performed a single unguarded `Bun.write` + `fs.rename`. Node's `rename` maps to `MoveFileEx` on Windows: when another process holds a handle on the source or target without `FILE_SHARE_DELETE` — antivirus scans, OneDrive sync, or Synergy's own cross-process readers — the call fails with `EPERM`/`EACCES`. Such sharing violations typically clear within milliseconds, but one occurrence escalated fatally: the write error propagated up the invoke loop, was persisted as a terminal assistant-message error, and `selectResultMessage` rethrew it as `SessionTerminalError`.
+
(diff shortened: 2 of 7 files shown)


yzxoi:
Review feedback addressed in 5601a54a0 — thanks @chatgpt-codex-connector and synergy-agent:

- **P1 (document the retry contract)**: the atomic-write retry semantics now live in the canonical storage reference (`docs/reference/storage-and-paths.md`, JSON Storage section) and in `.synergy/skill/change-persistence/SKILL.md` as a durable rule for future storage changes, alongside the decision record.
- **P2 (cleanup can hit the same sharing violation)**: terminal-failure cleanup now goes through `removeTempFile`, which retries transient `EPERM`/`EACCES`/`EBUSY` unlink errors with the same 50–200 ms backoff before suppressing. Covered by a new test that injects a transient `EPERM` on both `rename` and the first `unlink`, asserting the temp file is still removed.
- **Coverage gap (diagnostics read-side hardening)**: added a behavioral test that injects one transient `EPERM` on the `info.json` read and asserts the session still appears in `sessions.pendingReply`.

Verification: `bun test test/storage/ test/observability/diagnostics.test.ts test/util/io-retry.test.ts` — 22 pass; `bun run typecheck`, `decision:check`, `doc:check`, `skill-check` all pass. Decision record updated to match the shipped cleanup semantics.

PR #1249, opened 2026-08-22
fix(attach): auto-expand delivered attachments in session timeline

TTAWDTT:
## Summary

Currently, when the `attach` tool delivers files to the user, the session timeline shows them as a **collapsed one-line card** reading e.g. "add attachment" that requires a click to expand into the attachment gallery. This is a poor UX for the primary purpose of the tool — showing the user the files that were just delivered.

This change hides the attach tool card (`display.toolCard: "hidden"`) and reuses the same upstream path that media-generation tools (e.g. `dalle`, `gpt-image`) already use to surface their results: the auto-expanded `AttachmentGallery`. Delivered attachments now render expanded in the timeline immediately, with no click required.

## Behavior change

- Before: `attach` result rendered as a collapsed tool card in the session timeline.
- After: `attach` result routes to the `tool-attachments` gallery and is auto-expanded.

## How it works

- `packages/synergy/src/tool/attach.ts`: the tool result now emits `metadata.display.toolCard: "hidden"`. The timeline (`session-turn.tsx` → `timelineKindForPart`) turns a completed non-media tool with hidden card + attachments into `"tool-attachments"`, which renders `<AttachmentGallery>` directly (auto-expanded).

## Tests

- `packages/synergy/test/tool/attach.test.ts`: asserts the attach tool emits hidden tool-card display metadata (single-file + multi-file cases).
- `packages/ui/test/components/session-turn-timeline.test.ts`: asserts a completed hidden-card `attach` tool with attachments routes to the `tool-attachments` gallery.

Verified locally: `packages/synergy` typecheck + attach tests pass; `packages/ui` session-turn-timeline, tool-result-presentation, session-turn-activity, and session-turn-projection tests pass. (Two unrelated pre-existing `session-turn-timeline` failures around token/cost label formatting exist on the clean tree and are not affected by this change.)


yzxoi (changes requested):
# PR Review Body: #1249 request changes

The mechanism reuse here is right, and I verified the full routing chain — one blocking issue (CI format gate), the rest is approve-ready.

## Blocking: Quality / `format:check` fails

`packages/ui/test/components/session-turn-timeline.test.ts` has a double blank line between the new test's closing `})` (L1223) and the next `test(` (L1226). Reproduced locally with `bunx prettier --check`; it's the only gate failing in the Quality cluster. `bunx prettier --write packages/ui/test/components/session-turn-timeline.test.ts` (or `bun run format`) and push should turn CI green.

## Verified

- **Routing**: `timelineKindForPart` sends hidden-card + `completed` + visible attachments → `tool-attachments` → auto-expanded `<AttachmentGallery>`; `error` status still returns `"part"` so failed attach calls (e.g. file-not-found) keep their card.
- **No invisible-delivery risk**: attach never sets `presentation.hidden` on its attachment parts, so the `visibleAttachmentParts(...).length === 0 → undefined` branch can't strand a delivery with no rendering.
- **Activity Trace**: `isActivityGroupableTool()` excludes hidden-card tools, so attach results won't be swallowed into activity groups.
- Plain `toolCard: "hidden"` (without `kind: "media-generation"`) is the correct shape — attach shouldn't get `MediaGenerationCard` pending treatment. Matches the `openai-image-shared.ts` precedent minus the media fields.
- The gallery preserves what the collapsed card used to show (filename, size via `metadata.attachment.size`, "Generated by attach" source tag), so no information is lost.

## Notes (no action needed)

- While attach executes (status pending), the timeline now renders nothing for it (previously a pending card). Attach is near-instant so this seems fine — just flagging the trade-off inherent to hiding the card.
- Historical attach results (no `display` metadata) keep rendering as collapsed cards. No backfill needed for a presentation-only preference; if we ever want old deliveries expanded, that belongs in `session/migration.ts`, not a read-time fallback.
- Decision record: I couldn't find an existing display/toolCard decision record in `docs/decisions/` to piggyback on. If this counts as riding the existing display-policy mechanism, feel free to leave as-is; otherwise a short `implemented` record would satisfy the repo policy for behavior changes.

Tests are well-targeted — the UI test pins the distinction that matters (hidden card ≠ media-generation; completed attachments must still surface), and the tool tests cover single- and multi-file emission.

Requesting changes solely for the format gate.

yzxoi (dismissed):
# PR Review Body: #1249 approve (follow-up)

Follow-up to my request-changes round: the only blocker (format gate) is resolved and everything else stands.

- Pushed `style(ui): fix prettier formatting in session-turn-timeline test` removing the stray double blank line; the `Quality` job (format:check) is now green, along with Typecheck, Coverage, Desktop/Windows Checks, Package Validation, Smoke, Secret Scan.
- The failing `Test` job is unrelated to this PR: all 13 failures are in the `arXiv PDF download` suite with `HTTP 404` errors. The same signature fails on the latest `dev` CI runs (`635b12a6`, `0c196c70` — both before this fix was pushed), the original PR head's Test run passed, and the suite passes locally (12/12). Looks like a CI-environment issue affecting that suite, worth a re-run or a separate look on `dev`.
- All findings from the previous round were already approve-ready: error-status fallback, no invisible-delivery risk, Activity Trace passthrough, correct plain `toolCard: "hidden"` shape, no information loss vs. the collapsed card.

Approving — thanks for the clean minimal change.

yzxoi:
# PR Comment: #1249 collapsible follow-up explanation

Follow-up commit `fix(ui): keep delivered attachments collapsible in the session timeline` — pushed after live-instance testing surfaced an interaction regression in the original change.

**What was found**: hiding the tool card routed deliveries to the raw auto-expanded `AttachmentGallery`, which has no collapse affordance. Multi-file or large deliveries permanently consumed vertical timeline space — the ordinary attach tool card had always been collapsible, so this silently dropped a capability. (CI was green; only manual testing caught it.)

**The fix**: the shared `tool-attachments` timeline path now renders through a `BasicTool` collapsible with `defaultOpen`. Deliveries stay immediately visible with zero clicks (the intent of this PR) and regain one-click collapse. Applies uniformly to attach and media-generation tools since they share the path. The trigger header reuses `getToolInfo`/`classifyTool` semantics: icon, title, filename or pluralized count, and a total-size tag.

**Verification**: new DOM behavior test (`session-turn-attachments-collapse.dom.test.ts`) covers expand → collapse → content removed for both an attach tool and a media-generation tool; related timeline/projection/activity suites (128 tests), typecheck, and format all pass. Decision record added under `docs/decisions/implemented/bug-fix/`.

yzxoi (approved):

PR #1255, opened 2026-08-24
fix(provider): cap Anthropic max thinking budget to keep max_tokens > budget_tokens

TTAWDTT:
## What does this PR do?

Caps the Anthropic `max` thinking variant's `budgetTokens` to `⌊model.limit.output / 2 − 1⌋` (mirroring the existing `high` variant's half-output discipline), so the derived `max_tokens` always strictly exceeds `budget_tokens`.

## Why?

Anthropic's API requires `0 < budget_tokens < max_tokens`. The previous `max` variant used `budgetTokens: Math.min(31_999, model.limit.output - 1)`. For any model whose output limit is ≤ 31_999 — e.g. the common 8_192 / 16_384 default-output Claude models — that made `budget_tokens` equal `output - 1`, while `maxOutputTokens()` derives `max_tokens = modelCap - budgetTokens`, collapsing `max_tokens` to `1`. That violates `budget_tokens < max_tokens` and the request fails with an HTTP 400 before any work happens.

On large-output models (> ~64k output) the `31_999` floor still keeps the `max` budget above `high`, so high-throughput models are unaffected.

## How was it tested?

- New regression guard in `packages/synergy/test/provider/transform.test.ts`: a `test.each` over `[8192, 16384, 32000, 64000]` asserts both `high` and `max` satisfy `0 < budget_tokens < max_tokens` (where `max_tokens` is derived from the same budget via `ProviderTransform.maxOutputTokens`).
- Updated the existing output-8192 expectation: the `max` budget is now `4095` (was `8191`).
- `bun test test/provider/transform.test.ts` → 110 pass, 0 fail.
- `tsgo --noEmit` (packages/synergy typecheck) clean.

```bash
cd packages/synergy
bun test test/provider/transform.test.ts
bun run typecheck
```

## Checklist

- [x] `bun run quality:quick` passes — the touched package's gates (format, lint, typecheck, tests) pass for this change. Note: on a Windows checkout the repo-wide `quality:quick` has pre-existing red gates unrelated to this PR (a tracked symlink `packages/app/src/custom-elements.d.ts` parses as broken TS on Windows; plus `gitleaks`/`actionlint` not installed locally). These do not occur on the Linux CI runners; the files this PR touches are clean under `prettier`, `oxlint`, and `tsgo`.
- [x] New or moved tests live under the owning package's `test/` directory (or the root `test/` directory for repository tests) — the new `test.each` is appended to the existing `packages/synergy/test/provider/transform.test.ts`.
- [x] `bun run package:check` passes if package exports, build output, release logic, or publishable packages changed — no package exports/build/release changes.
- [x] Browser capability or App bootstrap changes pass the browser crypto contract and private HTTP browser smoke — N/A, provider transform only.

### DIFF
--- a/packages/synergy/src/provider/transform.ts
+++ b/packages/synergy/src/provider/transform.ts
@@ -616,7 +616,15 @@ export namespace ProviderTransform {
           max: {
             thinking: {
               type: "enabled",
-              budgetTokens: Math.min(31_999, model.limit.output - 1),
+              // Cap the thinking budget to half the output limit (minus one) so the
+              // resulting max_tokens (= modelCap - budgetTokens) always strictly
+              // exceeds budgetTokens. The previous `model.limit.output - 1` collapsed
+              // max_tokens to 1 for any model whose output limit <= 31_999 (e.g. the
+              // common 8 192 / 16 384 default-output Claude models), violating the
+              // Anthropic constraint budget_tokens < max_tokens and yielding a 400.
+              // Mirrors the `high` variant's half-output discipline; on large-output
+              // models (> ~64k) the 31_999 floor still keeps `max` above `high`.
+              budgetTokens: Math.min(31_999, Math.floor(model.limit.output / 2 - 1)),
             },
           },
         }
--- a/packages/synergy/test/provider/transform.test.ts
+++ b/packages/synergy/test/provider/transform.test.ts
@@ -1863,13 +1863,63 @@ describe("ProviderTransform.variants", () => {
           budgetTokens: 4095,
         },
       })
+      // At output 8192 the `max` budget is capped to ⌊8192/2 − 1⌋ = 4095 (same as
+      // `high`); the previous `model.limit.output - 1` = 8191 collapsed
+      // max_tokens to 1 and tripped Anthropic's budget_tokens < max_tokens rule.
       expect(result.max).toEqual({
         thinking: {
           type: "enabled",
-          budgetTokens: 8191,
+          budgetTokens: 4095,
         },
       })
     })
+  })
+
+  describe("Anthropic thinking variants - budget_tokens < max_tokens invariant", () => {
+    // Regression guard for BUG-002: for every model whose output limit fits in
+    // the common 8k–64k band, both the `high` and `max` Anthropic thinking
+    // variants must satisfy the Anthropic API constraint
+    //   0 < budget_tokens < max_tokens
+    // where max_tokens is derived by maxOutputTokens() from the same budget.
+    // The old `max` variant used `model.limit.output - 1`, which for these
+    // limits produced budget_tokens >= max_tokens and yielded an HTTP 400.
+    test.each([8192, 16384, 32000, 64000])(
+      "keeps max_tokens > budget_tokens > 0 for high and max at output %i",
+      (output) => {
+        const model = createMockModel({
+          id: "anthropic/claude-4",
+          providerID: "anthropic",
+          api: {
+            id: "claude-4",
+            url: "https://api.anthropic.com",
+            npm: "@ai-sdk/anthropic",
+          },
+          limit: { context: 200000, output },
+        })
+
+        const variants = ProviderTransform.variants(model) as Record<
+          "high" | "max",
+          { thinking?: { type?: string; budgetTokens?: number } }
+        >
+        expect(Object.keys(variants)).toEqual(["high", "max"])
+
+        for (const key of ["high", "max"] as const) {
+          const budgetTokens = variants[key].thinking?.budgetTokens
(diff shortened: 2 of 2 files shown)


synergy-agent (approved):
### Review: APPROVE (non-blocking notes)

The fix is correct. I verified the arithmetic against `ProviderTransform.maxOutputTokens` (packages/synergy/src/provider/transform.ts:841) and the `variants()` Anthropic branch (transform.ts:594-630), and checked the call site in `session/llm.ts:433` where `max_tokens` is derived from the same `options.thinking.budgetTokens`.

**Correctness**

- For the new `max` budget `b = min(31_999, ⌊output/2 − 1⌋)`, the derived `max_tokens = output − b` (the `budgetTokens + standardLimit > modelCap` branch always applies for `max`), giving `max_tokens ≥ output/2 + 1 > b`. The invariant `0 < budget_tokens < max_tokens` holds for every `output ≥ 4`, across all four tested limits and the shared-context branch of `maxOutputTokens` (where `modelCap = contextLimit − 32_000`; e.g. 200k-context → max_tokens 136_001 > 31_999).
- The previous formula was actually broken for a wider window than the PR text states: `min(31_999, output − 1)` yields `max_tokens = output − 31_999 ≤ 31_999` for **every `output ≤ 63_998`** (not just ≤ 31_999). The fix resolves that whole range; consider tightening the PR/comment wording so future readers know the true bug window.
- Large-output models (`output ≥ 64_000`) are byte-for-byte unaffected: both old and new formulas cap at 31_999.
- No other provider is touched; the `high` variant formula is unchanged.

**Tests**

- The new `test.each([8192, 16384, 32000, 64000])` guard asserts both `0 < budget` and `maxOutputTokens(...) > budget`, and the stale 8191 expectation is correctly updated. Good regression coverage.

**Minor, non-blocking**

1. The new test calls `maxOutputTokens` without `contextLimit`, so the shared-context branch (`modelCap = contextLimit − OUTPUT_TOKEN_HEADROOM`) is not exercised. The invariant holds there arithmetically, but one `contextLimit` case would lock it in.
2. For unrealistically small `output` (≤ 3) the budget floors to 0, violating Anthropic's `0 < budget_tokens`. Pre-existing in the `high` variant and unreachable for real catalog models; no action needed, just noting.
3. Wording nit: "BUG-002" in the test comment is an internal reference with no public meaning; a plain "regression guard" label would read better for external contributors.

None of these block merge. Thanks for catching this — the 8k/16k-default Claude models would have 400'd on every `max` request.

yzxoi:
Thanks for the fix — the direction is right (the old `max` formula genuinely starved the text budget), but after digging into it we're going to close this PR in favor of a broader rework. Two findings from the investigation:

**1. The 400 narrative doesn't hold on current dependencies.** With `@ai-sdk/anthropic` (v2 line), when thinking is enabled the provider adds `budgetTokens` back into the wire `max_tokens` (`max_tokens = maxOutputTokens + thinkingBudget`, then clamped to the model's output cap). A mocked-fetch probe of the *old* formula on an 8k-output model produces `max_tokens=8192, budget_tokens=8191` — a valid request. The real symptom is different: `streamText`'s `maxOutputTokens` is the **text-only** budget, so `min(31_999, output - 1)` left exactly **1 token of text** — responses truncated after a single visible token (`stop_reason: max_tokens`), with thinking consuming the rest. The change is still the correct fix for that real bug, but the comments and test framing should describe text starvation, not a 400 that can't occur through this path. (The `budget < max_tokens` invariant is structurally guaranteed by the SDK add-back as long as text ≥ 1.)

**2. The API contract has split into two generations.** Per current official docs: `thinking: {type: "enabled", budget_tokens}` (≥ 1024, < max_tokens) is supported on 4.5-era and earlier models, deprecated-but-working on 4.6, and **rejected with a 400 on 4.7+**, which require `thinking: {type: "adaptive"}` + `output_config.effort` (low/medium/high/xhigh/max). Independently, the currently pinned `@ai-sdk/anthropic@2.0.56` zod-rejects `adaptive` and `effort: max|xhigh` outright, so the existing 4.6/4.7 variant branches fail at runtime today; a patch-level bump within the 2.x line (e.g. 2.0.98) unlocks them.

Given that, we're opening a rework PR covering: the SDK bump, generation-based capability gating (adaptive vs budget), an envelope-aware budget split (`high = clamp(min(16_000, ⌊envelope/2 − 1⌋), ≥ 1024)`, `max = min(31_999, envelope − 1024)`, and no thinking variants at all when the envelope can't fit a legal budget), plus wire-level regression tests. Your `high`-tier formula survives as the product default.

For the record: the old `max = min(31_999, output - 1)` shape also exists upstream in opencode's transform with the same text-starvation issue — worth a look there too.

Appreciate the report and the fix — the envelope math in it was a useful forcing function for the rework.

PR #1256, opened 2026-08-24
fix(plugin): use shared atomic writer for marketplace registry

TTAWDTT:
## What does this PR do?

Switches the marketplace registry's `writeJsonFile` from a hand-rolled fixed-`.tmp` rename to the shared `Storage.writeJsonAtomic` writer, which uses a unique temp name per invocation (pid + time + random) and retries on transient IO.

## Why?

`PluginMarketplaceRegistry.writeJsonFile` staged its temp file at a fixed path `${filepath}.tmp`. Two concurrent writes to the same registry file therefore share that temp path: one writer's content is clobbered before the rename lands, and the two `rename` calls race (one can hit `ENOENT` on a temp the other already moved). Under concurrent registry refreshes this surfaces as a torn/empty registry file or a failed refresh.

The shared `Storage.writeJsonAtomic` already solves this (unique temp names + retry); this PR exports it and delegates to it instead of duplicating the logic.

## How was it tested?

- New `packages/synergy/test/storage/storage-atomic-write.test.ts`:
  - 8 concurrent writers racing on the same target all land without clobbering or leaving stale temp files; the final on-disk content is exactly one of the written payloads (never torn/empty).
  - Spies on `Bun.write` to assert each invocation stages a *distinct* temp name (the fixed-`.tmp` regression would surface the same `${target}.tmp` for every call).
- `bun test test/storage/storage-atomic-write.test.ts` → 2 pass, 0 fail.
- `tsgo --noEmit` (packages/synergy typecheck) clean.

```bash
cd packages/synergy
bun test test/storage/storage-atomic-write.test.ts
bun run typecheck
```

## Checklist

- [x] `bun run quality:quick` passes — the touched package's gates (format, lint, typecheck, tests) pass for this change. Note: on a Windows checkout the repo-wide `quality:quick` has pre-existing red gates unrelated to this PR (a tracked symlink `packages/app/src/custom-elements.d.ts` parses as broken TS on Windows; plus `gitleaks`/`actionlint` not installed locally). These do not occur on the Linux CI runners; the files this PR touches are clean under `prettier`, `oxlint`, and `tsgo`.
- [x] New or moved tests live under the owning package's `test/` directory — new test at `packages/synergy/test/storage/storage-atomic-write.test.ts`.
- [x] `bun run package:check` passes if package exports, build output, release logic, or publishable packages changed — `Storage.writeJsonAtomic` is newly exported, but it is an internal runtime helper within `packages/synergy` (not a published package export / build output), so package publishing validation is unaffected.
- [x] Browser capability or App bootstrap changes pass the browser crypto contract and private HTTP browser smoke — N/A, plugin/storage layer only.


synergy-agent (dismissed):
## Review: correct and well-scoped — approve

Verified against base `dev` (diff `21ca561a9..7dc6f9038`) and the head implementations of `Storage.writeJsonAtomic`, `isRetryableIOError`, and `PluginMarketplaceRegistry.writeJsonFile`.

**Atomic write semantics — preserved.** `writeJsonAtomic` stages the temp file in the target's own directory (same volume) and renames over the target, so the rename stays atomic on POSIX and Windows. The fixed-`.tmp` collision the PR fixes is real: two concurrent writers sharing `${filepath}.tmp` clobber each other's content before rename and race on `rename` (ENOENT). The per-invocation unique name (`pid + time + random`) removes that race.

**Concurrency — safe for this use case.** `writeJsonFile` remains lock-free (unlike `Storage.write`/`update`, which take `Lock.write`), but unique temp names + atomic rename mean concurrent writers each land a complete file and the last rename wins — never a torn file. That matches the new test's final-content assertion and is acceptable for a registry cache guarded by `isFresh` TTL. pid+time+random naming also covers cross-process writers.

**Error handling — improved.** The shared writer retries EPERM/EACCES/EBUSY (4 attempts, 50→200 ms backoff) and removes the temp file on terminal failure (with retried unlink), whereas the old path had no retry and could leave a stale `.tmp` behind. Non-retryable errors still propagate, so `remoteRegistry`/`getOfficialEntry` fail-closed behavior is unchanged.

**Consistency — good.** The new export is an internal helper only (no published-package contract change). `isTempFile` already matches both the old `.tmp` suffix and the new `.tmp-*` names, so any legacy leftover temps remain filtered from scans. The test lives under `test/storage/` with the owning package, and the `using` + `spyOn(Bun, "write")` pattern follows the existing `storage-retry.test.ts` convention. CI is green on all checks.

Non-blocking observations (no action required):
- The retry loop reuses the same temp path across retries; safe because `Bun.write` truncates, and naming stays unique per invocation.
- Last-writer-wins without serialization means a slow stale refresh can overwrite a newer one; fine for this cache today, but worth revisiting if the registry cache ever becomes authoritative state.

yzxoi:
Thanks for this — solid, well-scoped fix and a clear write-up. Delegating to the shared `writeJsonAtomic` choke point is exactly the right layering, and the concurrent-writer + unique-temp-name tests are a great addition.

One note from review: the same fixed-`.tmp` pattern still lives in two sibling plugin stores — `consent/approval-store.ts` and `incompatible-store.ts` — and `plugin-approvals.json` is authoritative state (not a cache) with an unsynchronized read-modify-write on top. I'll push a follow-up commit to this branch migrating both stores to the same shared writer (plus read-modify-write serialization for the approval store).

yzxoi:
Follow-up pushed in ca21ea4b2, as promised:

- `consent/approval-store.ts`: read-modify-write (`saveApproval`/`removeApproval`/`writeApprovals`) now holds `Lock.write` across read→transform→write, so concurrent mutations serialize in-process and can't drop records; persistence delegates to `Storage.writeJsonAtomic` (unique temp names + transient-IO retry). `writeAll` stays lock-free since `Lock` is non-reentrant.
- `incompatible-store.ts`: same fixed-`.tmp` staging replaced with `Storage.writeJsonAtomic`.
- New behavioral tests: concurrent `saveApproval` + `removeApproval` keep every record with no temp residue (fails on pre-fix code with the exact `ENOENT`/lost-update symptoms), plus a concurrent-write case for the incompatible store.
- Decision record: `docs/decisions/implemented/bug-fix/2026-08-27-plugin-store-atomic-write-serialization.md`.

Verified: new tests 4 pass; consumer regressions across installation-transaction / migration / loader / install-lifecycle / permission-diff / plugin-registry-routes / install-discovery 64 pass; `tsgo --noEmit`, prettier, oxlint clean.

yzxoi:
@synergy-agent  @codex review

PR #1257, opened 2026-08-24
fix(util): age out stale file locks on Windows to defeat PID recycling

TTAWDTT:
## What does this PR do?

Ages out stale file locks past `staleMetadataMs` on Windows even when the owner PID still appears live, so a lock whose owner PID has been recycled (and now belongs to an unrelated live process) can be reclaimed instead of spinning every later acquirer to the timeout.

## Why?

`acquireFileLock` reclaims an existing lock when its owner is provably gone (`processExists(pid)` is false). On Windows, PIDs recycle aggressively and `process.kill(pid, 0)` returns `true` for a recycled PID that now belongs to an unrelated live process. So a stale lock whose owner died — but whose PID was since handed to another process — looks live forever and is never reclaimed: every later acquirer spins until `timeoutMs`, and the lock is effectively dead.

This PR consolidates reclaim logic into `isLockReclaimable` and adds a wall-clock age tiebreaker on Windows: a lock older than `staleMetadataMs` is treated as stale even if its PID looks live, using `acquiredAt` from the lock metadata (preferred over `mtime`, which antivirus/indexers can touch) with an `mtime` fallback. On POSIX the PID check is reliable, so a genuinely-live owner is never displaced by age alone.

The owner record now also persists `acquiredAt` (read best-effort; old lock files without it fall back to `mtime`).

## How was it tested?

New tests in `packages/util/test/fs-lock.test.ts`:

- `test.skipIf(!isWindows)`: reclaims a stale lock whose recycled PID looks live — stages a lock with `pid = process.pid` (so `processExists` is true, simulating a recycled PID) and `acquiredAt` 10s in the past (beyond a 5s grace); `withFileLock` acquires it.
- `does not reclaim a fresh live-owner lock by age alone`: a fresh lock (live PID, current `acquiredAt`) within the grace period times out and is left intact — proving the age gate doesn't displace a genuine live owner.

`bun test test/fs-lock.test.ts` → 6 pass, 0 fail. `tsc --noEmit` (packages/util typecheck) clean.

```bash
cd packages/util
bun test test/fs-lock.test.ts
bun run typecheck
```

## Checklist

- [x] `bun run quality:quick` passes — the touched package's gates (format, lint, typecheck, tests) pass for this change. Note: on a Windows checkout the repo-wide `quality:quick` has pre-existing red gates unrelated to this PR (a tracked symlink `packages/app/src/custom-elements.d.ts` parses as broken TS on Windows; plus `gitleaks`/`actionlint` not installed locally). These do not occur on the Linux CI runners; the files this PR touches are clean under `prettier`, `oxlint`, and `tsc`.
- [x] New or moved tests live under the owning package's `test/` directory — appended to the existing `packages/util/test/fs-lock.test.ts`.
- [x] `bun run package:check` passes if package exports, build output, release logic, or publishable packages changed — `packages/util` is a publishable package; this change adds no new exports (the new `isLockReclaimable`/`lockFileAgeMs` are module-private), so package publishing validation is unaffected.
- [x] Browser capability or App bootstrap changes pass the browser crypto contract and private HTTP browser smoke — N/A, util fs-lock only.

### DIFF
--- a/packages/util/src/fs-lock.ts
+++ b/packages/util/src/fs-lock.ts
@@ -66,16 +66,9 @@ async function acquireFileLock(options: FileLockOptions): Promise<{ release(): P
     } catch (error) {
       if ((error as NodeJS.ErrnoException).code !== "EEXIST") throw error
       const owner = await readLockOwner(filename)
-      if (owner?.pid && !processExists(owner.pid)) {
+      if (await isLockReclaimable(filename, owner, staleMetadataMs)) {
         await fs.unlink(filename).catch(() => {})
         continue
-      }
-      if (!owner?.pid) {
-        const stat = await fs.stat(filename).catch(() => undefined)
-        if (stat && Date.now() - stat.mtimeMs > staleMetadataMs) {
-          await fs.unlink(filename).catch(() => {})
-          continue
-        }
       }
       if (Date.now() - startedAt >= timeoutMs) {
         throw new Error(options.timeoutMessage ?? `Timed out acquiring file lock for ${options.key}`)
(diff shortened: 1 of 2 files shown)


synergy-agent (changes requested):
### Review: REQUEST_CHANGES

The Windows PID-recycling motivation is sound, and consolidating the reclaim decision into `isLockReclaimable` is a clean refactor: the no-PID (mtime) and provably-dead-PID paths behave exactly as before, the `acquiredAt` metadata is already written by the current writer (baseline line 52), and the fallback for older lock files is `mtime`. I verified the existing 4 tests pass on Linux (`bun test test/fs-lock.test.ts` → 4 pass, 0 fail); the new Windows-only test faithfully simulates a recycled PID with `process.pid`, and the fresh-live-owner test guards the age gate on every platform. No public exports change, so `packages/util` package compatibility is unaffected.

**Blocking**

1. **Age-based reclaim of a *live* owner + unguarded `release()` unlink breaks mutual exclusion on Windows (ABA race).**
   The new Windows branch reclaims a lock whose owner PID still looks live — meaning the owner is an active process that will eventually call `release()`. But `release()` is an unconditional `fs.unlink(filename)` with no ownership check:
   - A holds the lock (`pid=A`).
   - B waits, sees the lock older than `staleMetadataMs` (default 5 s) on Windows, unlinks it and creates its own lock (`pid=B`).
   - A finishes its critical section and calls `release()` → unlinks **B's** lock file.
   - C acquires, and B and C now run the critical section concurrently — exactly the lost-update/interleaving this lock exists to prevent.
   The old dead-PID path could never hit this because a dead owner never calls `release()`. Once age can displace a live owner, `release()` must verify ownership first (read the file and unlink only if the `pid`/`acquiredAt` still match what this process wrote; an inode check via `stat` before/after is a stronger variant). Please add that guard plus a regression test for the steal → release → re-acquire sequence (e.g. Windows-only: A acquires, simulate B stealing after `staleMetadataMs`, A releases, assert B's lock file survives).

**Non-blocking**

2. **`staleMetadataMs` (default 5_000) becomes a hard maximum hold time for live Windows owners.** No caller overrides it — `holos/accounts.ts` `withWriteLock`, `migration/index.ts` `mergeDomainLog`, `provider/api-key.ts` `withLock`, and `synergy-link/holos/auth.ts` all use defaults — and those critical sections do JSON read-modify-write + temp-file rename. On a slow disk or a large store they can exceed 5 s, after which a second process enters concurrently. If 5 s is intended as the Windows lease, callers should opt in explicitly (or a separate lease option with a documented contract), and the PR body should call out that `staleMetadataMs` semantics changed from "metadata-only fallback" to "active-owner lease on Windows".

3. **Test gap:** no coverage for the `mtime` fallback path on Windows (lock file without `acquiredAt`). The fallback is conservative (AV/indexer-touched mtime delays reclaim), but it is precisely the path that matters on Windows and it is untested.

4. **Wall-clock dependence (minor):** `Date.now()` epoch-ms comparison has no timezone issue — good. But a large system clock step (NTP/ manual) can prematurely age locks out on Windows or delay reclaim. Acceptable for single-machine lock files; a one-line comment acknowledging the wall-clock dependence would help future maintainers.

yzxoi:
Thanks @TTAWDTT for the diagnosis — the Windows PID-recycling stall is a real bug, and this PR is what surfaced it. The test documenting the failure mode was particularly helpful.

We're closing this PR in favor of a different fix, for one core design reason: **the wall-clock age gate redefines `staleMetadataMs` (5s default) into a maximum hold time for legitimate lock owners on Windows.** All current call sites (provider credential store, Holos accounts store, migration-log merge) use that default and hold the lock across several file operations, so a live owner that is merely slow — AV/indexers can push a single rename into multi-second territory on Windows (#1247) — gets displaced mid-critical-section. That trades today's loud 60s timeout for a silent mutual-exclusion violation on exactly the stores this lock exists to protect, which is a worse failure than the bug. Secondary issues from the review:

- Wall-clock aging breaks under clock rollback (a negative age re-introduces the permanent stall) and the mtime fallback also ages out live owners holding legacy `{pid}`-only payloads in mixed-version deployments.
- The reclaim branch's `continue` skips both the timeout check and the retry sleep, so a lock file whose unlink keeps failing (AV holding a handle) spins one core forever without ever hitting the timeout.

The approach we'll take instead already exists in-repo: `ServerProcessLock` (`packages/synergy/src/daemon/server-process-lock.ts`) records a `processStartIdentity` — the owner process's creation time from `/proc/<pid>/stat`, wmic/PowerShell `CreationDate`, or `ps -o lstart=` — and treats a live pid as recycled only when its current start identity no longer matches the recorded one. That deterministically detects recycled PIDs on **every** platform (pids recycle on posix too, just slower) and never displaces a live owner by age.

Follow-up PR (incoming) extracts that identity query into `packages/util`, uses it in `fs-lock` reclaim, and adds owner-token validation on release — a gap this PR's review also surfaced. Thanks again for the report; the fix is built on it.

PR #1259, opened 2026-08-26
refactor(agent): narrow preamble messages to key moments

EricSanchezok:
## Summary

- Rework the injected `## Preamble Messages` agent guidance so agents stop narrating every visible tool call.
- Agents now send a short message only at key moments: starting a multi-step task, key decisions, unexpected findings or risks, and completion.
- Keeps the same section heading, so the idempotent injection paths (`session/llm.ts` native turns, `session/invoke.ts` external-agent instructions) are unchanged.

## Why

The old guidance ("Before using tools, send one brief sentence...") produced constant narration of routine reads, searches, and edits on surfaces where tool calls are already visible (desktop/web), drowning out the messages that actually need user attention.

## Verification

- `bun run quality:quick` — all 15 local gates passed (format, lint, typecheck, decision:check, doc:check, deadcode, etc.)
- No tests assert the literal prompt text; no code paths changed


yzxoi (approved):
## Verdict

Approve as of 71ce27914 — the PR now contains the preamble narrowing (4f6a2720) plus a clean revert of the proxy env change (71ce27914). The remaining Test-job failures are the pre-existing dev signature, fixed separately in #1263.

## Preamble narrowing (4f6a2720): sound

- Heading kept identical, so `withPreambleSection`'s idempotent injection in `session/llm.ts` and `session/invoke.ts` is unaffected; both call sites verified, no other embedded copies of the old text exist in the repo.
- No tests assert the literal prompt text (no `test/agent/prompt/` suite, no grep hits).
- The channel-surface trade-off (fewer mid-task messages where tool calls are not visible) is acknowledged in the record's alternatives; acceptable product call.

## Proxy neutralization (f964d0c, reverted by 71ce27914): wrong diagnosis + a regression

- The proxied-runner theory does not hold: the same 12 arXiv 404 failures reproduce by co-running `test/script/script-identity.test.ts` with `test/tool/arxiv-download.test.ts` on a machine with no proxy variables set. Root cause is the module-level `globalThis.fetch` 404 stub leaking to shard siblings — fixed in #1263.
- The pinned `NO_PROXY` deterministically broke `test/provider/proxy.test.ts` ("proxy option routes wrapped fetch through configured proxy"): in Bun 1.3.14 a pre-first-fetch `NO_PROXY` overrides the explicit per-request `proxy` option. That case failed locally on this branch's head and on its CI Coverage job; after the revert it passes 9/9 locally.
- Both effects are recorded in #1263's decision record; the reverted commit's record was removed with it.

PR #1261, opened 2026-08-26
feat(app): calm prompt input hover, lighten transition card, truncate trace titles

EricSanchezok:
## Summary

Three polish fixes from UI feedback on the composer and session surfaces:

1. **Calmer prompt input hover** (`packages/app/src/index.css`) — removed the shell `:hover` background change so only focus-within frames the composer; toolbar controls keep their own discrete hover affordance.
2. **Lighten Worktree session transition card** (`packages/app/src/components/session/session-transition-card.css`) — card surface + 14px radius instead of modal-style surface + 8px; header icon 34→28px, step icons 28→22px, step rows 58→44px min-height.
3. **Truncate activity trace titles in narrow containers** (`packages/ui/src/components/activity-trace.css`) — step/receipt titles now shrink with ellipsis instead of overflowing the message column on narrow windows.

## Tests

- Added Playwright CSS contract tests: `packages/app/test/components/prompt-input/prompt-input-hover-style.test.ts`, `packages/app/test/components/session/session-transition-card-style.test.ts`, `packages/ui/test/components/activity-trace-narrow.browser.test.ts` (all red before, green after)
- Existing regressions: `packages/ui` activity-trace 24 pass, `packages/app` session components 15 pass
- `bun run quality:quick` — 15/15 gates pass (format, lint, typecheck, doc:check, decision:check, secrets, etc.)
- `bun run decision:check` — passes

## Notes

- Two decision records added: `implemented/feature/2026-08-27-calmer-input-adjacent-feedback.md`, `implemented/bug-fix/2026-08-27-activity-trace-narrow-overflow.md`
- No theme tokens or public API changes


yzxoi:
Thanks for the polish pass — the direction on all three changes reads well, and the flex truncation fix itself (`flex: 0 1 auto` + `min-width: 0` + ellipsis) is the right approach. Verified locally on this branch: the three new suites pass (6/4/2), and the surrounding activity-trace regressions (73 tests) stay green.

One note up front: the `Test` check failure is not introduced by this PR — dev at the same base commit (`4de7a53b8`, run 32633715210) fails the same 12 `arXiv PDF download` tests (ECONNREFUSED from CI network policy) plus the `ObservabilityStore` cap test (timing-sensitive). Might be worth a line in the PR body so reviewers don't misattribute it.

### Should fix before merge

**1. Truncated titles have no recovery path — the decision record's reachability claim doesn't hold** (`docs/decisions/implemented/bug-fix/2026-08-27-activity-trace-narrow-overflow.md`)

The DR says _"the full title remains reachable through the existing expand/Collapsible behavior and the step subtitle"_, but none of those paths actually surface the full title:

- Neither title span (`packages/ui/src/components/activity-trace.tsx:235`, `:337`) has a `title` attribute or a `Tooltip`.
- Expanding a step renders `ToolResultBody resultOnly`, which skips the trigger/title entirely (`packages/ui/src/components/basic-tool.tsx:163` returns only `props.children` under the result-only context).
- The motivating case is the `delegate` family ("Call subagent …"), but `specializedActivityDetail` only covers `modify-files` and `coordination/dag`, so a delegate receipt is not expandable at all — it renders through the fallback branch (`activity-trace.tsx:370`). The subtitle is also optional.

Net effect: the narrower the window, the more of a delegated task's title is hidden, with no way to recover it. Suggested fix: a native `title` attribute on both spans (one line each) makes the DR claim true without restructuring the trigger buttons or paying a Tooltip portal per row.

**2. New Playwright suites aren't registered in the serial batch**

`packages/app/script/test.ts` documents the contract in its own comment: _"Playwright suites launch Chromium. bun test runs files in parallel worker processes and reaps dangling children when a worker exits… Run every Chromium suite serially after the main batch."_ `prompt-input-hover-style.test.ts` and `session-transition-card-style.test.ts` launch Chromium but are missing from `playwrightIsolated` (the direct precedent `question-prompt-style.test.ts` is registered). Same for `activity-trace-narrow.browser.test.ts` in `packages/ui/script/test.ts` — it follows `activity-trace-layout.browser.test.ts`, which is itself unregistered (pre-existing gap worth fixing while here). CI passed this run, but this is exactly the flake pattern the runner comment warns about, and it also adds two browser instances to the parallel batch.

### Worth addressing

**3. Opacity-guaranteed token replaced with an alpha-permitting one** — the old `var(--surface-raised-stronger-non-alpha)` guaranteed an opaque surface (the same reason popovers use that token, `index.css:159`); the new `var(--workbench-card-bg, var(--surface-raised-base))` lets custom themes feed alpha through, and this card sits over message content. It does match the `session-inbox.css:177` precedent, so the card language alignment is defensible — but the opacity-semantics change should be recorded in the decision record either way.

**4. Dead transition + intentional behavior split undocumented** — after removing the `:hover` rule, `background-color 150ms ease` in `.prompt-input-shell`'s transition list is dead code (focus changes border/shadow; dragging changes border; nothing changes the shell background anymore). Also worth a DR note: `.workbench-input-surface` (`index.css:225`, used by e.g. `activity-view.tsx:68`) still brightens on hover, so the two input surfaces now intentionally diverge — otherwise someone will "fix" the inconsistency back.

**5. Subtitle shrink behavior changed silently** — previously the title was `0 0 auto` (never shrank; the subtitle absorbed all the shrink); now both shrink proportionally, so subtitles lose visible width they used to have. Likely the better balance, but the DR and tests only cover the title.

### Nits

- Both narrow tests compute `traceLeft` in the evaluate payload but never assert on it.
- `prompt-input-hover-style.test.ts` injects all of `index.css`, but its leading `@import` appears after `:root{}` and is silently ignored per CSS spec — worth a comment so the test doesn't read as a full-style-contract harness.
- `.session-transition-step-icon`: `width` was moved above `display`, opposite ordering from every other block in the file (pure churn).

PR #1266, opened 2026-08-27
fix(provider): gate Anthropic thinking by generation with envelope-safe budgets

yzxoi:
## Summary

Anthropic extended thinking split into incompatible generations, and the runtime's single budget formula broke on both sides of the split:

- The pinned `@ai-sdk/anthropic@2.0.56` zod-rejected `thinking: {type:"adaptive"}` and `effort: "max"|"xhigh"`, so the existing 4.6/4.7 variant branches failed at request time.
- The budget fallback (`max = min(31_999, output − 1)`) left exactly **1 token** of visible-answer room on 8k/16k/32k-output models — the SDK adds `budgetTokens` back into wire `max_tokens`, so requests were legal but `streamText`'s text-only budget was 1, truncating answers after one token.
- Official docs confirm `enabled + budget_tokens` is **rejected with HTTP 400 on 4.7+** (adaptive + effort required), deprecated-but-working on 4.6, and the only mode for ≤4.5.

This PR fixes the whole surface:

1. **SDK bump** `@ai-sdk/anthropic` `2.0.56 → 2.0.98` (same 2.x line, `LanguageModelV2`, compatible with pinned `ai@5`) — backfills `adaptive` thinking and the full `effort` enum.
2. **Generation gating** in `ProviderTransform.variants()`: 4.7+/5.x → adaptive + `display: "summarized"` + effort (5 levels); 4.6 → adaptive + effort; 3.x–4.5 → budget variants from the effective output cap.
3. **Envelope-safe budgets**: `high = clamp(min(16_000, ⌊cap/2−1⌋), ≥1024)`, `max = min(31_999, cap − 1024)` — `max` is distinct from `high` again and always leaves ≥1024 text tokens; envelopes < 2048 emit no thinking variants; shared-context corners are clamped with the same headroom as `maxOutputTokens`.
4. **Catalog efforts narrowed**: `reasoningEfforts` trusted only for unknown/unversioned ids and Opus 4.5 (the sole extended-only model that also supports effort).
5. **Wire-level regression tests** (mock-fetch through the locked SDK): budget legality across `envelope ∈ {2047…128000}`, generation purity (4.7+ never `budgetTokens`, ≤4.5 never `adaptive`), adaptive variants pass the SDK validator with correct `output_config.effort`.

## Test

- `bun test test/provider/transform.test.ts` — 126 pass / 0 fail
- `tsgo --noEmit` — clean
- `bun run doc:check`, `bun run decision:check` — pass

Closes #1255 (superseded by this rework).


yzxoi:
@synergy-agent review

PR #1271, opened 2026-08-27
feat(config): add synergy config export with secret redaction

SII-jake:
## Summary

- **New CLI**: `synergy config export [--include-secrets] [--only <domain>...] [--scope global|project] [--output file]`. Exits 1 on errors. Secrets are replaced with `__REDACTED__` sentinel by default — the redacted output is a *valid import payload*: `synergy config import` merges sentinels back with the machine's stored credentials, so same-machine round-trips restore secrets while cross-machine imports keep them as placeholders (machine-independent, no `$schema` — the runtime's only schema URL is the install-local `file://` path, a broken link elsewhere).
- **New API**: `GET /config/export` (`operationId: config.export`) returning `{scope, scopeID, secretsIncluded, domains[], config}`. No `meta({ref})` on the query schema — hono-openapi registers ref'd query objects as single-parameter components and silently drops all but the first property (pre-existing bug affecting `PerformanceSummaryQuery` too; visible and fixable in a separate PR). `includeSecrets` uses `z.stringbool(...)` because `z.coerce.boolean()` maps `"false"` → `true` — HTTP probes confirmed `includeSecrets=1` returns 400.
- **CLI reference fix** (`script/gen/gen-cli-reference.ts`): the top-level table resolved row descriptions through one global name-keyed map of command blocks, so a nested subcommand sharing a top-level name (the new `config export` colliding with the session-export root) overwrote the row (`export` showed "Export config as JSONC...", `config` and `doctor` rows were also wrong). Rows now resolve from the command's own module; detail sections keep the merged view.

## Test plan

- `bun test test/config/export.test.ts` (6 pass; isolation test with tmp Scope covers redaction by default, plaintext with `includeSecrets: true`, import round-trip with sentinel → stored-secret merge, and `--only` domain selection)
- `bun test test/server/config-export-route.test.ts` (5 pass; real HTTP probes over Hono app mounted on `/config/export`)
- `bun test` (503 total, 1 known flaky parallel-race failure in `test/config/degraded.test.ts` that also fails on baseline `dev` — verified by running the full suite on a clean stash before and after my change; passes when run alone)
- `bun run packages/synergy/src/index.ts config export` — CLI E2E against isolated `SYNERGY_HOME` with seeded inline keys (validates default redaction, `--include-secrets`, `--only`, `--output` + 0600, and warning)
- `bun run typecheck` (11 packages) and `bun run quality:quick` (all gates pass except `secrets:check` which requires `gitleaks` — environment gap, not related to this change; CI runs the equivalent job)

## Notes for reviewers

- The export deliberately reads **domain files at the target scope** (not the merged runtime `Config.current()`), so it captures exactly what the user declared vs. env-expanded/merged layer — more truthful to the "config" concept of import/export.
- Plaintext exports are chmod'd `0600` when written via `--output` and the command prints a warning; docs mark the output as sensitive. A redacted export never carries machine-resolvable credentials (sentinels merge back with stored secrets, they don't travel).
- Cross-machine note: `provider-auth.json` (OAuth/device-code token store and credential pool) is **not** covered by this feature — it's runtime state, not declarative config. Importing auth state is a scope too far for this change; no regression there.

🤖 Generated with [Synergy](https://github.com/SII-Holos/synergy)


yzxoi:
Reviewed. The direction and the sentinel round-trip design are good, but **request changes** — the "secrets redacted by default" promise doesn't fully hold yet. Findings (repro'd locally on this branch):

- **D1 (blocker)** Default redaction misses real secret carriers: MCP remote `headers` (`Authorization: Bearer …`), MCP local `environment` (`*_API_KEY`), agent `options` all export in plaintext by default (and stdout exports get neither 0600 nor a warning).
- **D2** `GET /config/export?includeSecrets=true` is the first HTTP endpoint returning plaintext secrets; with loopback-wide CORS and no token check, any localhost page can read them. Proposal: HTTP stays redacted always; plaintext only via CLI `--include-secrets`.
- **D3** Export resolves `{env:}`/`{file:}` references at read time, so a redacted re-import that changes anything else in the domain rewrites the stored reference as resolved plaintext on disk. Needs documented semantics at minimum.
- **D4** Export is not read-only: a domain file with a syntax error gets quarantined (renamed) by the read path and the domain silently disappears from the export. Should warn instead.
- **D5** `./relative` plugin specs export as install-local absolute `file://` URLs, contradicting the machine-independent payload goal (the `$schema` removal handled exactly this class of problem).
- Minor: no warning for `--include-secrets` on stdout (docs' first example is a redirect); chmod failure swallowed; dead `meta({ref})` on `ConfigExport.Query`; no read lock during export; tests don't cover the surfaces above.

Fixes incoming on this branch: expanded redaction (MCP headers/environment + agent options, secret-shaped keys) with symmetric sentinel merge, HTTP redacted-only (explicit 400 for `includeSecrets=true`), export warnings incl. non-quarantining reads, plugin spec relativization, docs for `{env:}`/plugin portability, plus the minor items.

yzxoi:
Follow-up fixes pushed in e9a59f5bb, addressing the review findings:

- **D1** `redactForClient`/`mergeRedactedSecrets` now cover MCP remote `headers`, MCP local `environment`, and agent/provider/model `options`. Secret-shaped keys match by split components (`token`, `password`, `authorization`, `apiKey`, `credential`, `secret`, `privateKey`, `cookie`) so `ANTHROPIC_API_KEY`/`GITHUB_TOKEN`/`Authorization` are caught while `max_tokens`/`NODE_ENV` are not; sentinel merge is symmetric so redacted round-trips restore these values too (tested end-to-end through `ConfigImport.apply`).
- **D2** `GET /config/export` is always redacted: `includeSecrets=true` → `400 ConfigExportSecretsRejectedError` pointing at the CLI. Plaintext export is CLI-only.
- **D3** `{env:}`/`{file:}` read-time resolution semantics (and the write-back-as-literal consequence on re-import) documented in `configuration-layout.md` + cli-guide.
- **D4** Export is read-only: new `Config.domainReadForExport` reads under the domain read lock without quarantining; broken files become `warnings` in the API result and stderr warnings in the CLI (tested: file stays on disk untouched).
- **D5** Plugin specs are re-relativized to the config directory (`./dev-plugins/foo` round-trips instead of an absolute `file://` URL); cross-machine plugin-path caveat documented.
- Minor: plaintext warning now also on stdout path, chmod failure warns instead of being swallowed, dead `meta({ref})` on `ConfigExport.Query` removed, read lock added, docs/decision record/SDK/OpenAPI regenerated.

Validation: `bun test test/config/ test/server/config-export-route.test.ts` → 174 pass; `bun run typecheck` (11 pkgs), `decision:check`, `doc:check`, `gen-cli-reference --check` all green; `quality:quick` 14/15 (only `secrets:check` fails locally — missing `gitleaks`, covered by CI).

yzxoi (approved):

PR #1274, opened 2026-08-28
perf(server): serve global session search from nav indexes

yzxoi:
# PR Body: perf(server): serve global session search from nav indexes

## Summary

`GET /global/session` (`global.session.search` — backend for the Web global search modal (Cmd/Ctrl+K) and the archived-sessions panel) previously read one info JSON per candidate session on every request, serially, then filtered by title in memory — every keystroke (after the 250 ms debounce) cost O(all sessions) file reads, almost all discarded by the limit slice.

The route now runs entirely on the `session_nav_v2` indexes: one parallel index read per scope, archived/parentOnly/title filtering and all sorting on the in-memory entries, and a batched info read only for the returned page (≤ limit) to supply `lastExchange` previews. The response schema is byte-identical; the frontend needs no changes.

- `SessionNavEntry` gains optional `createdAt` / `archivedAt`, filled on every mutation path (`toNavEntry`) and on rebuild (`buildNavIndexUnlocked`); migration `20260828-session-nav-timestamps` rebuilds all nav indexes to backfill them, and the route lazily backfills from info files when it meets a legacy index
- When `search` is present and no explicit `sortBy` is passed, results rank prefix match > word-boundary match > substring match (recency as tiebreaker); explicit sorts behave exactly as before
- SDK types regenerated via `bun ./script/generate.ts` (the new `SessionNavEntry` fields enter the public contract)

## Performance

Same 400-session fixture, isolated test home, 10-request average: new route **1.83 ms/request** vs the previous access pattern (page index + one serial info read per session + in-memory filter) **20.57 ms/request** — **11.2x**, identical match counts. Read amplification drops from O(N) to O(scopes + limit).

## Tests

New behavior tests (written failing first):

- `packages/synergy/test/session/nav-index.test.ts`: nav entries carry createdAt/archivedAt; archivedAt clears on restore
- `packages/synergy/test/server/global-session-search.test.ts`: `sortBy=created` in both directions; `lastExchange` preserved; exact response values when the nav index predates the new fields; relevance ranking plus explicit-`sortBy` immunity

All pre-existing tests stay green (cross-scope aggregation, search filtering, parentOnly, archived:only, restore, sort contracts).

Verification run:

- `bun test test/server test/session` from `packages/synergy`: 1425 pass / 0 fail (135 files)
- `bun test test/tool/session-search.test.ts`: 9 pass
- `bun run typecheck` (repo root): 11/11 packages clean
- `bun run quality:quick` (repo root): 13/15 gates pass; the 2 failures are local-environment only (gitleaks not installed, actionlint download blocked by network policy) and are owned by CI jobs
- `bun run decision:check`: passed

## Decision record

`docs/decisions/implemented/architecture/2026-08-28-global-session-search-nav-index.md` (alternatives considered: extending the page index, batched-only reads, SQLite FTS)

## Follow-up (not this PR)

The `Session.list` search branch (SDK/CLI surface) has the same full-scan shape and can get the same treatment in a small follow-up.


yzxoi:
Review feedback addressed in 9ca078994:

**1. Timestamp migration coverage (P1)** — Added a `session nav timestamps migration` suite in `packages/synergy/test/session/migration.test.ts`: it seeds a legacy nav index (timestamp fields stripped) plus authoritative info with deterministic timestamps, runs `20260828-session-nav-timestamps.up()` twice for idempotence, asserts the backfilled created/updated/archived values, and asserts the migration is registered after the channel-provider-fields migration.

**2. Authoritative updated timestamp (P2)** — `SessionNavEntry` gains `updatedAt`, filled on every write path (`toNavEntry` + `buildNavIndexUnlocked`) and mirroring `time.updated`. It is unaffected by the `preserveActivityAt` freeze that `upsertNavEntry` applies to `lastActivityAt` while a session stays pendingReply. The route now derives `sortBy=updated` and the response's `time.updated` from `entry.updatedAt` (with lazy backfill for legacy indexes), so ordering and reported timestamps keep their pre-change meaning; `lastActivityAt` remains the sidebar navigation key.

**3. Deleted sessions / ghost entries (P2)** — `removeInternal` now drops the nav entry *before* deleting the authoritative info files, so an interrupted delete leaves a live session rather than a ghost index entry; and the route skips any nav entry whose info file no longer exists, restoring the old route's behavior of excluding orphans from global search and the archived panel.

Verification: `bun test test/server/global-session-search.test.ts test/session/nav-index.test.ts test/session/migration.test.ts` → 53 pass / 0 fail (includes new pendingReply-freeze, ghost-exclusion, and migration upgrade-path tests); `bun run typecheck` 11/11 packages; `bun run decision:check` passed; SDK types regenerated.

PR #1275, opened 2026-08-28
fix(ui): render completed media-generation deliveries as inline galleries

yzxoi:
# PR Body: media-generation deliveries render inline

## Summary

- Completed media-generation tool results (image/video/audio tools) render as a bare `AttachmentGallery` inline in the conversation flow again — no tool-card header, no chevron, identical chrome to user-attachment galleries.
- Attach deliveries keep the titled `BasicTool defaultOpen` collapsible (filename/count subtitle + size tag); only the completed-media presentation is split off, by display policy, in the renderer (`TimelineItemDisplay` in `packages/ui/src/components/session-turn.tsx`). Timeline kinds, projection, stable keys, spacing, and activity passthrough are untouched.
- Root cause: the collapsible-card follow-up to #1249 wrapped the shared `tool-attachments` path in a `BasicTool` collapsible; media-generation shared that path and was collateral damage. The superseded decision is archived and sealed (`archived/bug-fix/2026-08-22-tool-attachments-collapsible-card.md`); the new record is `implemented/bug-fix/2026-08-28-media-generation-inline-gallery.md`.
- Archiving the record exposed a gate conflict — the record format mandates `Archived:` directly below `Status:`, while `doc:check` enforces one physical line per paragraph, and sealed archives can never be reflowed. `script/doc-check.ts` now classifies those two exact metadata shapes as structural lines; prose lines merely starting with `Status:` remain reportable (negative case pinned in `test/script/doc-check.test.ts`).

## Verification

- `bun test test/components/session-turn-attachments-collapse.dom.test.ts` (packages/ui) — attach collapsible + media bare-gallery assertions, red→green
- `bun test test/components/session-turn-timeline.test.ts test/components/session-turn-projection.test.ts test/components/session-turn-activity.test.ts` — 105 pass
- `bun run --cwd packages/ui test` + `bun run --cwd packages/ui typecheck` — pass
- `bun test --config /dev/null test/script/doc-check.test.ts` — 13 pass; `bun run doc:check`, `bun run decision:check`, prettier, oxlint — pass


yzxoi:
Valid catch — addressed in 6a8e60971.

The first cut exempted both `Status:` and `Archived:` shapes, which let wrapped prose ending in an exact `Status: implemented` line escape the gate (and postmortems already use that exact form).

Narrowed to the archive seal only: `DECISION_META` now matches just `Archived: YYYY-MM-DD`. Since wrap violations and reflow merging key off the *next* line being non-special prose, exempting the seal alone shields the `Status:` line directly above it in archived records — no adjacency state or path scoping needed. Every other `Status:` line (postmortems, ordinary records, wrapped prose) keeps full one-physical-line enforcement.

Pinned in both directions: `test/script/doc-check.test.ts` now asserts the archived shape passes, and `["wrapped prose line", "Status: implemented"]` reports line 1. Decision record updated to document the seal-only rule and the rejected first cut.

PR #1276, opened 2026-08-28
fix(session): sanitize non-JSON-safe provider metadata in prompt projection

EricSanchezok:
## Summary

Model prompts could contain non-JSON-safe values in provider-controlled metadata, causing `AI_InvalidPromptError: The messages must be a ModelMessage[]` at AI SDK prompt validation. The reported instance came from `@openrouter/ai-sdk-provider`'s streaming accumulation, which attaches `reasoning_details` arrays containing `signature: undefined` (an own key with an undefined value) to tool-call `providerMetadata`; those values survive in memory and across v8 IPC, and AI SDK's zod `providerOptions` schema rejects `undefined`. Disk persistence hides this because JSON serialization drops undefined keys and coerces `NaN`/`Infinity` to `null` — so only the in-memory prompt path ever failed, only on providers whose metadata carried such values, and only from a turn's second step onward.

`MessageV2.projectModelMessages` now sanitizes the payload fields it derives from untrusted in-memory state before they enter the prompt:

- New `sanitizePromptPayload` helper performs a JSON round-trip with a `BigInt` → string replacer (undefined-valued keys are dropped, `NaN`/`Infinity` become `null`, arrays lose undefined entries to `null`) and returns `undefined` if serialization throws (e.g. circular references).
- Applied at the projection exits that copy provider-controlled values: `modelProviderMetadata()` (text/reasoning `providerMetadata`, tool-call/tool-result `providerOptions`) and tool `state.input` / `state.output`.

The sanitizer targets payload fields before `convertToModelMessages` rather than the returned message graph: the AI SDK converter itself adds schema-legal undefined-valued structural keys (e.g. `providerExecuted`) that existing tests pin; only projection-owned payloads need the JSON-safety guarantee. Clean metadata round-trips idempotently, so well-behaved providers are unaffected beyond the conversion cost.

Decision record: `docs/decisions/implemented/bug-fix/2026-08-28-sanitize-provider-metadata-model-prompt.md`.

## Test plan

- New behavioral tests in `packages/synergy/test/session/message-v2.test.ts`:
- Exact openrouter pollution shape (`signature: undefined` inside `reasoning_details`) no longer appears in any projected `providerOptions`; fails before the fix, passes after.
- Semantics pins: `NaN`/`Infinity` → `null`, `BigInt` → string, undefined key → dropped, undefined array entry → `null`.
- Clean metadata is unchanged; absent metadata stays absent.
- Circular metadata is stripped instead of throwing.
- `bun test test/session/message-v2.test.ts` — 30 pass.
- `bun run test:changed`, `bun run quality:quick` (15 gates), `bun run decision:check` — all green.

The same upstream defect will be reported to `OpenRouterTeam/ai-sdk-provider` (suggested fix: avoid assigning `signature` when both sides are absent, or strip undefined-valued keys from accumulated details).


yzxoi (approved):
## Approve ✅

Excellent root-cause work on this one. The transport-asymmetry diagnosis — in-memory + v8 IPC preserves `undefined`-valued keys while disk/HTTP JSON serialization silently drops them — explains every symptom (second-step-only, provider-specific, restart-heals, invisible to disk forensics), and reproducing it end-to-end against the real provider package made the fix airtight. Thanks @EricSanchezok!

Reviewed with runtime evidence; everything checks out:

- ✅ Root cause chain confirmed against `@openrouter/ai-sdk-provider@1.5.2` source (`lastDetail.signature = lastDetail.signature || detail.signature` creating an own `signature: undefined` key) and `ai@5.0.212`'s zod prompt validation
- ✅ Sanitizing at the projection boundary is the right owner-level call — the decision record's alternatives analysis (ingest-time, post-conversion, upstream-only, pin/upgrade) is thorough and each rejection is well-reasoned
- ✅ The follow-up commit (ff9ed239e) closes the one real gap found in review: provenance stringified the raw tool input, so BigInt/circular inputs could still throw from inside the projection itself — prompt fields and provenance now share one sanitized product. It also adds a string fast-path for tool output (schema-pinned `z.string()`, no round-trip needed) and surfaces `sanitization` counters so future provider regressions into this failure class show up in logs instead of silently altered prompts
- ✅ Verified: 32 tests in `message-v2.test.ts` (incl. new regression cases for non-JSON-safe tool input and counter semantics), 125 downstream tests (compaction / summary / title / invoke-compaction / image-context), typecheck, and `decision:check` all green

Nice catch on also filing the upstream issue — hopefully the `??` fix lands there soon. Merging from my side.

PR #566, opened 2026-07-12
feat(plugin): replace the plugin platform with Plugin API 3

EricSanchezok:
## Overview

This branch replaces the development-era plugin stack with one Plugin API 3 architecture and then extends that architecture with the general capabilities required by complex plugins such as Truthward. It is intentionally breaking for plugin authors: there is one descriptor, one generated manifest, one runtime protocol, one dispatcher, one UI host, and one capability model. No legacy plugin adapter or dual-read path remains.

## Plugin contract and build pipeline

- Makes `definePlugin()` the only author-written plugin definition.
- Uses one flat discriminated `contributions` collection for operations, events, tools, hooks, agents, skills, MCP servers, auth providers, UI surfaces, settings, themes, icons, and lifecycle handlers.
- Generates `plugin.json` from the descriptor instead of maintaining duplicate manifests, handler maps, permissions, schemas, or dependency declarations.
- Validates unique contribution IDs, capability requirements, executable handler parity, top-level object tool schemas, artifacts, protocol versions, generations, and integrity hashes.
- Rewrites plugin-kit build, validate, pack, sign, create, and dev generation handling around the generated contract.
- Bundles dependencies during build and removes install-time `Bun.install()`.
- Produces deterministic archives with stable top-level members so signing and validation behave identically across Windows and Linux.
- Compiles trusted Solid UI components as dedicated artifacts.

## Runtime and dispatch

- Introduces the process-wide `PluginRuntimeRegistry`, keyed by plugin ID, version, and generation.
- Removes Loader-side initialization and the previous double-init/runtime duplication.
- Shares one active runtime generation across enabled Scopes while constructing a fresh invocation context for every call.
- Routes tools, operations, hooks, lifecycle handlers, and task callbacks through the same generation-aware runtime manager.
- Adds structured cancellation, timeouts, output validation, health reporting, crash isolation, and stale-generation rejection.
- Keeps external plugin process isolation as a reliability boundary, without claiming an OS security sandbox.
- Removes the overlapping worker path and obsolete interaction route.

## Capabilities and host services

- Replaces raw SDK client/token/server access with capability-gated host services.
- Centralizes capability-to-service mapping for Session, Workspace, Settings, Secrets, Events, Tools, and delegated Cortex Tasks.
- Adds `task.delegate` with owned `start/get/cancel` semantics instead of blocking plugin `task.run()`.
- Persists plugin task owner, generation, Scope, correlation ID, resolved model, output, timestamps, usage, and cost.
[...]

PR #1282, opened 2026-08-29
fix(plugin): deduplicate concurrent runtime starts

### LINKED ISSUE #1281: Plugin runtime startup race spawns duplicate untracked processes
## Summary

Concurrent calls to `PluginRuntimeManager.start()` for the same plugin generation can spawn duplicate external runtime processes. The registry key is shared, so each new entry overwrites the previous `starting` entry; overwritten processes remain alive in the Synergy cgroup but are no longer reachable by normal stop/recycle logic or counted by Performance telemetry.

## Evidence

Observed on Linux with `dev@37788a15e`, then confirmed against current `origin/dev@9559ee48e`:

- Nine `plugin-runtime/runner.ts` processes for the same `holos-research@1.1.1` artifact started in the same second.
- All nine remained children of the Synergy server.
- The manager registry can retain only one entry for `pluginId@version#generation`.
- Eight overwritten processes contributed roughly 390 MiB of extra RSS at the later forensic sample.
- `resourceStats()` enumerates registry entries, so it reports a complete single runtime while the cgroup actually contains nine.
- There is no concurrent-start test in `packages/synergy/test/plugin-runtime/manager.test.ts`.

This is separate from the same incident's Linux page-cache pressure; the duplicate runtimes do not explain all cgroup memory, but they make the baseline larger and make attribution incomplete.

## Root cause

`start()` currently only reuses an existing entry when `existing.state === "ready"`:

```ts
const existing = this.registry.get(key)
if (existing?.state === "ready") return existing
```

If another caller arrives while the first start is still `starting`, it proceeds through startup again and calls `spawnPluginProcess()`. Because both entries use the same key, `PluginRuntimeRegistry.set()` overwrites the first entry without stopping its process.

This remains unchanged on `origin/dev@9559ee48e`.

## Expected behavior

- At most one startup attempt may be in flight per `pluginId@version#generation`.
- Concurrent callers should await the same startup promise and receive the same entry.
- A failed startup should clear the in-flight slot so a later call can retry.
- Runtime resource counts should match the actual managed process count.

## Proposed fix

Add per-runtime-key single-flight deduplication around `start()`, with focused tests for:

1. concurrent callers share one process and one registry entry;
2. failed startup clears the single-flight state and permits retry;
3. generation keys remain independent.

## Reproduction

Call `ensureRuntime(plugin)` concurrently from several scopes before the first external runtime reaches `ready`, then compare OS child-process count with `pluginRuntimeManager.registry.list()` / `resourceStats()`.

marzich:
## Summary
- serialize plugin runtime startup per `pluginId@version#generation`
- make concurrent callers await one startup attempt instead of spawning overwritten processes
- clear failed startup promises so later calls can retry

Fixes #1281.

## Root cause
`PluginRuntimeManager.start()` only reused a runtime after it reached `ready`. Concurrent callers arriving during `starting` each spawned a process, while the shared registry key retained only the last entry. The overwritten processes stayed alive but disappeared from stop/recycle and Performance accounting.

## Changes
| File | Change |
|---|---|
| `packages/synergy/src/plugin-runtime/manager.ts` | Add keyed single-flight startup deduplication with `finally` cleanup |
| `packages/synergy/test/plugin-runtime/manager.test.ts` | Cover nine concurrent callers and retry after shared startup failure |

## Verification
- `bun test test/plugin-runtime/manager.test.ts --timeout 30000`: 15 passed
- `bunx prettier --check packages/synergy/src/plugin-runtime/manager.ts packages/synergy/test/plugin-runtime/manager.test.ts`: passed
- `git diff --check`: passed
- package typecheck was attempted, but the reused worktree dependency state has unrelated pre-existing resolution errors in `plugin-kit`, `synergy-link`, and `server-process-lock.ts`; no error referenced either changed file

## Production note
The live service was not restarted. The observed duplicate runtime processes remain in the current cgroup until a separately approved restart.


yzxoi (approved):
Approve ✅

Clean single-flight fix — exactly what we asked for back in the #566 review. Concurrent `start()` for the same `id@version#generation` key now shares one pending promise, failed starts aren't negatively cached (retry test 👍), and the identity-checked `finally` cleanup is the right pattern. Verified locally: plugin-runtime manager tests 15/15 pass.

One tiny follow-up for a later PR: a `docs/decisions/implemented/bug-fix/…` record, per repo convention (#1276 / #1280 style).

Thanks for the minimal, well-targeted fix and the solid concurrency tests! 🙏

PR #1284, opened 2026-08-29
fix(scope): filter ephemeral test worktrees from sidebar and archive existing artifacts

EricSanchezok:
## Summary
- Filter ephemeral test worktrees from the sidebar scope listing.
- Archive existing test artifacts (8 modified + 6 untracked files) instead of surfacing them.

## Context
Resume of sidebar blueprint work (session fc0e87, interrupted on 8/27 by worker pool recycling). Committed and pushed on 8/29 09:00, verified green in the worktree before pushing.

## Verification
- packages/synergy: 42 + 10 tests pass (scope, cli, artifact filter, migration artifacts)
- packages/app: 39 + 6 tests pass (incl. ephemeral-test-worktree), build contracts 2 pass
- Root: `bun run test-layout:check`, `bun run decision:check` pass


yzxoi (dismissed):
Thanks @EricSanchezok for the clean fix and the thorough decision record — the effort shows, especially picking this back up after the interrupted session and landing it with full test coverage on all three layers.

**Verdict: Approve** ✅

**What I verified:**

- **Layering is right.** Making `Scope.list()` the single authoritative filter covers every backend consumer — traced `scope.list` / `scope.index` routes, `SessionNav.buildScopeIndex`, `boss-runtime`, `session-search`, and the CLI scrap; none bypass it (`listScopeIDs()` has no other callers).
- **Predicate correctness.** The backend basename + tmpdir-containment double guard correctly rejects real projects like `~/projects/synergy-test/foo`, and the macOS `/var` vs `/private/var` realpath handling matches how the fixture scaffolding persists paths.
- **Migration semantics.** Reusing `Scope.remove()` (archive guards + `time.archived`) keeps it consistent with the rest of the domain; idempotency and data preservation are covered by tests. Records whose temp dirs no longer exist still fall through the existing detached-worktree auto-archive path, so both branches converge on archived.
- **Re-ran locally:** 21 backend tests (artifact filter / migration / project / cli scope) pass, 6 frontend tests pass, `test-layout:check` and `decision:check` pass.

**Two non-blocking notes — fine to address in a follow-up:**

1. The decision record's "Real projects are never hidden" is slightly stronger than reality: the frontend basename-only mirror can still hide a real project literally named `synergy-test-*` (browsers can't check tmpdir). The tradeoff is documented; that one sentence could be softened.
2. Pre-existing and out of scope here: `stats/engine.ts getAllSessions()` scans scope records directly without checking `time.archived`, so archived test sessions still count in stats views. Worth a follow-up issue.

yzxoi (approved):

PR #1285, opened 2026-08-30
perf(server): gzip-compress API responses

SII-jake:
## Summary
- Mount Hono's `compress({ encoding: "gzip" })` middleware on the App chain (after cors, before all routes), so the boot-path JSON payloads — `scope/bootstrap` (~10 MB with the full provider catalog) and `session/:id/message/page` (up to 200 messages with inlined tool outputs) — transfer compressed.
- Safety comes from the middleware's built-in guards: `text/event-stream` is excluded so SSE streams stay uncompressed, responses already carrying `Content-Encoding`/`Transfer-Encoding` are skipped (no double-encoding behind a compressing reverse proxy), and strong ETags are downgraded to weak.
- Measured on a real deployment: first-load time dropped from ~10 s to ~3 s (-70%). Session message data gzips to ~20% of raw; the provider model catalog to ~11%.
- Second commit fixes a latent bug the new test exposed: `Server.openapi()` and `/doc` each generated specs independently, but hono-openapi consumes route resolvers in place on first generation — any process serving `/doc` then calling `Server.openapi()` emitted `$ref`s without components. Spec generation is now a per-process singleton shared by `/doc`, the CLI, and tests.

## Root cause of the first CI failure
The new compression test was the first test-file-level consumer of `/doc`. In the shard process it ran before `openapi.test.ts`, so the first `generateSpecs` pass (triggered by `/doc`) rewrote the route resolvers, and the subsequent `Server.openapi()` produced paths without components (`providerListSchema` was `undefined`). This only surfaces when two spec generations share a process, which is why it never failed before.

## Test plan
- `bun test test/server/compress.test.ts` — 3 cases: gzip round-trip on a large JSON endpoint, identity pass-through without `Accept-Encoding`, SSE never compressed
- `bun test test/server/openapi.test.ts` — includes new regression case `survives a /doc request earlier in the same process`
- `bun test test/server/` — full server suite
- `bun run typecheck`, `bun run quality:quick` — pass (secrets:check needs gitleaks locally; CI secret-scan covers it)
- CI: all checks green

Decision records: `docs/decisions/implemented/feature/2026-08-30-http-gzip-compression.md`, `docs/decisions/implemented/bug-fix/2026-08-30-openapi-spec-generation-singleton.md`


yzxoi (approved):
## Verdict: Approve ✅

Reviewed locally on the PR head (`8ffd6c1c0`) with dependency-source cross-checks, not just the diff.

### Commit 1 — gzip compression

One middleware mount covering the whole API surface, with the safety contract verified against `hono@4.12.30`'s compress source rather than taken on faith:

- `text/event-stream` is excluded via the compressible-regex negative lookahead — `/event` and other SSE routes stream through untouched, no route-level opt-outs needed.
- Responses already carrying `Content-Encoding` / `Transfer-Encoding` are skipped, so the session-export gzip download is never double-encoded and a compressing reverse proxy composes safely.
- Strong ETags downgraded to weak, `Cache-Control: no-transform` honored, 1 KiB threshold keeps small responses identity.

Tests pin the three public invariants (gzip round-trip on a large JSON endpoint, identity pass-through without `Accept-Encoding`, SSE never compressed). A ~70% first-load improvement for three lines of code is exactly the kind of leverage we want.

### Commit 2 — OpenAPI spec generation singleton

The root-cause analysis is spot on: confirmed in `hono-openapi@1.1.2` that `getSpec` consumes each route's resolver **in place** (`raw.schema = result2.schema` on first generation), so any second generation in the same process emits `$ref`s with no backing components. Making once-per-process a public invariant — one memoized promise shared by `/doc`, the CLI, and tests — is the right structural fix instead of papering over test ordering. Turning a seemingly-unrelated CI failure into a real invariant plus a regression test is great engineering.

### Minor notes (non-blocking, for the record)

- `/doc` `info.version` moved `0.0.3` → `1.0.0` as a side effect of unifying on `Server.openapi()` — actually a good alignment with the canonical spec.
- The compress middleware doesn't set `Vary: Accept-Encoding`; irrelevant for today's direct-connect/LAN deployments, but worth adding at the proxy layer if a shared cache/CDN ever sits in front.

### Verification

- `bun test test/server/compress.test.ts test/server/openapi.test.ts` → 22 pass / 0 fail, including the new `/doc`-first regression case.
- CI fully green across all 11 checks (incl. Windows, Coverage, Secret Scan).
- Decision records for both changes are in place and well-argued.

Thank you @SII-jake for the clean, well-diagnosed work — the commit messages, root-cause writeup, and decision records made this a genuinely pleasant review. 🙌

yzxoi:
Approved in the review above ✅

Thanks @SII-jake for the well-scoped change and the excellent writeups, and @synergy-agent for the co-authored root-cause analysis that turned a seemingly-unrelated CI failure into a proper public invariant with a regression test. Small diff, big payoff (~10s → ~3s first load), and the repo came out structurally safer too. 🎉

PR #1286, opened 2026-08-30
feat(app): draggable persisted sidebar width and top-bar outline

yzxoi:
## Summary

- **Draggable persisted sidebar width** — the expanded global sidebar now mounts the shared `ResizeHandle` on its trailing edge (same component as Side Workspace and the file Explorer). Width is adjustable in a 220–420 px band; dragging below ~230 px collapses to the 48 px icon rail. The chosen width persists in the global layout store as one user preference (not per-workspace) and survives reloads and workspace switches. Only explicit user resizes are remembered: stored widths without a `resized` flag are dropped by `migrateWorkbenchLayout`, so legacy installations carrying the old dead default do not get silently resized on upgrade. The mobile drawer is a separate modal surface and is unaffected.
- **Top-bar bottom outline restored** — the desktop (≥768 px) session top bar regains its hairline `border-bottom`. The line had been reset to `none` incidentally when the bar became an absolute overlay in an earlier mobile keyboard-layout fix; the removal rode along with that change and had no product rationale.

Decision record: `docs/decisions/implemented/feature/2026-08-30-sidebar-resize-topbar-outline.md`. Product contract for the width behavior added to `packages/app/PRODUCT.md`.

## Tests

- `bun test test/context/layout/defaults.test.ts test/context/workbench/layout-migration.test.ts` — new cases written red→green: un-flagged width dropped, flagged width clamped to the band, migration idempotent
- `bun test test/workbench-surface.test.ts test/components/sidebar/` — existing sidebar/workbench suites stay green
- `bun run --cwd packages/app test` — full app suite, exit 0
- `bun run decision:check`
- Manual verification on an isolated dev instance: drag in band, collapse threshold, width persists across reload, top-bar outline visible in light and dark themes


yzxoi:
Addressed all four review comments in 628e34e41:

1. **Keyboard/touch resize (sidebar.tsx)** — fixed at the shared-component layer: `ResizeHandle` now uses pointer events with `setPointerCapture` + `touch-action: none` (touch drag works), renders as a focusable `role="separator"` with arrow/Home/End stepping (edge-direction aware via `resize-handle-model.ts`), and carries localized accessible names at all three consumers (sidebar, side workspace, file explorer; en/zh-CN/pseudo catalogs updated).

1. **Reduced motion (sidebar.css)** — added a `prefers-reduced-motion: reduce` override that disables the `.sb-root` width transition.

1. **Workspace sizing (defaults.ts)** — `sidebarOccupancy()` (expanded width / 48 px rail / 0 on mobile) is now subtracted in both the layout store's default side-workspace width and `maxSideWidth`; a persisted size that no longer fits is clamped at render time (`displaySize`) instead of clipping the session column.

1. **Collapse band (defaults.ts)** — the persisted band floor now equals the collapse threshold (230), so a drag-to-collapse can no longer persist a width in 220–229 and a reopened sidebar never re-collapses on a bare press; the invariant is pinned by a unit test.

Verification: layout/migration/workspace/sidebar/workbench suites (207 pass), ui keyboard-model + css-token-integrity suites, `typecheck` for both packages, `localization:check`, `decision:check` — all green.

PR #1288, opened 2026-08-30
fix(theme): register plugin themes globally to stop session-switch skin flips

yzxoi:
# PR body: global plugin theme registration

## Summary

- **Root cause**: theme *selection* is a global user preference (config `general.theme`), but plugin theme *registration* followed the Scope-scoped plugin-host reload. Switching sessions swapped the theme registry, dropped themes the new scope lacked, and the destructive ThemeProvider fallback overwrote the persisted skin choice — flipping the skin to default and later flipping it back when the registry refilled.
- **Server**: new global endpoint `GET /plugin/ui/contributions/themes` (`plugin.listGlobalThemeContributions`) aggregates `ui.theme` contributions from the process-wide plugin catalog across every enabled scope (`packages/synergy/src/plugin/global-themes.ts`, `packages/synergy/src/server/plugin-routes.ts`, `isGlobalRoute` entry in `packages/synergy/src/server/server.ts`). The plugin asset route now falls back to the process-wide catalog so a project-scope plugin's theme asset resolves globally (the generation check still pins exact artifacts).
- **Web**: plugin themes are registered by `GlobalPluginThemesRegistrar` (`packages/app/src/plugin/global-themes.tsx` + injectable factory in `global-theme-registrar.ts`), mounted inside the router tree so it survives Scope switches; it is now the only caller of `replacePluginThemes`. The Scope-scoped plugin host no longer touches the theme registry and strips `ui.theme` before its asset load to avoid double-fetching.
- **UI**: the `ThemeProvider` fallback is non-destructive — when the ready registry is missing the selected theme (scope not yet activated after restart, transient asset failure), it keeps the selection, renders degraded default tokens, and never overwrites `synergy-skin-cache-v1`; the theme re-applies automatically once it re-enters the registry (`packages/ui/src/theme/context.tsx`).
- **Docs**: `docs/plugins/ui-contributions.md` (themes as the global exception to the Scope reload cycle + cold-start boundary) and `docs/reference/frontend-theming.md` (global registration lifecycle) updated; decision record added at `docs/decisions/implemented/architecture/2026-08-30-global-plugin-theme-registration.md`.

Known boundary (documented): after a server restart, the aggregate omits themes from scopes nobody has visited until one activates; the Web keeps the selection and degraded rendering during that window, then self-heals on the registrar's next refetch.

## Tests

```bash
cd packages/synergy
bun test test/plugin/global-themes.test.ts test/server/plugin-global-theme-routes.test.ts
bun run typecheck
bun run --cwd packages/ui test        # includes new test/theme-provider-fallback.test.ts (browser-conditions batch)
bun run --cwd packages/app test       # includes new test/plugin/global-themes.test.ts
```

- `./script/generate.ts` regenerated; SDK/OpenAPI diff contains only `plugin.listGlobalThemeContributions`.
- `bun run quality:quick` gates run individually: format / lint / typecheck / deadcode / localization / doc / decision / deps / monorepo / package / brand all pass. `secrets:check` (gitleaks) and `workflow:check` (actionlint download) could not run on this machine — CI owns both; this diff touches no workflows.
- Manual end-to-end on an isolated dev instance with a fixture theme plugin: selected the plugin skin, then same-scope session switch, cross-scope switch, rapid alternating switches, and server restart + page reload — the skin never flipped or lost. Cold-start aggregate confirmed empty before the owning scope activates, and the theme re-applied after activation.


yzxoi:
@codex @synergy-agent review

PR #1289, opened 2026-08-31
feat(app): open HTML attachments in the user's browser via new tab

SII-jake:
## Why

Interactive HTML attachments (e.g. Archify-generated diagrams) render **blank** in the attachment workbench: the preview strips all `<script>` tags (DOMPurify) and wraps the result in a sandboxed iframe with `script-src 'none'`. For script-driven documents this silently degrades to a half-broken page — some bare text survives, the actual content does not — with no indication of why, and no easy way to view the file as intended.

## What

- **Add "Open in your browser"** toolbar action on all attachments, via `platform.openLink`: opens the asset URL (`…/asset/<hash>.<ext>`) directly in the user's own browser — `shell.openExternal` into the OS browser on Desktop, `window.open` new tab on web. No WebRTC, no host process, no `server.cors` / origin whitelist dependency. The URL derives from the current server connection (`sdk.url`), so it transparently follows whatever origin/prefix the user accessed Synergy through.
- **Sandbox HTML assets into an opaque origin (server-side security fix).** `asset.ts` serves `text/html` with `Content-Security-Policy: sandbox allow-scripts allow-forms allow-popups allow-modals`, so interactive documents still render and execute their own scripts but with no same-origin authority over the app: no app localStorage, no unauthenticated HTTP control plane, no event WebSocket, no Desktop preload bridge. Attachment content is untrusted; the previous shape would have handed a malicious or prompt-injected attachment full app-origin privilege. SVG is intentionally excluded — it is consumed via ``/CSS where scripts never run.
- **Rename "Open in browser" → "Open in Synergy browser"** to disambiguate the two actions; the side-panel WebRTC surface keeps its existing semantics unchanged. The new-tab action gets a distinct `action.export` icon.
- **Add a notice in the HTML preview** explaining that interactive documents do not run scripts there, pointing at the new "Open in your browser" action.

## Notes

- English + zh-CN translations included; `i18n:extract` and strict `i18n:check` pass; `typecheck` passes.
- New behavioral tests on the asset route cover the sandbox policy and non-HTML pass-through (`packages/synergy/test/server/runtime.test.ts` — 12/12 pass). The narrow-panel contract fixture now includes the fourth toolbar action with the updated labels.
- `packages/app/test/components/attachment-workbench/pdf-preview.dom.test.ts` and `expectSessionWorkbenchPaneTracksBottomSurface` fail on this machine even with this PR's changes reverted (isolated via temporary reverse-patch) — baseline/environmental here, unrelated to this PR. Noting that a reviewer's checkout reported these passing, so the discrepancy itself may merit a separate look.


SII-jake:
Thanks for the careful review — the blocking issue was real. Fixes pushed in `f08b21cd7`, item by item:

**Blocking (same-origin execution of untrusted HTML) — fixed server-side.**
`asset.ts` now serves `text/html` assets with `Content-Security-Policy: sandbox allow-scripts allow-forms allow-popups allow-modals`. The document gets an opaque origin: scripts still run (Archify-style self-contained documents are unaffected) but app localStorage, the unauthenticated HTTP control plane, the event WebSocket, and the Desktop preload bridge are all cross-origin now. Deliberately did **not** apply it to `image/svg+xml` — SVG is overwhelmingly consumed via ``/CSS, where scripts never execute anyway, and sandboxing would needlessly change that rendering path.

**Medium (Desktop opens an Electron child window, not the OS browser) — fixed.**
The action now goes through `platform.openLink`, which routes via `shell.openExternal` on Desktop and `window.open` on web — same pattern as the sidebar GitHub link and provider auth flows. The label ("Open in your browser") is now accurate on both surfaces.

**Suggestions:**
- **Tests**: added two behavioral tests on the asset route (`html assets render in a sandboxed opaque origin` / `non-html assets do not receive the sandbox policy`) — 12/12 pass in `runtime.test.ts`. Also added the fourth toolbar action to the narrow-panel contract fixture with the updated labels. Skipped a component-level notice-render assertion: the notice lives in the Solid render layer and `html.test.ts` is a pure sanitizer unit test; standing up a component test host for one `<div>` felt disproportionate. Happy to add it if you feel otherwise.
- **Icon**: the new-tab action now uses `action.export`; `action.open` stays with the Synergy browser panel action.

On the verification mismatch: `expectSessionWorkbenchPaneTracksBottomSurface` fails on this machine even with my fixture edit reverted (isolated via temporary reverse-patch), so it's environmental/baseline here rather than something this PR introduced — good to know it passes in your checkout; the discrepancy itself might be worth a look separately.

yzxoi:
Reviewed and pushed a hardening pass to this branch (`f8a93f00b`):

1. **SVG now sandboxed like HTML** — the new open-in-browser action is offered for every attachment, and a top-level SVG would run under the SPA baseline CSP (`script-src 'unsafe-inline'`) with full same-origin authority. `asset.ts` now applies the same `sandbox allow-scripts allow-forms allow-popups allow-modals` CSP to `image/svg+xml`.
2. **`/global/event/ws` is now Origin-gated** — WebSocket isn't covered by CORS, so a sandboxed attachment page (`Origin: null`) could subscribe to the full session event stream. Same-origin/loopback clients pass; opaque, missing, and cross-origin origins are closed with `1008`. Verified live: app on `127.0.0.1:3001` still receives events; `null`/`evil.example` get rejected.
3. **`noopener,noreferrer` on web `openLink`** so a sandboxed page can't navigate the opener tab.
4. Icon → `action.external`, plus a decision record per repo policy.

All green: runtime 18/18 (incl. new SVG-sandbox + WS-origin tests), workbench 19/19, css-contract, i18n strict, typechecks, `decision:check`.

yzxoi (approved):

PR #1290, opened 2026-08-31
feat: open workspace HTML files in a sandboxed browser tab

SII-jake:
## Summary

Workspace HTML files in the file workbench can now be opened in a new browser tab with JavaScript executing, while staying isolated from the app by a sandboxed opaque origin.

- New path-shaped route `GET /workspace/files/raw/{scope}/{path...}`: the scope segment is the literal `home` or a base64url-encoded directory (decoded in `provideRequestScope`, no header/query scope needed). Because the URL is path-shaped and ends in the file name, relative `src`/`href` references in the HTML (`assets/cover/image1.jpeg`, CSS, JS, fonts) resolve to siblings under the same prefix automatically — the query-style `/content?path=...&raw=true` form broke this and 400'd via the scope-required prefix (plus Chrome ORB on the JSON body).
- HTML responses (`.html`/`.htm`) carry `Content-Security-Policy: sandbox allow-scripts allow-forms allow-popups allow-modals`, mirroring the attachment asset route: scripts run but the page lands in an opaque origin, so it cannot reach app `localStorage`, the unauthenticated HTTP control plane, the event WebSocket, or the Desktop preload bridge. Non-HTML resources pass through with their real MIME type and no sandbox CSP. Traversal/absolute/NUL segments are rejected (400/403/404) without leaking existence or content.
- `GET /workspace/files/content` returns to PDF-only duty; OpenAPI/SDK regenerated.
- File workbench toolbar gains an "Open in browser" pill (icon + label, `.html`/`.htm` only) that builds the URL via `buildWorkspaceFileBrowserUrl(baseUrl, path, scope)` and opens it through `platform.openLink` (new tab on web, shell open on Desktop). i18n for English / 中文 / pseudo.

Decision record: `docs/decisions/implemented/feature/2026-08-31-workspace-html-open-in-browser.md`

## Test plan

- `cd packages/synergy && bun test test/server/workspace-files.test.ts` — 35/35 pass (new: sandboxed HTML, `.htm`, static resource MIME + no sandbox, nested subdirectories, traversal/encoded-traversal rejection, missing scope token 404, missing file 404)
- `cd packages/app && bun test test/utils/workspace-file-url.test.ts test/components/file-workbench/open-in-browser.dom.test.ts` — 9/9 pass
- `bun run typecheck` (app: `tsgo -b`; synergy pkg clean) — no errors
- `bun run i18n:check` — contract passes
- Manual: opened a slide-deck HTML file with relative images (`assets/cover/image1.jpeg`) in a new tab — page renders with scripts executing and images loading

Note: `bun run test:changed` reports pre-existing failures in embedding/config/snapshot domains on this machine (unrelated to these files; present on the base as well).


yzxoi:
Reviewed and pushed a hardening pass to this branch (`2b78ae510`):

**Sandbox SVG/XML in the raw route** — the route previously sandboxed only `.html`/`.htm`, but the workbench's raw route serves any file as a top-level document. An SVG with `<script>` (or XML with XSLT processing instructions) opened in a new tab would run under the global SPA baseline CSP (`script-src 'unsafe-inline'`) with full same-origin authority — app localStorage, the unauthenticated HTTP control plane, the event WebSocket, the Desktop preload bridge. Verified empirically: `GET /workspace/files/raw/<token>/evil.svg` returned the baseline CSP, not the sandbox. Now `.svg`/`.xml`/`.xhtml` get the same `sandbox allow-scripts allow-forms allow-popups allow-modals` CSP as HTML, mirroring `asset.ts`. Static sub-resources (images/CSS/JS/fonts) keep their real MIME type and no sandbox, since a sub-resource's CSP doesn't apply to the parent page.

Tests: workspace-files 37/37 (added SVG + XML sandbox cases), typecheck + decision:check green; decision record updated.

Note: this branch still lacks the `/global/event/ws` Origin gate from #1289 — worth merging #1289 (with its hardening) first, then rebasing this PR, since the sandbox's isolation claims depend on it.

yzxoi (approved):

PR #1299, opened 2026-09-01
fix(app): restore expanded workspace folders on file explorer mount

SII-jake:
## Summary
- Fix: expanded folders in the workspace Files panel now restore their children when the file explorer mounts, instead of rendering collapsed after a session switch.
- Root cause: the expanded-folder list is persisted per workspace, but the tree node/directory caches live only in the `FileProvider` context. Switching sessions across Scopes recreates the provider with an empty tree, and the mount hook only loaded the root directory — so previously expanded folders painted with a stale `expanded=true` flag and no children (the first click just cleared the flag, hence the perceived "two clicks to expand").
- Fix: `FileExplorer` walks the persisted expanded list on mount and calls `loadChildren` for each folder; the directory cache's existing complete-and-fresh fast path skips folders whose children are already warm.
- Adds a Playwright DOM regression test (`explorer-restore.dom.test.ts`) with a stub file model holding persisted expansion state against an empty tree, and registers it in the isolated test runner.
- Adds a bug-fix decision record.

## Test plan
- `bun test test/components/file-workbench/explorer-restore.dom.test.ts` — red before the fix (1 row vs expected 3), green after. ✅
- `bun run script/test.ts` (full `packages/app` orchestrator, including all isolated Playwright suites and the production-build CSS contract) — passes. ✅
- `bun run --cwd packages/app typecheck` — passes. ✅
- `bun run quality:quick` — 15/16 gates pass locally; `secrets:check` requires gitleaks which is not installed on this machine (covered by CI). ✅
- `bun run decision:check` / `bun run doc:check` — pass. ✅

🤖 Generated with [Synergy](https://github.com/SII-Holos/synergy)

### DIFF
--- a/docs/decisions/implemented/bug-fix/2026-09-01-file-explorer-restores-expanded-directories.md
+++ b/docs/decisions/implemented/bug-fix/2026-09-01-file-explorer-restores-expanded-directories.md
@@ -1,0 +1,23 @@
+# Decision Record: File explorer restores persisted expanded folders on mount
+
+Status: implemented
+
+## Problem
+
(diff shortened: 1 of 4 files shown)


yzxoi (approved):
✅ **Approved** — tested manually and verified by CI

I ran an isolated dev instance with this PR applied and tested the fix:

- **Session switching** — switching between sessions across scopes now restores previously expanded folders on mount; no more collapsed folders or "two clicks to expand".
- **Opening files** — opened files from the restored tree; everything works great.

Automated evidence also checks out: the new `explorer-restore.dom.test.ts` is red without the fix (`Expected: 3, Received: 1`) and green with it; `typecheck`, `decision:check`, Prettier, and the adjacent file-workbench suites all pass; CI is fully green.

The one-line mount change reuses the existing `loadChildren` fast path (`complete && !stale`), so warm folders aren't refetched — clean, minimal, well-documented fix. Thanks for the great work, contributors! 🙏

PR #1301, opened 2026-09-01
fix(app): stop sidebar project list jump on unrelated navigation

yzxoi:
## Summary

Clicking any project or session under the sidebar Projects section made the whole project list visually "jump" — every row faded and slid as if freshly mounted. Runtime instrumentation traced it to three compounding causes, fixed at each owning layer:

1. **Server event storm** — `Scope.fromDirectory` rewrote the persisted scope record and broadcast `scope.updated` on every directory-scoped request, even when nothing changed. It now persists and broadcasts only when the record actually changed (first creation, or a change in `directory`/`worktree`/`vcs`/sandboxes).
2. **Frontend index churn** — the layout store rebuilt `scopeIndex` after every debounced `scope.updated` event. `loadScopeIndex` now compares the fetched index against the current one via a new `sameScopeIndex` helper and skips `setScopeIndex` when equivalent, preserving signal identity for downstream memos.
3. **FlipList baseline pollution** — Solid's render effect fires once at mount before the container ref is assigned, so the first pass stored an empty position map; every later refresh treated all rows as new and replayed the entrance animation. The FLIP logic moved to `flip-list-model.ts` and now skips the pre-ref pass, so the first real snapshot is the baseline and only genuinely new/repositioned rows animate.

## Verification

- `cd packages/synergy && bun test test/scope/` — 43 pass (includes new `from-directory-event-storm.test.ts`)
- `cd packages/app && bun test test/components/sidebar/flip-list.test.ts test/context/layout/nav.test.ts` — 38 pass
- `bunx turbo typecheck --filter synergy` and `--filter @ericsanchezok/synergy-app` — pass
- `bun run decision:check` / `bun run test-layout:check` — pass
- `bun run quality:quick` — code gates pass; the two remaining failures are environment-only (actionlint binary unavailable, gitleaks not installed locally)

Decision record: `docs/decisions/implemented/bug-fix/2026-09-02-stop-sidebar-project-refresh-jump.md`

### DIFF
--- a/docs/decisions/implemented/bug-fix/2026-09-02-stop-sidebar-project-refresh-jump.md
+++ b/docs/decisions/implemented/bug-fix/2026-09-02-stop-sidebar-project-refresh-jump.md
@@ -15,8 +15,8 @@ Clicking any project or session under the sidebar Projects section made the enti
 Each layer stops triggering its downstream churn only when its own input changed.
 
 - `Scope.fromDirectory` (packages/synergy) persists and broadcasts `scope.updated` only when the record actually changed: first creation, or a change in `directory`, `worktree`, `vcs`, or the sandboxes list. Unchanged lookups stay read-only.
-- `FlipList` (packages/app) skips the pre-ref pass entirely: when the container ref is not yet assigned it returns without touching `previousPositions`. The first real snapshot becomes the baseline, so later refreshes animate only genuinely new rows and repositioned rows.
-- `loadScopeIndex` (packages/app) compares the freshly fetched index against the current one with a new `sameScopeIndex` helper and skips `setScopeIndex` when the entries are equivalent, preserving the signal identity that downstream memos depend on.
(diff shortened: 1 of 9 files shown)


yzxoi:
Addressed both review findings in f98375ae7.

**Codex P2 (FlipList baseline):** FlipList now seeds its baseline from `onMount` — once the container ref and its rows are in the DOM — so the first genuine entries change animates new or repositioned rows instead of being absorbed as a silent baseline. The runner still returns without touching `previousPositions` while the ref is unassigned, so the mount-time empty snapshot can never pollute the baseline.

**Managed channel ordering freshness:** Managed Channel projects order within their account by the server's `latestActivityAt`. With the per-request event storm gone, `session.updated` events for managed project sessions (`category === "channel"` + `channelAccountId` + managed scopeID) now schedule the same 300 ms debounced scope index refresh. The `sameScopeIndex` equality guard keeps it inert when nothing changed, so generic project churn cannot return.

Verification: 40 tests pass across the two affected suites (including new coverage for the predicates), app typecheck passes, decision record updated.

PR #1302, opened 2026-09-02
feat(app): add close-other-tabs action to workbench tab surfaces

SII-jake:
## Summary

- Add a "Close other tabs" action for workbench tab surfaces (Files, Browser, Terminal, plugin panels), which previously only supported one-by-one tab closing.
- Two entries share one model path: a toolbar more-menu targeting the active tab, and a right-click context menu on any tab header offering "Close {title}" and "Close other tabs" without requiring prior activation.
- The workbench context awaits every closed tab's `onCloseTab` hook (terminal pty removal, plugin cleanup) before committing one atomic store update; single-tab surfaces render the item disabled, and the menu-opened tab keeps a hover-equivalent highlight. zh-CN translations ship with the new copy, plus an implemented decision record.

## Test plan

- `cd packages/app && bun test test/workbench-surface.test.ts test/context/workbench/panel-model.test.ts` — 46 pass (4 new close-others model cases + 1 surface contract case)
- `cd packages/app && bun run typecheck`
- `bun run format:check && bun run lint`
- `bun run decision:check && bun run doc:check`


SII-jake:
Thanks for the detailed review — all three findings were spot-on (the sibling-popover issue additionally made the zero-size trigger participate in the sortable flex layout, adding a stray gap beside tabs). Fixed in 91ce14f14:

**1. Escape routing** — `resolveWorkbenchEscapeAction` now takes a generalized `menuOpen` input and returns `close-menu`; the capture-phase handler covers the add menu, the toolbar more-menu, and tab context menus (dismissing whichever is open before falling through to closing the surface). Model tests updated to the new contract.

**2. Popover anchoring** — the context-menu Popover moved inside the tab element, so the `.workbench-surface-tab > [data-slot="popover-trigger"] { position: absolute; inset: 0; visibility: hidden; }` overlay now matches and the menu anchors to the tab's bounds. The trigger no longer participates in the sortable flex layout, and the never-referenced `.workbench-surface-context-trigger` class hook was removed.

**3. \-\-context highlight** — the classList now wires `"workbench-surface-tab--context": props.menuOpen`, and the surface contract test asserts the hookup (plus the Escape routing and the removed class) instead of only the CSS text.

**Minor (keyboard path)**: agreed — the tab context menu is mouse-only for now. The toolbar more-menu remains keyboard-reachable, and every tab already supports the `Delete` key to close, so there's no accessibility regression; a `Menu`/`Shift+F10` keybinding can land as a follow-up.

**Verification**: `bun test test/workbench-surface.test.ts test/context/workbench/panel-model.test.ts` → 46 pass / 0 fail; `typecheck`, `decision:check`, `doc:check` pass; staged lint clean.

yzxoi (approved):
## Review: approve ✅ (review fixes pushed to head)

Reviewed head `868c6c652`. This PR adds "Close other tabs" to workbench tab surfaces (toolbar ⋯ menu + tab right-click context menu) and derives synergy-link pid start times without timezone parsing. The model layer is clean, the earlier review findings (Escape routing, Popover anchoring, context highlight wiring) are addressed in `91ce14f`, and decision records ship alongside the change.

### Issues found in this review round — both fixed on the branch

1. **synergy-link: `etimes` is procps-only, guard silently disabled on macOS** (`72c97ec5`)
   The follow-up fix read `ps -o etimes=`, a keyword macOS/BSD `ps` rejects; the call fell through its catch and `isPidRunningSince` accepted any live pid, disabling the recycled-pid identity guard exactly on the platform that reported the original bug (and in busybox containers). Now reads portable `ps -o etime=` and parses `[[dd-]hh:]mm:ss` via an exported `parsePsEtime`, with a future/distant start-time rejection test so a regression to the silent fallback fails the suite instead of going green.

2. **Workbench: Escape on one surface collapsed the other** (`868c6c65`)
   Side and bottom surfaces each register a capture-phase document keydown listener; `stopPropagation` cannot stop a second listener on the same node, so dismissing a tab context menu on one surface also closed the other open surface. A shared escape-menu registry (`registerWorkbenchEscapeMenu`/`anyWorkbenchEscapeMenuOpen`/`closeAllWorkbenchEscapeMenus`) now arbitrates: a menu-dismiss Escape closes every menu and `stopImmediatePropagation`s before any surface falls through to `close-surface`.

3. **Workbench: close-others async window race** (`868c6c65`)
   While `closeOtherTabsOnSurface` awaited each tab's `onCloseTab` hook, the store could gain tabs (swept by the final commit anyway), lose the kept tab (hooks ran but the commit no-opped), or run the same hook twice against a concurrent `closeTab`. The batch now snapshots its closing ids, bounds the commit to that set (tabs opened mid-batch survive), reuses the per-tab `closeGuard`, isolates hook failures, guards the surface against re-entrant batches, and closes the surface when emptied.

### Verification (macOS local)
- synergy-link: 10 pass incl. new parse matrix + future-timestamp rejection case
- app workbench (panel-model + surface contract): 52 pass incl. 5 new cases (bounded closing set ×3, escape registry ×2)
- `tsgo -b`/`tsc` typecheck, prettier, oxlint 0 warnings, `decision:check` + `doc:check` all green

Note: `closeOtherWorkbenchPanelTabs` no longer matches the original decision record's "closeGuard deliberately not used" line; both decision records were updated to the shipped behavior.

PR #1308, opened 2026-09-03
fix(enforcement): treat null-device sinks as non-paths and write-redirect targets as writes

EricSanchezok:
## Summary

Fix autonomous-profile false-positive denials of benign read-only bash compound commands, and close an inverse write-classification hole.

- **Null-device sinks are never filesystem paths.** Path candidates now terminate at closing shell punctuation (`)`, `}`), and the `/dev/null`-family sink set (`/dev/null`, `/dev/zero`, `/dev/random`, `/dev/urandom`, `/dev/stdin|stdout|stderr`, `/dev/fd/N`) is dropped from absolute-path extraction and shell-argument operand parsing. A redirect such as `2>/dev/null)` glued to a subshell/loop close no longer surfaces as the pseudo path `/dev/null)` → non-bypassable `file_external_write`.
- **`stripAllowedRedirects` recognizes all null-sink spellings** (`>`, `1>`, `2>`, `&>`, `>>` to the null device, including glued closing punctuation), so read-only classification is not misled by an empty sink.
- **Write-redirect targets classify as writes.** A new `writeRedirectTargets` walker treats statically resolvable `>`/`>>`/`>|`/`&>`/`&>>`/`N>`/`<>` targets as genuine writes even when the rest of the segment is read-only — closing the hole where `git status > /tmp/out` misclassified an external write as a bypassable external read (allowed in autonomous, actually writing the host file).
- **Conservative boundaries preserved:** literal external writes (`cp a /etc/x`, `mv` across the boundary, `dd of=/dev/sda`), protected paths, and dynamic-`cd` compounds (`for d in */; do (cd "$d" && …)`) keep their existing denies, asserted by regression tests.

The motivating case — a read-only git repo-scan loop over subdirectories with `2>/dev/null` sinks and a `/tmp` output variable — is now allowed under autonomous instead of hard-denied.

## Verification

- New regression file `packages/synergy/test/enforcement/autonomous-sink-regression.test.ts`: 14 tests covering the repo-scan script (allowed, no external path), the null-sink spelling family (allowed), dynamic-`cd` conservative deny, literal external writes / protected paths (denied), input redirection preserved as read, and write-redirect target classification (external = denied write, workspace = workspace write).
- Enforcement suite: 1212 pass / 0 fail (`bun test test/enforcement`).
- Bash tool + workspace suites: 83 pass / 0 fail.
- Control-profile + sandbox suites: 308 pass / 0 fail.
- `bun run typecheck`, `bun run lint`, `bun run decision:check`, `bun run doc:check`, `bun run test-layout:check`, prettier: all pass.
- `quality:quick`: 15/16 — the single failure (`deps:check`) is a local environment issue (dependency-cruiser does not support the local Node 23.11); CI runs on the supported LTS line and this change adds no imports or dependency-graph edges.
- `test:changed` failures observed during a full run (BossRuntime/daemon/library/migration timeouts) reproduced as load flakes: isolated re-runs pass (e.g. ExperienceRecall UCB1 5/5).

## Scope note

This change is the classifier-fix layer of the Blueprint (Phase A plus the write-redirect security fix). The OS-sandbox enforcement layer (wrapping autonomous bash, controlled temp write roots, fail-closed fallback) is intentionally not included: platform probing showed macOS `sandbox-exec` cannot remap `/tmp` to a controlled root (no mount namespace), the current `workspace_write` sandbox profile cannot express autonomous's full-disk-read / network contract, and the nested environment cannot run sandbox-exec end-to-end. That layer needs a dedicated multi-platform design + E2E matrix and is tracked as follow-up.


EricSanchezok:
Correction to the scope note in the PR description: this PR now includes the OS-sandbox enforcement layer (Blueprint Phase B) in commit d1028bc25.

- Autonomous profile-auto-allowed bash no longer bypasses the OS sandbox: the resolver only calls markShellSandboxBypass for non-autonomous profiles, so autonomous bash runs inside the workspace_write wrapper (variable redirect targets such as `out=/tmp/...` are contained at execution time).
- Autonomous writable roots include the workspace-controlled temporary root (`<workspace>/.synergy/tmp`, session-scoped), the gate seeds its sandbox write roots from the resolved profile roots, and sandboxed bash points TMPDIR/TMP/TEMP at the controlled root. Host shared /tmp writes remain file_external_write deny.
- The autonomous sandbox fallback defaults to deny (fail-closed); guarded keeps warn, and sandbox.fallbackPolicy / sandbox.enabled / profile switch remain the escapes.
- The macOS deny-default backend is repaired: Seatbelt allow rules nest parameters inside path filters (bare param filters are rejected by sandbox-exec), and the profile imports Apple's system.sb (required for child viability on current macOS; verified: workspace + controlled-root writes succeed, host /tmp writes denied with no file created).

Regression anchors: tool-resolver-autonomous-sandbox.test.ts (incl. real-OS-sandbox execution tests), controlled-write-root.test.ts, sandbox-fallback.test.ts; architecture doc and an implemented decision record are updated in the same commit.

yzxoi:
## Review findings — verified on head `373d6c43e`

Verified working: the null-sink fix (`2>/dev/null)` no longer mints a pseudo path) and the `>`/`>>`/`&>` write-redirect classification; all new suites pass locally (22/22). Four blockers remain — three introduced by the Phase B sandbox commit (`d1028bc25`).

### 1. `writeRedirectTargets` alternation dead branch — `>|` write hole stays open (`gate.ts`)

`>` precedes `>\|` in the alternation so `>\|` is dead code: `>\| /tmp/out` matches bare `>`, captures target `|`, and resolves the pseudo path `${cwd}/|`. `>&word` (non-fd) is misparsed the same way (`${cwd}/&`). Gate probe on head (autonomous, worktree workspace):

| command                    | head decision | capabilities                                         |
| -------------------------- | ------------- | ---------------------------------------------------- |
| `git status > /tmp/out`    | deny          | `file_external_write:/tmp/out` ✅                    |
| `git status >\| /tmp/out`  | **allow**     | `file_external_read:/tmp/out` + `file_write:…/\|` ❌ |
| `git status 2>\| /tmp/out` | **allow**     | same shape ❌                                        |
| `git status >& /tmp/out`   | **allow**     | `file_external_read:/tmp/out` + `file_write:…/&` ❌  |

In bash, `>\|word` (clobber) and `>&word` (combined stdout+stderr to file) are unconditional writes — exactly the hole this PR closes. The pseudo paths also leak into the sandbox policy: `getSandboxPolicy()` accumulates writable roots `…/|` and `…/&` plus their `.git` entries in `protectedPaths`.

### 2. The walker is quote-blind — new hard denies for benign commands

`segment.split(/\s+/)` pairs a `>` that is quoted string content with the next whitespace token. `git log --oneline --grep='a > /tmp/y'` → **deny** `file_external_write:/tmp/y` on head (base `dev`: allow). The `>` is a search pattern inside single quotes; nothing is written. Same family: `rg "> /tmp/out" src`, and `(( i > 5 ))` mints a bogus `file_write:${cwd}/5`.

### 3. Phase B sandbox has no outbound network for autonomous bash

`macos.ts` (deny-default path) hardcodes `approvedNetwork: false`; the Linux helper profile hardcodes `network.mode: "restricted"` (`--unshare-net`). The gate's accumulated `approvedNetwork` and the profile network contract are never threaded into `prepareWrapper`. Real `sandbox-exec` probe with the product-generated profile: DNS resolution fails, direct-IP connect fails. Phase B routes **every** autonomous bash through the sandbox, so `git fetch`/`git push`/`gh`/`curl`/`npm install`/`bun install` are dead — while the autonomous ruleset allows `network_request`.

### 4. Sandbox/permission parity breaks git itself (macOS)

- `workspace/.git` is write-denied (`PROTECTED_METADATA_PATH_NAMES` + per-writable-root `.git` injection), but the gate classifies `git commit` / `git branch` as workspace `file_write` → **allow**. Real sandbox probe: both fail with `Operation not permitted`. Allowed-but-always-fails, under `fallback: deny` with no warning surface.
- Read roots miss developer-toolchain locations. On a Homebrew machine `git log` itself fatal-errors in-sandbox: `fatal: unable to access '/opt/homebrew/etc/gitconfig': Operation not permitted`. `/private/etc` (TLS config for curl), `/usr/local`, `/private/var/select` are likewise unreadable.
- `autonomous` declares `readRoots: ["/"]`, but the sandbox enforces runtime-roots + home sibling-deny, so full-disk read silently becomes workspace + runtime read.

### Fixes

Pushing to this branch next: shell-aware rewrite of `writeRedirectTargets` (quote/escape/`((`/`[[` context, correct operator order, `>&word` handling); network mode threading so autonomous sandboxed bash keeps the profile's network contract; macOS platform read-root additions; `.git` sandbox-write parity with the permission layer while keeping `.agents`/`.codex`/credential protection; regression tests for each.

yzxoi:
## Fixes pushed — both review rounds addressed

Two commits pushed to this branch:

### `5e0543599` — fix(enforcement): quote-aware write-redirect walker and network capability parity

Addresses review items 1–2 (walker dead branch + quote blindness):

- **`>|`, `2>|`, `>&word` now classify as external writes.** The walker is rewritten shell-aware instead of regex-splitting tokens: operator alternation is ordered so `>|`/`&>`/`>>`/`&>>` resolve their real targets, and `>&word` (combined stdout+stderr to a file) is distinguished from fd duplication `>&N`/`>&-`. The pseudo paths (`/ws/|`, `/ws/&`) that previously leaked into the sandbox policy are gone.
- **Quoted `>` is string content**: `git log --grep='a > /tmp/y'`, `rg '> /tmp/y'`, and arithmetic/`[[ ]]` comparisons no longer mint write targets. Heredoc bodies are masked before every extractor (`cat <<'EOF'` body lines are literal data).
- **Network capability parity**: `NETWORK_PATTERNS` now includes the developer workflow commands (`git fetch/pull/clone/push/ls-remote`, `npm`/`bun`/`pnpm`/`yarn` install/add, `go get`/`go mod download`), so the gate mints `network_request` for them.
- 30 new regression tests (all spellings, quotes, arithmetic, heredocs, null sinks, network commands, sandbox-mode parity).

### `df98cbba8` — fix(sandbox): thread gate network mode, macOS read roots, granular .git protection

Addresses review items 3–4 (sandbox/permission parity):

- **Sandbox network follows the gate.** `networkMode` threads from the gate's sandbox policy into the macOS/Linux/Windows backend profiles (previously hardcoded restricted). macOS full networking pairs `(allow network*)` with `system.sb`'s `(system-network)` helper — a bare `(allow network*)` fails DNS under `(deny default)`; validated live with `sandbox-exec` (HTTPS 200, DNS resolves).
- **macOS read roots fixed.** Deny-default profiles now read the developer toolchain roots (`/opt/homebrew`, `/usr/local`, Command Line Tools, `/etc`), and home sibling denies skip explicitly allowed user runtime read roots (`~/.gitconfig`, `~/.config/git`, `~/.bun`) or their ancestors — sibling denies previously won over the allows, so `git log` itself fatal-errored in-sandbox on Homebrew machines.
- **Granular `.git` write protection.** Whole-directory `.git` protection (Seatbelt subpaths, Linux helper ro-binds, and the `.git` metadata-name blanket / create-monitor) is replaced by `.git/hooks` + `.git/config` read-only: `git commit`/`git branch`/index/object/ref writes keep working while the hook-execution and config-tamper surface stays protected. `.agents`/`.codex` remain blanket-protected.
- Sandbox parity suite added (12 tests); existing multi-root/phase2/linux-multi-root tests updated to the granular semantics; docs and the sandbox-containment decision record updated.

**Verification** (all local, macOS): enforcement + sandbox + control-profile suites 1509/1509 pass; `tsgo --noEmit` clean; oxlint 0 warnings; prettier clean. CI is running on the pushed head.

yzxoi:
## End-to-end re-verification of the full PR — two more fixes pushed

Re-reviewed the entire PR on the final head (now rebased onto the latest `dev` via the merge). All previously fixed areas verified, and the audit found two more defects — both pushed in `52fc5d0df`:

### `52fc5d0df` — fix(sandbox): gate network mode on plugin shell.run; skip missing metadata ro-binds

1. **Plugin `shell.run` path was still offline.** `host-services-runtime.ts` prepared its sandbox wrapper without `networkMode`, so a plugin `shell.run` the gate approved as a network command (`git fetch`, `curl`) still compiled a restricted profile and lost outbound access — the same bug class already fixed for the tool-resolver path, left on the second execution chain. It now threads `sandboxPolicy.network.mode` identically.
2. **Linux helper would have hard-failed on every sandboxed bash.** The Linux helper's per-writable-root metadata ro-bind loop mounts `.agents`/`.codex` unconditionally whenever `protectedMetadataNames` is set, and bwrap hard-fails on a missing `--ro-bind` source — most workspaces have neither directory at launch. `df98cbba8` activated this latent loop by emitting `[".agents", ".codex"]`, which would have broken Linux entirely (TS tests only parse the JSON; Rust tests never execute bwrap). Fix: mount only metadata directories that exist; creation inside the sandbox stays blocked by the `ProtectedCreateMonitor`, so the create vector remains covered. New Rust regression test `metadata_ro_bind_skips_missing_and_binds_existing`.

### Full-PR audit summary (head `52fc5d0df`, after `dev` merge)

- **Walker/classification** (`5e0543599`): quote-aware, `>|`/`2>|`/`>&word` hole closed, quoted/arithmetic/`[[ ]]`/heredoc blind spots fixed, network commands mint `network_request`. ✅
- **Sandbox network** (`df98cbba8` + `52fc5d0df`): gate `networkMode` now threads through tool-resolver **and** plugin host-service paths into macOS/Linux/Windows backends; macOS `full` pairs `(allow network*)` with `(system-network)` (DNS validated live via `sandbox-exec`). ✅
- **macOS read parity**: developer toolchain roots read; sibling denies skip explicitly allowed user runtime read roots — `git log` no longer fatal-errors. ✅
- **Git write parity**: granular `.git/hooks` + `.git/config` protection (Seatbelt + Linux helper), `.agents`/`.codex` still blanket-protected; `git commit`/`branch` writes work. ✅
- **No leftover hardcoded `approvedNetwork: false`** in any backend. ✅

### Verification (all local)

- TS: enforcement + sandbox + control-profile + plugin shell-host-service suites **1515/1515 pass** on the final merged head.
- Rust helper: **108/108 pass** incl. the new metadata ro-bind regression test.
- `tsgo --noEmit` clean; oxlint 0 warnings.
- CI re-running on the pushed head.

yzxoi:
@synergy-agent review

yzxoi:
## Verified your re-review findings — fixing all three, pushing next

Thanks for the re-review. I verified each item against head `52fc5d0df` before responding:

1. **Blocking (external reads dropped by the wrapper) — confirmed with a live probe.** After `cat /etc/passwd` is auto-allowed, `gate.getSandboxPolicy().fileSystem.readableRoots` does contain `/etc/passwd` (accumulated at the allow decision), but the `sandboxPrepare` wrapper formula (`tool-resolver.ts:1591`) forwards only `[synergyRoot, ...trustedRoots, ...extRoots, ...input.extraReadRoots]` — so the gate-allowed path is invisible inside the wrapper. Fix: forward the full `sandboxPolicy.fileSystem.readableRoots` (deduped) at both the tool-resolver and the plugin `shell.run` wrapper sites, which also restores gate/allowed-read parity on Linux (`--tmpfs /` plans) without touching the profile's readRoots contract.
2. **Should-verify (Linux full-network `/etc`) — agreed.** Linux readable roots lack `/etc`; when `networkMode: "full"` the child shares the host netns but `resolv.conf` and CA stores are hidden by the tmpfs root. Fix: ro-bind `/etc` when full network is approved (alongside an existence filter on Linux readable roots so missing home-runtime roots can't hard-fail bwrap).
3. **Minor (`sessionKey` has no producers) — agreed.** Threading `sessionKey` from the session into `EnforcementGate.create` → profile resolution makes the sandbox write roots include the session-scoped controlled root, matching the `TMPDIR` derivation in `local.ts`; the decision-record claim then holds.

Tests for each will land in the same push.

yzxoi:
## All three re-review findings fixed — pushed in `db8aff14f`

### 1. Blocking — gate-approved external reads now reach the sandbox wrapper

Both wrapper sites now merge the full gate-readable set into `extraReadRoots` (deduped, order-preserving):

- `tool-resolver.ts` bash `sandboxPrepare`: `[...new Set([...(sandboxPolicy?.fileSystem.readableRoots ?? []), synergyRoot, ...trustedRoots, ...extRoots, ...input.extraReadRoots])]`
- `host-services-runtime.ts` plugin `shell.run`: same merge before `Global.Path.root` + trusted roots

The probe that demonstrated the gap now passes at the unit level: after `cat /etc/hosts` auto-allows under autonomous, `/etc/hosts` is in `getSandboxPolicy().fileSystem.readableRoots` **and** in the prepared wrapper's `extraReadRoots` — so on Linux `--tmpfs /` plans, gate-allowed external reads execute instead of ENOENT. Resolver-level regression test asserts the forwarding through the mock wrapper.

### 2. Should-verify — Linux full-network `/etc` ro-bind

`LinuxBackend.prepare` appends existence-filtered DNS/TLS config paths (`/etc`, `/etc/resolv.conf`, `/run/systemd/resolve`, `/var/run/systemd/resolve`) to `readableRoots` when `networkMode === "full"`. The Rust helper ro-binds them read-only; children sharing the host netns can now resolve DNS and verify TLS certs. Regression test asserts `/etc` is present under `full` and absent under `restricted`.

### 3. Minor — `sessionKey` now has producers

`EnforcementGate.create` accepts `sessionKey` and threads it into profile resolution (`ControlProfileCompiler.resolve`). Both tool-resolver gate sites pass `runtimeInput.session?.id`; the plugin `shell.run` gate passes `input.invocation.sessionId`. The autonomous controlled temp root is now session-scoped (`<ws>/.synergy/tmp/synergy-<pid>-<session>`), matching the TMPDIR derivation in `local.ts` — concurrent sandboxed shells cannot see each other's temp files, and the decision-record isolation claim holds. New unit test asserts the session-scoped root lands in `writableRoots`.

### Verification

- New focused tests (readable-root-session, resolver forwarding, Linux full/restricted `/etc`): **20/20 pass** (spot-checked), full targeted set **88/88 pass**
- Full enforcement + sandbox + control-profile dirs: **1512/1512 pass**
- `tsgo --noEmit` clean; oxlint 0 warnings
- CI re-running on `db8aff14f`

PR #1319, opened 2026-09-05
fix(enforcement): precise find/fd exec whitelist, substitution-aware words, slash-relative cd, regex pseudo-path filter

EricSanchezok:
## Summary

Fix autonomous-profile hard denials of benign read-only bash repository-scan commands. Four independent classifier defects (all reproduced live against dev HEAD before this change) denied commands such as LOC-counting loops over `find -exec cat/wc {} +`:

- **F1 — find/fd exec blanket rule.** `ARGUMENT_INJECTION_PATTERNS` flagged *any* `find -exec/-execdir/-ok/-delete` and `fd -x/--exec/--exec-batch` as destructive without inspecting the executed utility, so `find . -exec cat {} +` was treated exactly like `-exec rm`. Replaced with a closed read-only exec-tool whitelist scanner (`READ_ONLY_EXEC_TOOLS`: cat, wc, grep family, head/tail/sort/uniq/cut/tr, checksum tools, …). `-ok` / `-delete` and mutating/interpreter/network/unknown utilities stay `shell_destructive` (fail closed); the scanner recurses into `$()`/backtick/process substitutions.
- **F2 — substitution-blind word tokenization.** `shellWords` had no `$()`/backtick depth, so `files=$(find "$d" ...)` mis-split and the quoted fragment was read as a dynamic command name → `matched destructive pattern: sudo` mislabel + phantom opaque directory risk. Now tracks substitution nesting so an assignment prefix stays one word.
- **F3 — slash-relative cd CDPATH opacity.** `cd packages/synergy/src` was treated as CDPATH-dependent opaque even though the Bash execution env allowlist (`SANDBOX_ENV_ALLOWLIST`) never carries CDPATH. `analyzeDirectoryChanges` gains `resolveSlashRelativeCd` (wired by the gate); slash-relative targets resolve against cwd unless the command text itself defines CDPATH. Bare-name cd (`cd node_modules`) stays opaque.
- **F4 — awk/sed regex literals as paths.** `extractAbsolutePaths` surfaced bare `/`, `/^...` regexes, and metacharacter (`\` `` ` `` `$`) candidates as external write paths. Now rejected; real absolute/protected/`/dev/` paths unchanged.

## Relation to #1308 — both must merge

**This PR is the classifier-precision companion to #1308** ("treat null-device sinks as non-paths and write-redirect targets as writes", already merged into `dev`). #1308 fixed the `/dev/null)` sink pseudo-path family; this PR fixes the find/fd-exec, substitution-word, slash-relative-cd, and regex-pseudo-path families. They touch adjacent regions of the same enforcement files but disjoint rules:

- This branch was **rebased on the latest `dev` after #1308 merged** (`c12585986`), so the two change sets are already reconciled in this diff.
- Both are needed for the reported "autonomous profile still blocks bash" class of false positives to be fully resolved. Merge order is not sensitive; this PR is based on the post-#1308 `dev`.

## Verification

- New regression file `packages/synergy/test/enforcement/autonomous-exec-precision.test.ts` (35 tests): the four reported commands verbatim now evaluate to `allow` under autonomous with no destructive/external-write/sudo tag; destructive find/fd negative matrix (`-exec rm`, `sh -c`, curl, `-delete`, `-ok`, unknown utility, mixed `-exec rm ... -o -exec cat`) stays `shell_destructive`; non-find injection anchors (rg --pre, go test -exec, git show --output, git config --global, shell wrappers, interpreter subprocesses) unchanged; assignment+substitution and slash-relative-cd assertions; awk-regex external-path absence.
- Updated load-bearing anchors in `shell-safety.test.ts` to the refined semantics (read-only exec utilities pass; mutators/unknowns stay destructive).
- `bun test test/enforcement` → 1282 pass / 0 fail.
- `bun run typecheck`, `bun run decision:check`, `bun run doc:check`, `bun run test-layout:check`, `format:check`, `deadcode`, `lint`, `secrets:check`, `workflow:check`: all pass.
- `coverage:check` → all packages pass (`packages/synergy` 75.3%/75%).
- `quality:quick` 15/16 — the single failure (`deps:check`) is the known local environment issue (dependency-cruiser does not support local Node 23.11); CI runs the supported LTS line and this change adds no imports or dependency-graph edges.
- Docs: `docs/architecture/execution-boundaries.md` updated; implemented decision record added (`docs/decisions/implemented/bug-fix/2026-09-05-autonomous-bash-exec-precision.md`).

No SDK/OpenAPI/config/migration/i18n changes.


yzxoi (approved):
Approved after independent verification of the original fixes and the follow-up hardening in 9890f0add.

**Follow-up review findings (all resolved):**

1. **Quote-masked destructive exec (blocking).** `hasUnsafeExecTarget` scanned only quote-masked text and recursed solely into substitutions, so `sh -c 'find . -exec gzip {} +'` and the `eval` / `trap` / `xargs sh -c` payload family evaluated to `shell` → allow under autonomous, while the removed dev-base blanket regexes had denied them. The fix unwraps shell re-parse payloads (`sh -c` / `bash -c`, `eval`, `trap`, function bodies, multicall applets, directory wrapper payloads) before rescanning, mirroring `hasSudoInvocationRecursive`. Verified: every wrapped destructive form now denies, and wrapped read-only forms (`sh -c 'find . -exec cat {} +'`) still allow.
2. **Dynamic slash-relative cd (blocking).** `cd $DIR/x && touch y` and `pushd $D/x && touch y` lost their opaque denial because `slashRelativeCdResolvable` checked shape but not dynamism, pushing the unexpanded literal as a static target. The fix adds `!dynamicDirectoryTarget(target)`. Verified: dynamic `$`/backtick targets stay opaque; static `cd packages/synergy/src` still resolves.
3. **`-okdir` (suggestion).** Now in the always-destructive set alongside `-ok`, matching the scanner's stated contract.

**Verification:**

- `bun test test/enforcement` → 1298 pass / 0 fail (51 in `autonomous-exec-precision.test.ts`, including the wrapped destructive/allow matrices and dynamic-cd opacity anchors)
- Package typecheck clean; prettier, oxlint, sherif, `doc:check`, `decision:check` pass; the full pre-push hook suite (11-package typecheck) is green
- Additional probes beyond the regression tests — heredoc/herestring find-exec payloads, `..` traversal cd chains, `$`-bearing real write paths, sudo/curl inside merged substitution words — are all still correctly denied

**Non-blocking follow-up (not for this PR):** `extractAbsolutePaths` still surfaces awk brace-form regex literals like `/{print}/` as path candidates; observed cases were correctly denied by genuine redirect targets, so this is residual false-positive surface only.

All blocking findings are addressed. The PR's false-positive goal holds without widening autonomous write/execute authority — direct and wrapped destructive forms stay denied, and the reported benign scan corpus still allows.

PR #1352, opened 2026-09-08
feat: show composer attachment uploads immediately with status feedback

SII-jake:
## Summary

Selecting, pasting, or dropping a file in the composer previously showed no feedback until the upload finished — large files left the input silent for seconds and users could send mid-upload, silently dropping the attachment.

- Pending attachment cards appear **immediately** on selection (filename, size, spinner), backed by a session-local tracker that stays out of the persisted draft — a reload mid-upload drops the card instead of persisting a ghost placeholder.
- On settle, the real part is inserted under the same id, flashes "Uploaded", then hands over to the normal attachment card without a jump. Multi-file batches upload in parallel with per-file error toasts.
- Sending is gated while uploads are in flight: the Send control is disabled with an explanatory tooltip/aria-label, Enter-key submits show a toast, and stopping a running session stays available. Removing a pending card cancels its upload; a mid-flight session switch drops the result instead of inserting it into the wrong draft.

Real per-byte progress (XHR `upload.onprogress`) was implemented and rolled back because the signal wasn't visible in the deployed environment; rationale is captured in the decision record.

## Test plan

- `cd packages/app && bun test test/components/prompt-input/` — 141 pass
- `cd packages/app && bun test test/components/prompt-input/pending-attachment-card.dom.test.tsx` — 3 pass (Playwright DOM, runs isolated)
- `cd packages/app && bun run test` — full suite green
- `cd packages/app && bun run typecheck` — clean
- `bun run decision:check`, `bun run test-layout:check`, `bun run localization:check` — all pass

Rebased onto latest `origin/dev` and adapted to the `prompt.capture()` draft-retention refactor from #1343.


yzxoi (commented):
Reviewed upload, cancellation, session switching and submit behavior. Fixed cleared uploads being inserted after returning to their original session, including completion/error cleanup of cancellation markers. Kept remove controls keyboard accessible. Validation: 22 upload/state/submit tests, 3 browser component tests and all 16 local quality gates passed. No remaining blocking findings in the reviewed diff.

yzxoi:
Follow-up coverage audit found only the new pending attachment card missing from Bun lcov; all numeric thresholds passed. Its real Playwright regressions pass. PR #1357 adds an exact-file exemption with the browser-test provenance, preserving every coverage threshold, so the integrated branch includes the correction.

PR #1355, opened 2026-09-08
fix: support snapshot storage on Git 2.25.1

yzxoi:
## What does this PR do?

Restore snapshot storage on Git 2.25.1. Bare repository initialization explicitly selects SHA-1 through `GIT_DEFAULT_HASH`, verifies the resulting object format before recording success, and preserves Git exit code and stderr in initialization errors. Snapshot exclusions use NUL-delimited `update-index --force-remove` input so oversized files can leave the snapshot index without changing project files, the project's Git index, or historical objects.

## Why?

Git 2.25.1 rejects both `git init --object-format=sha1` and `git rm --pathspec-from-file`. The first prevents the initial session snapshot; the second interrupts subsequent captures when files must be excluded. Both operations have compatible forms, so no version probe or retry branch is needed. The storage format remains unchanged.

- [Git init documentation](https://git-scm.com/docs/git-init): `GIT_DEFAULT_HASH`; provenance marker in `packages/synergy/src/session/snapshot-store.ts`.
- [Git 2.25 update-index documentation](https://git-scm.com/docs/git-update-index/2.25.0): `--force-remove`, `-z`, and `--stdin`; provenance marker in `packages/synergy/src/session/snapshot.ts`.

## How was it tested?

From `packages/synergy`, with Git 2.25.1 and Git 2.52.0 separately first on `PATH`:

```sh
bun test test/snapshot test/session/snapshot-isolation.test.ts test/session/snapshot-timeout-regression.test.ts test/session/rollout-snapshot.test.ts test/cli/data-snapshots.test.ts
```

89 tests passed with each version. Verification used actual Git executables on macOS, including Git 2.25.1 built from its upstream release source; the reporting user's Linux server was not accessed. The original initialization failure and exclusion failure were reproduced before their respective fixes. After tightening regression assertions, both versions also passed:

```sh
bun test test/snapshot/initialization.test.ts test/snapshot/shared-store.test.ts
```

Root validation: `bun run quality:quick` and `git diff --check`.

## Checklist

- [x] Relevant tests pass and live under the owning package's `test/` directory
- [x] Storage reference, owning skill, and implemented decision record updated
- [x] Material external sources marked beside the implementation
- [x] No secrets, unrelated cleanup, or redundant compatibility wrappers included
- SDK, package exports, workflows, and frontend contracts are unchanged.


yzxoi:
Addressed the review feedback on `snapshot-store.ts`: the shared `command()` helper now preserves the Git exit code and stderr. Failed commands throw a `StorageError` carrying the exit code in the message and an `{ exitCode, stderr }` structured cause, matching the initialization failure shape, so the post-initialization `hash-object` verification keeps its diagnostics when a repository has `HEAD` but is incomplete or corrupt. A regression test reproduces that failure via a corrupt repository config and asserts both diagnostics.

Also reconciled the branch with `dev` after the package reorganization: the snapshot tests moved from `packages/synergy/test` to `packages/harness/test/snapshot` (with the harness test fixture import updated), and the persistence skill keeps the old-Git verification guidance alongside the relocated benchmark path. Pushed as a fast-forward merge so history stays linear from the reviewed commit.

Verified with the harness snapshot suites (87 tests), CLI `data-snapshots`, product-runtime snapshot tests, `quality:quick` (16 gates), and the full turbo typecheck.

PR #1357, opened 2026-09-08
fix: restore session action menu triggers

yzxoi:
The session action button could render without a Popover click handler because Solid reactive Tooltip JSX was mistaken for a component. Separate rendered `trigger` content from explicit `triggerAs` components, forwarding Kobalte events, ref and accessibility attributes to the actual session button while preserving the public plugin API.

Browser regressions cover reactive Tooltip content, repeated actions, Enter/Space activation, Escape dismissal, focus return and the public plugin component. Each focus scenario uses a fresh document and waits for content autofocus before keyboard dismissal.

Integration validation also fixes two test issues exposed during the batch review: the Channel acceptance completion observer only resolves for its target session, with a second real session exercising isolation; the attachment card added in #1352 has an exact-file coverage exemption because its three Playwright tests execute outside Bun lcov instrumentation. Coverage thresholds are unchanged.

Validation: five consecutive Popover browser runs (20 cases), all 766 connections tests, three attachment-card browser tests, the coverage manifest validator, and all 16 local quality gates passed. The combined frontend production build and production-host conversation/composer/plugin acceptance tests also passed. Commit and push hooks ran normally.


yzxoi (commented):
Reviewed the current Popover and plugin-binding changes. The real browser regression suite passes all 4 tests after building the public Plugin dependency, including pointer/keyboard interaction and focus restoration. No blocking findings. GitHub does not allow the PR author to approve this PR; the current account is the author.

yzxoi (commented):
CI exposed a focus-fixture race: scenarios reused one document and sent Escape as soon as the content was visible, before its deferred autofocus necessarily completed. Isolated each scenario with a fresh navigation and asserted content focus before keyboard dismissal. The original focus-return assertion remains intact. Five consecutive browser runs passed all 20 cases; all 16 local quality gates passed. Production behavior is unchanged by this follow-up.

yzxoi (commented):
The latest CI passed all four Popover scenarios. Its remaining Test failure was an unrelated Channel fixture observing another session wake; a second real session now reproduces that case and the completion observer is scoped to the intended session. All 766 connections tests pass. This integration also corrects the exact-file browser coverage declaration for #1352. Coverage policy validation, three attachment browser tests and all 16 local gates pass. The reviewed diff is limited to the menu fix and these validation corrections.

PR #1358, opened 2026-09-08
fix(sandbox): restore macOS shell selector and authorized-path metadata

yzxoi:
Sandboxed macOS commands cannot read the shell selector, and tools such as Git cannot stat ancestor components of an already-authorized path. Add the macOS selector roots and metadata-only ancestor permissions so these operations work without expanding directory data access.

The shared Git-directory grant has been removed after review. Automatic linked-worktree Git access is outside this PR; it needs a separate design for executable integrations, common-directory resolution and narrow metadata writes.

Validation: 4 enforcement regressions, 37 sandbox policy/parity tests and all 16 local quality gates passed.


yzxoi (commented):
Rechecked all three earlier findings. As agreed with the maintainer, removed automatic shared Git-directory authorization from the PR and narrowed its title, description and decision record to the two macOS fixes. The remaining change only adds shell-selector reads and metadata access on ancestors of already-authorized roots. An enforcement regression verifies no implicit original-checkout Git grant. Validation: 4 enforcement tests, 37 sandbox policy/parity tests and all 16 local quality gates passed.

PR #1361, opened 2026-09-08
fix: pin session switch to conversation bottom and keep jump button honest

yzxoi:
## Problem

Switching sessions intermittently opened the conversation at the top instead of the bottom, and the scroll-to-bottom button could not recover the view. Four defects compounded on one path:

1. **Init pin permanently skippable** — the initial bottom pin was a one-shot rAF chain armed once per session. A `messagesReady` flip (volatile refresh, reconnect replay) cancelled the scheduled chain, but the one-shot `initializedSessions` guard prevented rescheduling, so the forced pin never ran on the freshly mounted scroller (`scrollTop` 0).
2. **No follow after a forced pin in idle sessions** — the pin landed on partially laid-out content with `working` false and `settling` never opened. Late content growth (markdown images, code highlighting) increased `scrollHeight` without any scroll event while the scroller's `overflow-anchor` is deliberately disabled, drifting the viewport away from the bottom.
3. **Jump button blind to non-scroll growth** — button visibility was derived only in the scroll handler. Drift without scroll events never raised it ("无法点击"); when a manual scroll did raise it, the click pinned to the then-current partial layout with no re-pin, reading as a no-op.
4. **Stale hook state across switches** — the `createAutoScroll` instance survives session switches, so a previous session's `userScrolled` state disabled non-forced follow in the new session entirely.

## Fix

- `createAutoScroll`: every forced pin opens a settle window (default 1000 ms) during which the hook stays active, resize-driven growth re-pins to the bottom, and each growth extends the window. Upward wheel still releases follow instantly.
- `createAutoScroll`: new `onMeasure` callback (rAF-coalesced) reports the bottom distance when growth happens outside follow, so scrolled-up state stays honest without scroll events.
- `createAutoScroll`: remounting a scroller resets `userScrolled` (a fresh scroller always starts at the top); an in-flight settle window is left to self-expire.
- Session page: the init chain re-arms when readiness flips false, `onMeasure` is gated until the initial pin consumes (prevents button flash from partial layout), and `scrolledUp` resets on session switch.
- Kanban panes: wire `onMeasure` with the same 100 px threshold (no gate — panes render without a loading switch).

## Verification

- `packages/ui` hook behavior tests (4 passing): settle-window re-pin, growth-driven measurement, remount state reset, plus the existing rAF coalescing test.
- `packages/ui` and `apps/web` typecheck pass; `bun run quality:quick` all 16 gates pass.
- Decision record: [2026-09-09-session-switch-bottom-pin-contract](docs/decisions/implemented/bug-fix/2026-09-09-session-switch-bottom-pin-contract.md); `docs/architecture/frontend-data-sync.md` updated with the pin contract and invariant.


yzxoi (commented):
Addressed all three existing scroll findings and an additional queued-frame leak across scroller remounts. Regression suite passes all 8 tests with browser conditions; Web/UI typechecks and all 16 static gates pass. Required CI checks, Windows checks, and installed runtime artifacts pass on the current head. The only newer base change is the reviewed task-tool guidance PR. No remaining blocking findings.

PR #1362, opened 2026-09-08
fix(session): settle interrupted-turn rollout orphans and retry session wake

yzxoi:
## Problem

An aborted turn can leave rollout call/tool/process records in `running` state. After that, every message send on the session enqueued successfully into the durable inbox, but the fire-and-forget wake chain (`SessionDrive.request` → `scheduleWake` → `wake`) died once at the `finishRun` guard (`Rollout still has active calls`) with no retry. The queued message was never driven, and the UI showed the session as stuck at message sending until a process restart (startup recovery was the only fallback). Production logs recorded 33 such `async session wake failed` events in half a day, concentrated on sessions that had been aborted.

## Changes

- `RolloutLifecycle.reconcile` now settles orphaned records (calls/tools/processes still `running` on a silent run) as `interrupted` before finalizing, mirroring the existing recovery semantics. The active-run guard is unchanged: any running segment still returns the run unfinished.
- `SessionManager.wake` treats the pre-loop `repairAfterAbort` as best-effort — a failed repair logs a warning instead of blocking the loop.
- `SessionManager.scheduleWake` retries wake failures with bounded backoff (250ms/1s/2s/4s/8s), coalesces duplicate requests per session while a chain is live, and logs `retriesExhausted` when it gives up.
- New tests: reconcile settlement behavior (`rollout-recovery.test.ts`) and wake retry/coalescing/bounded-give-up (`wake-retry.test.ts`).
- Decision record: `docs/decisions/implemented/bug-fix/2026-09-09-interrupted-turn-rollout-settlement-and-wake-retry.md`.

## Verification

- Focused: `rollout-recovery`, `wake-retry`, `rollout-ledger` — 14/14 pass.
- Adjacent batches: `rollout-operation`, `rollout-archive`, `rollout-tool-output`, `summary`, `control-profile-transition`, `cortex/session-reuse`, `lifecycle/runtime` — all pass.
- `bun run typecheck` (harness), `bun run lint`, `bun run decision:check` — clean.


yzxoi (commented):
Rechecked the earlier active-process finding and fixed it: orphan recovery now preserves a live RolloutProcess writer. Also reproduced and fixed a lost wake when another release arrives while the existing wake is finishing. Real rollout storage regression plus recovery, ledger and wake tests: 16 passed. Static gates, required CI, Installed Runtime Artifacts, Windows and Coverage passed. No remaining blocking findings.

PR #1364, opened 2026-09-09
fix(web): render decision surfaces inline above the composer

yzxoi:
## Summary

The UI API 5 session-page rework (`855f01ec5`) moved `PermissionDock` and `QuestionPrompt` out of the prompt dock column into a single non-dismissible modal dialog pushed by `SessionDecisionSurface`. Every pending question or permission became a full-screen fixed overlay: a blurred backdrop covered the conversation, Kobalte's modal focus trap blocked all other interaction, and the question prompt's inline-designed collapse control degraded to a stranded pill inside the modal.

This restores the pre-existing inline presentation:

- `SessionDecisionSurface` renders `PermissionDock` followed by the first pending `QuestionPrompt` as plain components — no `Dialog` involved.
- The prompt dock renders it at the top of the active-input branch, so decisions stack vertically above the composer (permission first, then the question, matching the historical ordering).
- To keep host-owned decision views flowing through the UI API 5 presentation contract, `PluginComposerLayoutService.render` gains a host-owned `"decision"` part; the session page implements it and replaceable Shells consume it like the other dock parts. The binding layer already forwards parts opaquely, so plugins are unaffected and the UI API major version is unchanged.

## Test plan

- New browser fixture `apps/web/test/components/session/decision-surface.dom.test.tsx` (registered in the isolated Chromium suite) pins the behavior: a pending question and a pending permission each render inline with zero `dialog`/`dialog-overlay` nodes, and an idle session renders no decision surface.
- `bun test test/components/session/decision-surface.dom.test.tsx` (apps/web)
- `bun run --cwd apps/web test` — full suite green
- `bun run --cwd apps/web typecheck`, `bun run --cwd apps/web build`
- `bun run plugin-ui:test` — production-host acceptance with all eight packed templates
- `bun test test` (packages/plugin) — 121 pass
- `bun run localization:check`, `bun run decision:check`, `bun run quality:quick` (16 gates)

Behavior decision recorded in `docs/decisions/implemented/bug-fix/2026-09-09-restore-inline-decision-surfaces.md`; the `composerLayout` contract enumeration in `docs/plugins/ui-contributions.md` is updated.


yzxoi (commented):
Resolved both earlier findings. Decisions are owned by a host above the replaceable session page; native pages supply a private outlet, while custom plugin pages retain a host-rendered fallback. The combined permission/question stack has one bounded scroll region, including narrow viewports. Removed the proposed public composer-layout union extension. Validation: 5 browser cases, Web typecheck and all 16 local quality gates passed. No remaining blocking findings in the reviewed diff.

PR #1370, opened 2026-09-10
fix(cortex): fence queued follow-ups on cancellation

### LINKED ISSUE #1339: bug(cortex): cancelled child resumes queued follow-up and writes after parent takes over
## Summary

Cancelling a Cortex task can leave its child session able to consume a previously queued parent follow-up as a new root task. The Cortex task remains cancelled while the child continues editing the same workspace the parent has begun to reclaim.

This is an execution ownership bug, not an LLM instruction-following issue. The user also reports that the resumed work is no longer shown as active in the frontend subagent view. Backend continuation is verified from durable records; the exact frontend rendering/filter path has not yet been traced.

## Observed sequence

Inspected a Linux source runtime based on upstream `62c546c60`, with unrelated local provider allowlisting and task-description changes. No cancellation/control code was patched.

Sanitized timeline from one parent/child pair (2026-09-07):

1. Parent sends a corrective `session_send(role: "user")` to its running developer child: stop expanding and run focused tests. Tool returns queued/scheduled.
2. Approximately seven minutes later, parent calls `task_cancel`. Tool reports cancelled; durable Cortex metadata becomes `status: cancelled`.
3. At cancellation, the earlier correction is materialized as a **new root user message** in the same child session. Its origin points to the parent agent. Subsequent assistant messages reference this new root, rather than merely finishing the cancelled root.
4. The child successfully edits a source file about 35 seconds and 118 seconds after the recorded cancellation time.
5. Meanwhile the parent has accepted the cancellation response as proof of exclusive ownership and audits/removes the child's unfinished changes.
6. The child subsequently attempts more edits, sees missing files/reverted changes, and starts investigating an unexpected external reset.

This demonstrates both post-cancellation new-root execution and a real parent/child workspace handoff conflict. It is not just stale UI, a late persisted message, or one already-running tool completing.

## Relevant implementation

- [cortex/manager.ts](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/cortex/manager.ts), `cancel()`: calls `SessionInvoke.cancel(task.sessionID)`, persists cancelled status, and explicitly does **not** await the task run/processor settlement. It does not fence queued follow-up execution for that task.
- [cortex/tools/task-cancel.ts](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/cortex/tools/task-cancel.ts): returns an unqualified cancelled message after that operation.
- [tool/session-send.ts](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/tool/session-send.ts): sends the correction as user mail through `SessionManager.deliver(...waitForProcessing: false)`, without a task-generation binding or an explicit steer mode.
- [session/manager.ts](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/session/manager.ts): queues mail while running; release/finish can request pending work. Task cancellation and session inbox lifetime are not one fenced ownership boundary.

The old Cortex task's terminal status does not describe the new root's live execution. This explains the backend state mismatch; frontend presentation must also be tested.

## Expected behavior

Cancelling delegated work must prevent pending messages belonging to that cancelled execution from silently restarting it. The parent must be able to distinguish cancellation requested from execution fully stopped before taking write ownership.

Explicitly starting a later task in a reused child session must remain possible; do not globally disable normal user queues or permanently poison the session.

## Suggested fix

- Bind delegated follow-ups/steering to task identity and execution generation.
- Fence new starts/writes for the cancelled generation; quarantine or invalidate its queued follow-ups with an explicit disposition.
- Represent cancelling vs stopped distinctly, or return a stop acknowledgement only after the execution boundary is quiescent. Keep the control call bounded and provide a waitable handle if stopping takes time.
- Expose an actual in-flight steer path rather than presenting next-root mail as immediate correction.
- Keep task/session/frontend activity consistent. If explicit new work starts, it must be tracked and visible; a cancelled task must not mask live execution.
- Do not delete unrelated queued user work as a blanket workaround.

## Regression scenarios

1. Start a child, queue a parent correction, cancel it, then release the current tool/model call. Assert no new root or new write from the cancelled assignment.
2. Verify cancellation requested and stopped acknowledgement ordering, including an in-flight tool.
3. Parent takes ownership after stopped acknowledgement; no child writes can occur afterward.
4. Race enqueue/delivery/wake against cancel; repeat cancel and cancel-all.
5. Explicit authorized reuse after cancellation starts a new, visible execution generation.
6. UI continues showing cancellation in progress until stopped, and never hides a genuinely active resumed execution.
7. Normal unrelated user inbox processing remains intact.

Related: #193 covers a broader session inbox/UI inconsistency. This report specifically demonstrates cancelled Cortex work restarting from queued parent mail and conflicting with parent workspace takeover.

yzxoi:
Closes #1339

## Summary

Cancelling a Cortex task — or letting it hit its runtime limit — released the child session while mail queued before the cancellation was still runnable. The release drive and wake chains then materialized that queued follow-up as a new root user message, and the child kept editing the workspace the parent had begun to reclaim, while the durable task status stayed `cancelled` and the frontend showed no active work. This PR fences cancellation end to end:

- `SessionManager.signalAbort` accepts `fenceQueuedWork`, recorded on the loop owner and exposed via `SessionManager.isFenced`; `SessionInvoke.cancel` forwards the option.
- `Cortex.cancel` and the runtime-timeout path abort with the fence and synchronously discard the child session's queued inbox items (task/steer/context) before the terminal status is persisted, so the returned acknowledgement already means the queued follow-ups are gone.
- The session loop's abort boundary extends its existing steer/context disposal to task items when the active abort is fenced; unfenced user aborts keep task items exactly as before.
- A fenced run no longer requests follow-up work on release, closing the release-drive wake path for every internal `SessionManager.run` caller, not just Cortex.
- `task_cancel` no longer returns an unqualified "cancelled": it reports that queued follow-ups were discarded, that in-flight execution is stopping, and that the parent should wait for the session to go idle before taking over the workspace.

Explicit new work sent after cancellation still starts and is driven normally, so session reuse and ordinary user inboxes are unaffected. Rationale and rejected alternatives are in `docs/decisions/implemented/bug-fix/2026-09-10-fence-queued-followups-on-cortex-cancellation.md`; `docs/architecture/cortex.md` cancellation section documents the new semantics.

## Tests

New behavioral coverage in `packages/harness/test/cortex/cancel-queued-followups.test.ts`:

- cancelling a running task discards its queued follow-up, never restarts it, and a new follow-up sent after cancellation still drives
- a runtime timeout fences queued follow-ups the same way

Both reproduced the bug (queued item still runnable after cancellation) before the fix.

Verification run:

```bash
cd packages/harness
bun test test/cortex/ test/session/wake-retry.test.ts   # 77 pass, 0 fail
bun test test/cortex/cancel-queued-followups.test.ts    # 2 pass, 0 fail
bun run typecheck                                        # clean (tsgo)
# from repo root:
bun run quality:quick                                    # 17/17 gates pass
bun run decision:check                                   # pass
```


yzxoi (commented):
Reviewed the current implementation and reused the existing review findings. Fixed the two remaining cancellation boundaries: bulk cancellation now reports failed children after attempting every child, and inbox writes share a lock with cancellation fences/purges. The cutoff includes already-started writes and same-millisecond mail while preserving explicit post-cancellation work. Notifications occur after releasing the inbox lock.

Validation: three new behavioral regressions reproduced the failures before the fix; the Cortex suite passed 75 tests, all 17 local quality gates passed, and cancellation plus terminal-settlement tests passed 12 tests after integrating the current dev branch. No remaining blocking code findings; awaiting full CI on 60351c4073550086a532cdd57f3e246fc8cdfabb.

PR #1374, opened 2026-09-10
fix(session): serialize terminal tool settlement behind per-call state writes

### LINKED ISSUE #1335: bug(session): late Bash metadata can overwrite completed/error tool state with running
## Summary

A late Bash output/metadata update can overwrite an already-settled tool part with `status: running`, removing its terminal output and end timestamp. Observability records successful settlement while durable session state remains running.

Inspected a Linux source build based on upstream `62c546c60` (local changes are unrelated provider-model allowlisting). No production sessions were modified or restarted.

## Evidence

Two sampled Bash calls from separate child sessions show:

| Event | Call A, relative ms | Call B, relative ms |
| --- | ---: | ---: |
| tool.start | 0 | 0 |
| process.child.close | 23 | 17 |
| tool.execute.end | 24 | 17 |
| tool.settle.start (completed) | 31 | 25 |
| tool.settle.end (completed) | 32 | 26 |

Child close reported exit code 1 for A, 0 for B, and `drainTimedOut: false` for both. Bash normally returns its result even for a nonzero shell exit. Tool spans were `ok`, approximately 31 ms and 25 ms.

Subsequent durable parts nevertheless had:
- `state.status: running`;
- captured text in `state.metadata.output`;
- no terminal `state.output` or `state.time.end`.

The two inspected child sessions contained 18 and 38 such running Bash parts respectively, with no duplicate call IDs/terminal counterparts. These observations do not establish that this bug caused all child-session wall time; the model continued doing other work.

## Root cause

Relevant sources at the inspected upstream base:
- [Bash metadata/close handling](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/tool/bash/local.ts)
- [Processor state updates and settlement](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/session/processor.ts)
- [Full-part persistence](https://github.com/SII-Holos/synergy/blob/62c546c60/packages/synergy/src/session/index.ts)

1. Bash's `flushMetadata()` calls asynchronous `ctx.metadata(...)` without awaiting it. `finishClose()` can flush final output and resolve child completion immediately.
2. Processor `flushToolCallState()` snapshots a running part and passes a full `status: running` replacement to `Session.updatePart()`.
3. Metadata updates are serialized with one another via `toolCallStateUpdates`, but `settleToolPart()` does not share/drain that queue.
4. Persistence has asynchronous work before writing the full part. An already-started stale running update can commit after terminal settlement.
5. `settledToolCalls` remains set. Cleanup paths skip those IDs, so the persisted regression is not repaired.

The status guard before the asynchronous write is insufficient to prevent this interleaving.

## Deterministic isolated reproduction

I extracted the current `settleToolPart`, `stageToolCallState`, `flushToolCallState`, `queueToolCallStateFlush`, and `updateToolCallState` functions verbatim into a Bun harness, using the actual pure `ToolTimeout.mergeMetadata` helper. Persistence and observability were stubbed; this is a controlled race reproduction, not a full runtime integration test.

Arrange `Session.updatePart` to pause a running-state write before committing its snapshot, while allowing terminal writes through:

```ts
const pending = updateToolCallState(callID, {
  input: { command: "printf ok" },
  metadata: { output: "ok" },
});
await runningWriteEntered;
await settleToolPart(part, completedOutcome);
assert(disk.state.status === "completed");
releaseRunningWrite();
await pending;
// Actual: disk.state.status === "running"
// state.output and state.time.end are missing
// settledToolCalls.has(callID) === true
```

Repeated with an error outcome: `error -> running` reproduces identically.

## Expected behavior / suggested fix

Terminal tool states must not regress after late output/metadata updates.

- Serialize terminal settlement with per-call state writes, including updates already in flight.
- Reject or safely merge late metadata without replacing terminal state/output/timestamps.
- Await/drain final Bash metadata updates; do not rely only on fixing this producer, because the processor owns lifecycle correctness.
- Keep in-memory tracking and persisted state consistent; cleanup must not silently skip a durably nonterminal part merely because an ID was marked settled.

Suggested regression coverage: completion/error racing final metadata; multiple queued metadata updates; disposal/cleanup; ensure genuine running/background tools retain correct behavior.

Related but distinct: #857 concerns execution never settling after timeout. Here execution and settlement both finish successfully and durable state regresses afterward.

yzxoi:
## Summary

Fixes #1335.

A late tool metadata/output flush could overwrite an already-settled tool part with `status: running`, dropping its terminal output and end timestamp from durable state while observability recorded a successful settlement. `settleToolPart()` now registers the terminal write on the same per-call `toolCallStateUpdates` queue that already serializes metadata flushes:

- settlement waits behind queued and in-flight running-state flushes for the same call, and flushes queued mid-settlement run behind the terminal write
- the in-memory tool part is moved to its terminal state as soon as the durable write lands, so the running-status guard in `flushToolCallState`, the settled-stream-event filter, and the `settledToolCalls` set all stop stale writers
- the unresolved-parts cleanup path settles through `settleToolPart()` instead of issuing its own full-part error write, and `markAutoExpanded()` skips late writes once a call is settled

Bash keeps its fire-and-forget final metadata flush; the processor owns settlement ordering because it owns tool lifecycle correctness. A decision record under `docs/decisions/implemented/bug-fix/` documents the ordering contract and rejected alternatives (awaiting Bash metadata, durable re-read guards, storage-layer versioning, post-hoc repair).

## Test plan

- New regression suite `packages/harness/test/session/tool-settlement-race.test.ts` gates `Session.updatePart` to hold an in-flight running write while settlement runs, asserting the terminal write serializes behind it and the durable part ends completed/error with output and `time.end` intact. Fails on the pre-fix processor; passes after.
- `bun test test/session/tool-settlement-race.test.ts` — 4/4 pass
- `bun test test/session/` — 717/718 pass (one pre-existing `rollout-migration` flake that passes in isolation, unrelated to this diff)
- `bun run typecheck` — 27/27 packages
- `bun run quality:quick` — 17/17 gates


yzxoi (commented):
Reviewed the current head and reused the existing review findings. The latest implementation fences terminal settlement before queued writes, reads live streaming metadata after pending writes complete, and rejects replay during settlement. The regression asserts write/terminal ordering explicitly. No remaining blocking findings.

Validation: targeted tool-settlement tests (4 passed), all 17 local quality gates, and the full CI required gate passed on 2aefc3804a9d29621b414131ecc8be464ee49d51 after updating from dev.

PR #1375, opened 2026-09-10
fix(enforcement): unify read-only shell catalog to end autonomous external-write false positives

EricSanchezok:
Benign read-only compounds such as `ls <dir> && echo --- && ls <dir> | grep pattern` were denied in autonomous mode because command and find/fd execution classifiers used different utility catalogs. Both now validate through one invocation-aware catalog.

Quoted literals are masked only for shell-token scanning; command parsing retains the original arguments so quoted write flags cannot disappear. Validation rejects output operands, writing or subprocess options (including GNU long-option abbreviations), dynamic blocked options, and uniq/xxd batch or expanding inputs that can occupy an output operand. Find/fd validation checks arguments after placeholders and respects their distinct terminators.

Regression coverage preserves benign quoted text and ordinary reads while exercising quoted flags, output paths, substitutions, argument expansion, and find/fd batching. Architecture documentation, the implementation decision, and the execution-boundary Skill record the contract.

Validation: 1,352 passing tests across the relevant enforcement, control-profile, and permission suites; all 17 local quality gates passed. Full platform and coverage checks run in CI.


yzxoi (commented):
Reused the earlier quoted-flag and missing-cmp findings and verified both are fixed. Invocation validation now retains quoted arguments, rejects write/output and subprocess options, and checks find/fd arguments beyond placeholders. Regression tests additionally cover long-option abbreviations and positional output via batch or shell expansion.

Validated 1,352 related enforcement, control-profile, and permission tests plus all 17 local quality gates. No remaining blocking finding in this change; platform checks remain subject to the required CI aggregate.

PR #1378, opened 2026-09-10
fix(cortex): inherit parent session model for delegated sessions

SII-jake:
The native `task` tool previously preferred the subagent default over the parent session's selected model. It now inherits the parent assistant model while that model is available. A configured category model retains precedence, as does the existing optional context model override before category resolution.

If the parent model is unavailable, resolution tries the available agent model. If neither is available, the parent reference remains and normal provider resolution may fail. The shared Cortex manager fallback used by other callers is unchanged. Categories with a blank provider or model now fail before a child is launched.

The architecture and both affected decisions distinguish these paths and preserve cancellation fencing. Tests use real sessions and typed task preparation to cover inheritance, fallback, category precedence, and malformed identities.

Validation: `bun test --cwd packages/harness test/cortex` passed 83 tests; all 17 `quality:quick` gates passed. Full platform and coverage checks run in CI.


yzxoi (commented):
Reused the existing review and addressed its actionable findings. Category overrides now reject blank provider or model components before launching a child. Restored the cancellation fence and timeout documentation that this PR had inadvertently removed.

The older decision remains implemented for shared Cortex manager callers; its scope and the new native task caller behavior are now cross-linked explicitly. Documentation also describes unavailable-parent fallback accurately and does not claim an active writer for the legacy optional context override.

Validated 83 Cortex tests and all 17 local quality gates. New behavioral cases cover parent availability, category precedence, and malformed model identities; the launch boundary uses typed spies with real task preparation and session storage. No remaining blocking finding in the reviewed change.

PR #1381, opened 2026-09-10
fix(web): observe the async-mounted prompt dock so --prompt-height stays live

yzxoi:
The prompt dock mounts asynchronously through the plugin shell. The previous ResizeObserver read a plain element variable before mount and never subscribed to the eventual dock, leaving `--prompt-height` unset or stale and allowing floating controls to overlap the composer.

The dock reference is now reactive. Real Chromium regressions render the default session layout and verify late mount, resize, shell replacement and fractional-height rounding through the resulting CSS layout value. The exact helper has a documented Bun coverage exemption because Vite/Chromium executes it outside Bun instrumentation; package thresholds remain unchanged.

Validation on the latest development baseline:
- Six rendered prompt, inbox and conversation-row tests passed.
- Eight Cortex cancellation regressions passed.
- Production Web build and all 17 local quality gates passed.
- Real-host bounded history and reconnect acceptance passed with delayed message responses. The test now recognizes a completed latest-page recovery when it removes a disabled navigation control before the click finishes; present controls and other click errors still fail. The delayed case reproduced the previous failure before this fix.

The decision record and frontend development Skill document the reactive observer ownership rule. Final integrated CI is checked separately before merge.


yzxoi (commented):
Reused the existing review finding and replaced source-string checks with real rendered layout assertions. The fixture now owns the observer through a Solid component and renders DefaultSession, checking computed --prompt-height after late dock mount, resizing, and shell replacement, including fractional-height ceiling. The non-reactive target regression fails these behavioral checks.

Validated five related browser layout tests and all 17 local quality gates. The author’s original UI geometry evidence is consistent with the implementation. No remaining blocking finding; required CI is pending on the updated head.

PR #1382, opened 2026-09-10
fix(web): attribute viewport reference releases across keyed session swaps

yzxoi:
Switching between cached sessions can mount the successor viewport before the old view cleans up. The old unconditional release then clears the new scroller, leaving the conversation at the top and the back-to-bottom action without a target.

Scroll and content releases now carry the element that acquired the binding. The shared auto-scroll hook, native session view, Kanban and capability-bound plugin conversation service ignore stale releases while preserving ordinary owning and unattributed release behavior. The optional second argument is additive within UI API 5 and is documented for custom viewports.

Review follow-up restores unrelated animation-frame bookkeeping and adds plugin lifecycle regressions covering same-surface replacement, stale cleanup, cross-surface disposal, rejected rebinding after disposal and exactly-once release. These new regressions fail against the previous binding implementation. The author's isolated cached-session switching evidence was reused.

Validation:
- 10 auto-scroll tests and 11 plugin binding/access/lifetime tests passed.
- Full shared UI suite passed when run after artifact-building checks.
- Public plugin package and production Web builds passed.
- All packed UI templates and reference plugin browser tests passed through `bun run plugin-ui:test`.
- Full Web suite passed except its row-retention fixture, which supplied an undefined required reference callback. The corrected fixture then passed its browser test, including new page-error assertions; all other batches passed.
- All 17 local quality gates passed after the fixture follow-up.

The decision record and frontend architecture describe the ownership rule; plugin UI documentation and the owning development Skill carry the cleanup contract.

Integrated follow-up also carries the reviewed Cortex timeout-test setup correction: queue the durable follow-up before starting its short deadline. This removes an unrelated CI setup race without changing runtime behavior. On the latest development baseline, eight cancellation tests, twelve plugin/row-retention tests, the production Web build and all 17 local quality gates passed.


yzxoi (commented):
Reviewed the viewport-to-hook wiring and the capability-bound plugin cleanup path. Element attribution survives replacement and permission disposal; optional release arguments preserve existing UI API 5 callers. Restored unrelated animation-frame bookkeeping and added regression coverage for plugin binding ownership. No remaining blocking finding in the reviewed change. Final integrated CI is checked separately before merge.

PR #1391, opened 2026-09-14
feat(sandbox): flip macOS deny-default reads to a credential deny list

yzxoi:
## Summary

The macOS deny-default sandbox allowed file reads only through an enumerated root list, while the control-plane policy layer already treats `autonomous` reads as deny-list-shaped (`readRoots: ["/"]`). Every tool whose configuration lives outside the enumerated set — `gh` reading `~/.config/gh`, Kubernetes tools reading `~/.kube`, docker, cargo — failed at execution time with `Operation not permitted`, and fixing it per tool meant growing the enumeration forever (git alone took three rounds).

This flips the read side of the macOS deny-default Seatbelt profile to a deny list:

- `MacOSPolicy.compileProfile` now emits a bare `(allow file-read*)` and denies reads only for the credential and sensitive-data paths in the new `READ_DENY_PATHS` (`packages/harness/src/sandbox/policy.ts`). A subpath deny is more specific than the bare global allow and wins under Seatbelt's most-specific-match resolution — verified with real `sandbox-exec` runs: ssh keys and denied paths stay unreadable while arbitrary host files are readable.
- `READ_DENY_PATHS` extends `CREDENTIAL_PATHS` (which gains `.git-credentials`, `.azure`, `.kube`) with browser and mail data stores (`~/Library/Cookies`, `~/Library/Mail`, Chrome/Edge/Brave/Firefox profiles). `~/.config/gh` is deliberately not denied: gh hard-fails when it cannot read `hosts.yml` as configuration even with `GH_TOKEN` present, keyring mode keeps tokens out of that directory, and the Bash tool injects the managed credential as `GH_TOKEN`.
- Deny entries that equal, fall inside, or contain the workspace or a writable root are dropped at profile build, so a project rooted at a credential location keeps its own files readable.
- Write containment is unchanged: writes stay denied everywhere except the parameterized writable roots; `.git/hooks` and `.git/config` stay read-only inside writable roots; `.agents`/`.codex` stay blanket-protected. The Linux bind-mount model is unchanged (binds cannot express "allow then deny"), and the enumerated worktree gitdir grants now only gate what bind-based backends mount.
- Net simplification: the `PATH_READ_*` parameter rules, sibling-deny generation, ancestor metadata allows, and the unused `SBPL_PARAM_NAME` helper are removed from the macOS compiler (−140 lines overall).

Decision record: `docs/decisions/implemented/simplification/2026-09-14-macos-sandbox-read-denylist.md`

## Test plan (run locally)

- `cd packages/runtime-local && bun test test/sandbox/ test/session/tool-resolver-autonomous-sandbox.test.ts` — 227 pass, 0 fail (includes real `sandbox-exec` behavioral probes: denied path unreadable, plain host file readable, host write blocked, workspace write allowed; autonomous bash containment chain)
- `cd packages/harness && bun test test/enforcement/sandbox-readable-root-session.test.ts test/enforcement/multi-root.test.ts test/control-profile/ test/sandbox/ test/workspace/` — pass
- `cd packages/harness && bun run typecheck` / `cd packages/runtime-local && bun run typecheck` — pass
- `bun run decision:check`, `bun run doc:check`, `bun run quality:quick` (17 gates), `bun run test-layout:check` — all pass


yzxoi:
## Review response — findings addressed in d5f3c56

**Fixed:**

1. **SYNERGY_HOME mismatch** — denies now derive from `readDenyHomeDirs()` (OS home + `SYNERGY_HOME`/`SYNERGY_TEST_HOME` when set), so a custom runtime home's `.synergy/data/auth` store is denied. Covered by new harness-side owner tests.
2. **Plugin auth files** — `.synergy/data/plugin` (plugin `<id>/auth.json` root) added to `CREDENTIAL_PATHS`/`READ_DENY_PATHS`.
3. **Firefox path** — corrected to `~/Library/Application Support/Firefox` (the real macOS profile root; also added `~/.mozilla`).
4. **Cargo registry tokens** — `~/.cargo/credentials.toml` and the legacy `credentials` file denied; cargo config/caches stay readable.
5. **Explicit `dataDenyRoots` discarded** — the deny-default branch now forwards non-default explicit deny roots into `buildPermissionProfile`; they merge into the compiled read denies.
6. **Ancestor deny dropped wholesale** — now kept: a workspace nested inside a credential directory keeps the ancestor deny, and its deeper parameterized writable-root allow wins under most-specific-match. Verified with a real `sandbox-exec` carve-out probe (sibling private key unreadable, nested project readable and writable) plus a profile-level test.

**Maintainer decisions (recorded in the decision record):**

- **kube/docker**: `~/.kube` and `~/.docker/config.json` are **exempted from read denial** (operator decision, 2026-09-14, tool-compatibility first). Unlike `gh`, kubectl/docker have no credential-injection fallback — a kubeconfig embeds client certs and tokens, and docker reads `config.json` at startup — so denying the store breaks the tool outright. Write protection is unchanged (both remain in `CREDENTIAL_PATHS`). `READ_EXEMPT_CREDENTIAL_PATHS` makes reverting to strict read denial a one-entry change.
- **gh `hosts.yml`**: kept readable deliberately, with probed evidence — gh hard-fails when it cannot read `hosts.yml` as configuration *even with `GH_TOKEN` present*; keyring mode keeps tokens out of that directory; the Bash tool injects the managed credential as `GH_TOKEN`. A narrow deny of just `hosts.yml` would re-break gh.
- **Per-tool credential injection** (extending the GH_TOKEN bridge to kubectl/docker/etc.): evaluated and rejected as an alternative — it moves the enumeration treadmill from read paths to credential formats and never converges. Recorded under "Alternatives considered".

**Coverage shard**: the harness-side additions now have owner tests (`packages/harness/test/sandbox/policy-engine.test.ts` — deny-list contents, dual-home derivation, carve-out scope, dataDenyRoots merging; 56 pass locally), which should restore the core-server shard coverage that the first commit dropped.

**Verification**: `packages/runtime-local` sandbox + autonomous-session suites 228 pass (including real darwin sandbox-exec probes for the deny-list and the new carve-out case); `packages/harness` sandbox/enforcement suites pass; both packages typecheck; `decision:check`, `doc:check`, `test-layout:check`, and `quality:quick` (17 gates) all green.

PR #1392, opened 2026-09-14
fix(session): preserve rollout input totals without cache breakdown

SII-jake:
# fix(session): preserve rollout input totals without cache breakdown

## What changed

- Use the rollout ledger's exact provider input total for context-usage reconciliation when cache breakdown is unavailable.
- Keep `RolloutAccounting.project()` billing and token projection semantics unchanged.
- Extend rollout transport coverage for providers that report a total input token count without cache detail.
- Add regression coverage for the persisted total, residual reconciliation, and attributed usage while preserving the existing billing projection contract.
- Record the accounting-boundary decision in `docs/decisions/implemented/bug-fix/2026-09-14-rollout-context-usage-input-total-precision.md`.

## Validation

- `bun test test/session/processor.test.ts` from `packages/product-runtime` — 53 passed, 0 failed.
- `bun test test/session/rollout-accounting.test.ts test/session/rollout-usage-capture.test.ts` from `packages/harness` — 9 passed, 0 failed.
- Harness typecheck — passed.
- Product runtime typecheck — passed.
- `bun run decision:check` — passed.
- Prettier checks for all changed files — passed.
- `git diff --check` — passed.
- `bun run quality:quick` — formatting, linting, typecheck, documentation, decision, and 11 additional gates passed; the local secret-scan gate could not run because `gitleaks` is not installed.


yzxoi (approved):
**Approve** — reviewed at 457c614.

**What this fixes.** When a provider reports an exact `input_tokens` total without `input_tokens_details`, the rollout ledger records a known input total with an unknown cache/uncached split. `RolloutAccounting.project()` correctly normalizes the unknown components to zero for billing, but the processor was feeding that billing projection into context-usage reconciliation, so `totalInput` collapsed to 0 and the reconciler scaled every category estimate to zero. The fix routes context-usage enrichment to `stepAccounting.tokens.input.total` (provider-exact, `null` when unknown) while leaving the billing projection untouched, and skips the snapshot when the total is unknown — missing data instead of wrong data.

**Why the design is right:**
- Fixes the billing-projection vs provider-exact-total conflation at the consumer (processor) instead of touching shared accounting/ledger/context-usage modules — minimal blast radius.
- Locks both contracts independently in tests: billing tokens stay `{ input: 0, ... }` while `totalInput = 93_898`, residual reconciliation mode, and `attributedTotal == totalInput` conservation.
- Fail-safe direction is correct: unknown rollout total → no snapshot, rather than persisting an actively wrong `totalInput: 0`.
- The alternatives rejected in the decision record are real options with sound reasoning (re-conflating the projections, deferring the snapshot, inventing a cache split).

**Non-blocking observations:**
- The new `exactProviderInputTotal != null` guard branch (rollout + unknown ledger total + stream reports input usage) has no direct test; existing coverage only exercises the non-rollout missing-usage path.
- `summarizeCalls` sums across attempts, so on multi-attempt retries `totalInput` reflects the sum of attempts rather than the latest call — consistent with existing step accounting semantics, but worth noting against the "latest-call input total" phrasing.

**Verification.** Ran locally on this branch: product-runtime `processor.test.ts` (53 pass), harness `rollout-accounting` + `rollout-usage-capture` (9 pass), `decision:check` pass. The `Windows Checks` failure is in `runtime-local/test/session/tool-resolver-bash-profile.test.ts` — unrelated domain, and sibling branches the same day show the same Windows failure.

PR #1393, opened 2026-09-14
feat(storage): transactional agent authority

EricSanchezok:
Agent authority now commits through an explicit Runtime-owned storage handle instead of independent JSON files. Session, message, inbox, workflow, approval and registry mutations can commit their records, derived indexes, receipts and notifications together. SQLite remains the default; PostgreSQL 16–18 implements the same contract with explicit namespace ownership.

Existing installations upgrade through an immutable byte-for-byte backup, resumable import checkpoints, registered owner migrations, relationship validation and a single activation boundary. Malformed historical Session records retain their evidence and block execution; global corruption stops activation. There is no live JSON mirror. Pack, merge and move use checksummed logical exports and preserve files and Git snapshots separately, including skipped conflicting Session aggregates.

The change also covers bounded streaming writes, stale-writer fencing, worker shutdown, interrupted Session import/fork cleanup, plugin installation recovery, target-switch recovery, patched macOS SQLite packaging, maintenance commands, generated contracts, architecture documentation and operational recovery guidance. Runtime failover and simultaneous replicas sharing a namespace remain separate architecture work.

Validation completed locally:

- `bun run quality:quick` — all 17 gates passed, including typecheck, generated documentation, dead-code and workflow checks.
- `bun test test/storage` with real PostgreSQL 16 — 61 tests passed, including the SQLite contract, owner-loss recovery and three released JSON upgrade fixtures.
- Real CLI cancellation: four direct/process-group SIGINT and SIGTERM cases retained terminal evidence and accounting.
- Targeted fault tests for Session/Note/Blueprint atomicity, staged import recovery, interrupted target activation, worker loss and plugin installation recovery passed.
- Workspace package suites passed through the serialized Turbo run and complete reruns of Workbench and Product Runtime after fixture corrections; Product Runtime passed 1,447 tests.
- Benchmark Python contracts: 248 passed; 19 opt-in integrations were skipped locally. Docker, native harness and streaming integrations run separately in CI.
- The 30 MiB / 30,720-checkpoint completed and failed Rollout tests passed locally; completed, cancelled and failed outcomes passed in CI.
- `bun packages/product-runtime/script/build.ts --single --skip-install` succeeded.
- `bun test test/cli/artifact.test.ts` against the compiled product — all six isolated, outside-repository execution tests passed.

CI includes a mandatory PostgreSQL 16/17/18 matrix in addition to the repository's existing Linux, Windows, coverage and installed-artifact checks. Required check results and the complete job matrix are attached to this pull request.


yzxoi:
## Review Summary: Request Changes (1 blocker)

This is a high-quality, unusually well-engineered refactor. Atomicity/crash-safety wiring verified sound across the session domain (derived indexes commit with canonical writes, durable outbox notifications never precede commit, correct two-phase journal saga, clean reserve-publish-discard staging, no JSON dual-write leftovers, `deriveSemantics` contract untouched). SQL is fully parameterized, path traversal validation is complete, and the fault tests use real subprocess kills rather than mocks.

Four issues should be addressed before merge — one blocking, three strongly recommended (all small, localized fixes):

---

### 🔴 Blocker: `PartWriteBuffer` sticky failure wedges a session permanently (`part-write-buffer.ts`)

- `execute()` records a failure in `this.failures` (line 61) and **no code path ever clears it** (zero `failures.delete`/`.clear()` in the file).
- `defer()` (17-18) and `writeNow()` (74-75) rethrow the retained error *before* executing — so a later terminal write for the same key can never self-heal.
- `flushWhere()` (93-94) re-reports every retained matching failure on every drain; `flushPartWrites(sessionID)` filters by session, so one transient part failure poisons **every subsequent turn of that session** until Runtime restart — even after the database recovers.
- The buffer lives for the store's lifetime (`Storage.state()` in `session/index.ts:1461`), so restart is the only recovery.

Reachability is real: SQLite enforces per-operation deadlines, so one transient BUSY/deadline during a timed streaming flush is enough.

**Suggested fix:** clear the retained failure when a subsequent `writeNow` for the same key succeeds (`failures.delete(key)` in the success handler), plus a regression test that a transient part failure does not block later turns.

### 🟠 WAL sidecars are world-readable (permissions regression vs JSON era)

`PRAGMA journal_mode = WAL` (`sqlite-worker.ts:24`) creates `-wal`/`-shm` under the process umask (typically 0644); the main DB is chmod'd 0600 (`sqlite-driver.ts:90`) but the sidecars never are. On a multi-user host, recently committed session content and plugin approvals leave the 0600 discipline the legacy JSON layer guaranteed (`atomic-file.ts:33`). Fix: chmod 0600 on the sidecars after WAL setup (or a restrictive umask in the worker before open).

### 🟠 `data merge` accepts auth-adjacent record roots

The merge/bootstrap accept-filters (`product-runtime/src/cli/data/transfer.ts:18-25, 93-103`, `harness/src/storage/bootstrap.ts:241-254`) exclude only `local`/`derived` sets. A crafted archive can pre-place `plugin-approvals/records/<id>` with a broad grant, `permissions`, `registry`, `plugin-lock`, etc.; `buildApprovalReview` then finds a sufficient existing approval and suppresses the consent prompt for a later plugin install. Pre-existing in kind (old merge copied `plugin-approvals.json` too), but this PR owns the boundary now — please extend the reject lists to the auth-adjacent roots.

### 🟠 Read path takes a write transaction (`message-v2.ts:1452-1462`)

`messageOrderSnapshot` wraps the cache probe in `Storage.transaction()` — even a cache hit pays a full `BEGIN`/`COMMIT` round-trip through the subprocess writer slot. `readInfoList` (model context assembly) calls it per read; under active streaming + concurrent readers this queues behind the single writer and surfaces as `StorageBusyError` from `defer()`. `Storage.snapshot` (read-only) exists for exactly this; use it for the probe and escalate to a transaction only when a rebuild is needed.

### 🟠 30s IPC deadline kills the worker during `integrity_check` (`sqlite-driver.ts:112-116`)

`verify()` runs `PRAGMA integrity_check` as one request under the 30s per-request deadline. On a multi-GB home this can exceed the deadline → worker killed → driver permanently closed. Fails safe, but `data storage verify` self-destructs on exactly the datasets that need verification most. Suggest a deadline-exempt path for maintenance statements or chunked checking.

---

### Non-blocking follow-ups (tracked for later)

- `storage_nodes` rows are never deleted — index grows linearly with historical writes (`transactional-store.ts:186-194`)
- Retry policy split between store and driver layers with duplicated backoff formula (`transactional-store.ts:527` / `postgres-driver.ts:81`)
- `metadata()` key grammar is an undocumented coupling point for every future indexed domain (`transactional-store.ts:55-81`)
- Orphan `SessionMutation` import (`session/index.ts:60`); unreachable `untrusted` branch in `pending.ts:82-88`; `"__storage-worker-runner"` magic string duplicated across three entry points
- Merge copies archive symlinks verbatim (`cli/cmd/data/shared.ts:259-275`)

### Verified clean

- All storage SQL parameterized; namespace/limit/key-segment validation complete; no data interpolation into statements
- Import integrity: strict schemas, checksummed archives re-hashed end-to-end, identity cross-checks before activation, target-must-be-empty rule
- Plugin approval persistence **strengthened** vs `dev` (atomic transactions + checksummed recovery snapshots)
- `deriveSemantics`/`isSystemPart` contract untouched; PostgreSQL URL hygiene clean

### Validation performed locally

`quality:quick` 17/17 gates; `test/storage` 52 pass / 1 skip (needs real PG); `test/session` 779 pass; `test/migration` 42 pass; product-runtime orchestrated run, CLI/workflows/note/browser-runtime/library/workbench/runtime-local suites all green; CI matrix including PostgreSQL 16/17/18 all passing.

*Review performed by parallel specialist review (security, core storage maintainability, session-domain integration) plus independent verification of every finding cited above. Fixes incoming.*

yzxoi:
All five findings from the review above are fixed and pushed (aaa5fd039..01a0cf732):

- **PartWriteBuffer sticky failure** — a retained failure now clears when a later write for the same key succeeds and retries at every drain boundary (`failures.delete` in the success handler + retry-on-drain in `flushWhere`); `writeNow` no longer rethrows a stale error before executing. Regression tests cover same-key recovery and "transient failure does not block later turns of the session".
- **WAL sidecars world-readable** — the worker now enforces an owner-only umask before opening the database, and the driver re-tightens ``, `-wal`, `-shm` to 0600 on open (also repairing sidecars left by older engines). Regression test asserts 0600 on all three files.
- **Auth-adjacent record roots importable via `data merge`** — a shared `authorityRecordRoots` set (plugin approvals/audit/lock/runtime-state/incompatible/install-intents, permissions, permission-rules, registry) is refused by untrusted `data merge` and the bootstrap convenience import; `data move` (same-home relocation) opts in via `{ trusted: true }`. Regression test covers both directions.
- **Read path on the write transaction** — `messageOrderSnapshot` removed; `loadMessageOrder` reads through a read-only snapshot and escalates to its own transaction only when the stored index needs rebuilding. Cache-hit reads no longer touch the single writer slot.
- **30s deadline kills worker during `integrity_check`** — maintenance statements now carry an extended deadline (600s) through the IPC contract (`SqlQueryOptions.maintenance`); `verify()` passes it for `PRAGMA integrity_check`.

Docs updated (agent-storage architecture, decision record, migration guide). Full local validation: `quality:quick` 17/17, `test/storage` 64 pass (incl. new permission test), `test/session` 781 pass, product-runtime data-transfer 3 pass (incl. new authority rejection test), isolated CLI maintenance suites pass.

Non-blocking follow-ups from the review (storage_nodes pruning, retry-layer consolidation, `metadata()` key-grammar documentation) remain tracked in the review comment above.

yzxoi:
**Follow-up: CI regression fixed by reverting the message-order read-path change (d32db644c).**

The push at bf7813ce8 broke 4 Cortex tests in two shards (channels / product-platform): subagent assistant messages were read back empty after their finalizing turn — summary extraction found no messages, `final_response` captured empty text, usage reads missed the budget trigger, and durable task output came back without its result body.

Root cause isolation (all reproduced locally):

- baseline `aaa5fd039`: Cortex manager suite 55/55 pass
- `bf7813ce8` (buffer recovery + order-read rewrite): 4 fail
- revert **only** `message-v2.ts` while keeping the part-write-buffer fix: 55/55 pass

The read-only snapshot rewrite of `loadMessageOrder` changed read-after-write visibility for the just-finalized turn, so the P-1 optimization is reverted wholesale. The original `messageOrderSnapshot` semantics are restored until the read path can move off the single SQLite writer without changing read visibility (tracked with the other non-blocking follow-ups).

Kept intact: the PartWriteBuffer sticky-failure recovery (B-1), WAL sidecar 0600 permissions (S-1), authority-record rejection on untrusted merge (S-2), and the maintenance deadline exemption (P-2). Session suite re-verified after the revert: 781 pass / 0 fail.

yzxoi (approved):

PR #1398, opened 2026-09-15
fix: keep managed startup alive during storage upgrades

yzxoi:
## Summary

An existing home's storage upgrade can scan and import legacy data longer than Desktop's ordinary 30-second health deadline before the migration runner reports anything. Desktop then kills a healthy but unfinished startup. Report aggregate storage work from the initial scan through backup, portable archive hashing/import, verification and activation, and renew the existing inactivity deadline only when work advances.

Built-in commands and version flags now dispatch without discovering unrelated plugin metadata, so an interrupted storage upgrade cannot block server startup, recovery commands or version output. Root help and plugin dispatch retain namespace validation. Database verification and portable import emit progress from a bounded caller-side observer outside retryable transaction callbacks, counting repeated scan work monotonically.

Once storage is active, legacy-writer detection checks only authoritative namespaces using directory entries. It preserves rejection of recreated records, including nested Rollout metadata, while avoiding full artifact/configuration traversal and per-artifact stat calls.

## Validation

- `bun run quality:quick`: all 17 gates passed.
- Desktop suite: 262 passed, 6 platform/runtime skips; real Electron startup-progress test passed.
- Storage bootstrap/portable/verification suite: 11 passed, including real rollback/retry and caller-context assertions.
- Latest bootstrap and legacy-import suite: 23 passed, including seven recreated-record namespaces and startup with an unrelated artifact directory inaccessible.
- CLI composition suite: 8 passed; managed reporter suite: 6 passed.
- Earlier focused legacy-import, schema and ACP protocol regression checks passed.
- Real isolated managed startup with 25,000 legacy files and 5,000 portable records became healthy in 72.1 seconds; the subsequent restart completed in 1.8 seconds. Both processes were stopped cleanly.

Backup, checksum, ownership and activation policy remain unchanged. Progress contains aggregate counts and stage names only.

Operational verification restored a large home using the previous compatible runtime because the unchanged SQL upgrade space budget could not fit. Its health endpoint and main UI passed before and after a clean restart (34.6 seconds). The full SQL upgrade remains deferred; the isolated upgrade tests above verify the new startup behavior separately.


yzxoi (commented):
Reviewed built-in CLI dispatch, aggregate startup progress, retry-safe observation and authoritative legacy-writer detection at 75e6a80e. Progress only renews the inactivity deadline when work advances, and storage ownership checks still reject recreated canonical records. Focused regression suites, real managed upgrade/restart evidence and current CI are passing. No blocking findings. Longer physical SQLite verification receives its separate bounded maintenance budget in #1402.

PR #1399, opened 2026-09-15
fix(snapshot): reclaim duplicated legacy snapshot storage

yzxoi:
## Summary

Legacy owner registration leaves separate snapshot repositories in place, so repeated project baselines, loose-object allocation and retained old pools continue consuming disk space. Add offline lossless packing and complete shared migration so verified history can be consolidated and duplicate source objects reclaimed.

- `data snapshots pack-legacy` works before SQL bootstrap, inventories all local loose objects, verifies staged packs and preserves unknown objects. It refuses to rewrite a source covered by an interrupted storage backup.
- Session pilots filter registration and execution. Shared migration publishes durable retention refs for up to 32 repositories before switching any owner, supports older Git reference durability, migrates confirmed reclaimed sessions, and handles optional empty roots, virtual empty trees and history already recovered into shared storage.
- Scope-wide migration consolidates old shared pools after their owners migrate. It retains every object, publishes a durable relative Git alternate, and preserves original refs, indexes, unclassified artifacts and unowned borrowers. Retries, relocation and archive export retain readable history.
- Large transfers stage one compressed pack on disk for Git's strict indexer, avoiding JavaScript stdin write failures and retaining producer/indexer diagnostics. Temporary files are cleaned on completion or cancellation; staging requires room for one extra pack. Delta search uses two threads with a 64 MiB window per thread.

## Validation

- `bun test --cwd packages/harness test/snapshot`: 95 passed.
- Actual Git 2.25.1: `git.test.ts`, `maintenance.test.ts` and `pool.test.ts` passed 25 tests; `pack-legacy.test.ts` passed 4 tests.
- Product-runtime snapshot archive suite: 3 passed, including materializing consolidated pool alternates into a portable backup.
- CLI snapshot suite: 2 passed; isolated storage-command suite: 5 passed.
- `bun run quality:quick`: all 17 gates passed; commit/push hooks and the latest CI passed.
- Fault tests cover corrupt staging, rejected destinations, failed durable reference publication, migration checkpoints, interrupted pool cleanup and successful retries with source history retained.
- Isolated capture/restore benchmark: shared allocation was 864,256 bytes versus 8,642,560 bytes for separate repositories across 10 sessions, 200 files and 3 rounds.

## Real migration and recovery

[...]

PR #1401, opened 2026-09-16
perf(storage): avoid repeated scans during SQLite upgrades and exports

yzxoi:
## Summary

Large upgrades repeatedly enumerate migration metadata, then verify every record. SQLite was filtering already-visited kind-index prefixes, while whole-store verification and portable export scanned and sorted the namespace for every page. Prefix lookups used by recovery and Session cleanup could scan the entire namespace even for absent roots, and immediate-child enumeration collected all evidence descendants.

Use a composite cursor for ordered queries and bounded primary-key traversal for complete verification/export. Drive prefix record lookups from the recursive frontier into the key index and stop each immediate-child existence check at its first live descendant. This preserves logical records, revisions, tombstones and namespace isolation without adding an index or changing the database schema. Portable archive record order and therefore archive checksums can change; logical round trips remain equivalent.

## Validation

- Real SQLite query-plan regressions cover ordered seeks, complete exports, existing/absent prefixes and short-circuit child enumeration. The prefix regressions failed before the fix and pass after it.
- Latest integrated `bun test test/storage test/session/storage-indexes.test.ts test/session/nav-index.test.ts` with real PostgreSQL 16: 100 passed, no failures. All 17 `quality:quick` gates passed. Long Rollout CI retains the full 30 MiB stream, 30,720 checkpoints, cleanup and existing deadline.
- `cd packages/harness && bun test test/storage`: 60 passed; the PostgreSQL-only ownership case was exercised separately.
- Shared transactional and ownership tests against a dedicated PostgreSQL 16 instance: 20 passed, including SQLite cases.
- `cd packages/product-runtime && bun test test/cli/data-transfer.test.ts test/cli/data.test.ts`: 28 passed.
- CLI offline-maintenance test: 1 passed. Storage-command tests with their documented testing-package preload: 4 passed.
- Harness typecheck and `bun run quality:quick`: all 17 gates passed.
- Actual before/after export code over the same 100,000-record synthetic SQLite fixture produced identical logical entries. A single local direct-connection run took 49.52 s before and 0.205 s after; this excludes subprocess IPC and migration file I/O and is not a whole-upgrade speedup claim.

Backup packing, migration-inventory compaction and capacity admission remain separate work. This PR does not activate a production Home or claim to reduce its current backup requirements.


yzxoi (commented):
Reviewed composite pagination and key-index maintenance traversal at e829806a. Cursor comparisons preserve tied ordering, descending traversal and deletion semantics; complete scans retain namespace isolation, record revisions, receipts and event export. Real query-plan regressions and shared SQLite/PostgreSQL contract tests cover the changed paths. Current CI passes; no blocking findings. Archive ordering/checksums can change while logical round trips remain equivalent, as documented.

EricSanchezok:
@yzxoi I rebased this branch onto current `dev` and opened #1405 against `codex/sqlite-upgrade-scans` so it can merge without conflicts. Summary of what I found while testing the rebase:

**`dev` moved in two ways that overlap this branch**

- #1404 replaced the traversal liveness expression, so `scan` and `list` conflict textually.
- #1398 / #1399 restructured `verify()` around `observeStorageProgress` while this branch replaced its paging loop with `records()`.

**Your `scan` wins, and I kept it.** The form merged in #1404 only replaced the liveness expression, so root prefixes still collected and deduplicated the whole subtree. On a store with 89,445 live records (minimum of five runs per prefix):

| prefix | previous | #1404 form | your form |
| --- | --- | --- | --- |
| one message's parts, 1 live key | 95 ms | under 0.1 ms | under 0.1 ms |
| session messages, 81 keys | 114 ms | 3.6 ms | 0.7 ms |
| session root, 8 keys | 283 ms | 200 ms | 1.9 ms |
| scope session list, 6 keys | 573 ms | 453 ms | under 0.1 ms |
| `sessions` root, 2 keys | 588 ms | 459 ms | under 0.1 ms |

Enumerated keys are identical across all three forms on every prefix measured, and your `list` form measures equivalent to the merged `IN (SELECT ...)` form (within 0.3 ms) with identical result sets. So the rebase keeps your SQL for both statements and drops the merged revision.

**One regression the rebase exposed, now fixed in #1405.** Combining your `records()` walk with the progress contract that landed in the meantime reported only every 256 records, leaving the final value at 1024 while the total was 1200 once a retried pass replays the walk; `test/storage/verification.test.ts` asserts the total, and it failed on the rebased branch. The walk now reports the running total after it completes as well.

I also updated `2026-09-17-storage-traversal-liveness-probe.md` (merged in #1404) so it describes your shipped early-exit enumeration and defers the traversal implementation to your `2026-09-16-indexed-storage-upgrade-traversal.md`, instead of describing the superseded revision.

**Verified on the rebased branch:** `bun test test/storage` 74 pass / 1 skipped (PostgreSQL) / 0 fail; the rollout, message paging, history, navigation and scope suites 188 pass / 0 fail; `data-storage-command.test.ts` (export, resume, verify, maintenance) 5 pass / 0 fail; harness typecheck, `doc:check`, `decision:check`, `format:check`, `test-layout:check` pass. The PostgreSQL matrix and Windows suite need CI — the plan-shape assertions in `traversal.test.ts` are SQLite-specific by construction. `test/cli/data-snapshots.test.ts` fails in my environment identically on `dev` without this branch, so it is unrelated.

EricSanchezok:
Follow-up: I also checked this branch's CI, because it is currently red on GitHub.

**The only failing job is `Long Rollout Contracts (completed)`**, and it failed by timeout. `retains and validates a 30 MiB provider stream with 30,720 checkpoints` hit its 1,500 s budget (job wall 1,639 s), and the `Test` and `All checks passed` aggregates failed only as a consequence. Every other job passed, including all three PostgreSQL matrix entries — so `test/storage`, with the new `traversal.test.ts` plan assertions, is green in CI.

I could not attribute that timeout to this branch. Local A/B on the same machine, same command (`SYNERGY_ROLLOUT_LONG_STREAM=1 bun test test/session/rollout-long.test.ts --test-name-pattern ": completed$"`):

| revision | result | test time |
| --- | --- | --- |
| this branch, rebased onto `dev` | pass | 539.3 s |
| current `dev` | pass | 543.2 s |

Within 1% of each other, and the phase progress lines track each other too. The job's recent history is also wide: of the last twelve CI runs, ten passed between 280 s and 890 s, while this branch measured 890 s on its first commit and 1,639 s on its second. The 1,500 s budget looks tight against that runner variance rather than broken by the traversal change.

Suggested next step: let #1405 land (it rebases onto `dev` and fixes the verification progress regression the rebase exposed), then re-run CI. If the job times out again, the budget or the 30 MiB workload is the thing to look at rather than this change.

PR #1405, opened 2026-09-17
rebase: replay #1401 on current dev and keep the final verification progress

EricSanchezok:
Rebases `codex/sqlite-upgrade-scans` onto current `dev` (`b058ab0fe`) so #1401 can merge without a conflict, and fixes one regression the rebase exposed. Aimed at @yzxoi.

## Why a rebase is needed

`dev` moved ahead of this branch's base in two ways that overlap it:

- #1404 replaced the traversal liveness expression, so `scan` and `list` now conflict textually.
- #1398 / #1399 restructured `verify()` around `observeStorageProgress`, while this branch replaced its paging loop with `records()`.

## Conflict resolutions

- **`scan`** — kept this branch's immediate-child enumeration with the correlated existence probe. It dominates the form now on `dev`: on a store with 89,445 live records (minimum of five runs per prefix), `sessions` root 588 ms to under 0.1 ms, scope session list 573 ms to under 0.1 ms, session root 283 ms to 1.9 ms, session messages 114 ms to 0.7 ms. The merged form reached only 459 ms / 453 ms / 200 ms / 3.6 ms for the same prefixes because it still collected and deduplicated every node beneath the prefix.
- **`list`** — kept this branch's `CROSS JOIN` form. Measured equivalent to the merged `IN (SELECT ...)` form, within 0.3 ms on every prefix, and identical result sets.
- **`verify()`** — kept `dev`'s `observeStorageProgress` wrapper and its progress contract, and used this branch's `records()` iteration for the walk.
- **`docs/architecture/agent-storage.md`** — kept this branch's traversal sentence, which already states the primary-key probe and the early stop, together with the merged benchmark paragraph.

Enumerated keys are identical across the previous statement, the merged liveness-expression form, and the shipped form on every prefix measured.

## Regression found by testing the rebase

Reporting progress only every 256 records left the final value at 1024 while the total was 1200 once a retried pass replays the walk, so `test/storage/verification.test.ts` failed. The walk now reports the running total after it completes as well.

## Decision record re-measured

`2026-09-17-storage-traversal-liveness-probe.md` (merged in #1404) described the first revision's full-traversal form. It now describes the shipped early-exit enumeration, carries the three-way measurement table, drops an alternative that this branch already implements, and defers the traversal implementation to `2026-09-16-indexed-storage-upgrade-traversal.md`.

## Verification on the rebased branch

- `bun test test/storage` in `packages/harness`: 74 pass, 1 skipped (PostgreSQL), 0 fail.
- `bun test` over the rollout, message paging, history, navigation and scope suites: 188 pass, 0 fail.
- `bun test --cwd ../testing ../cli/test/cli/data-storage-command.test.ts` from `packages/cli`: 5 pass, 0 fail, covering the export, resume, verify and maintenance paths this branch changes.
- `bun run --cwd packages/harness typecheck`: pass.
- `bun run doc:check`, `decision:check`, `format:check`, `test-layout:check`: pass.

`test/cli/data-snapshots.test.ts` fails in this environment identically on `dev` without this branch, so it is unrelated to the rebase.

## Not verified here

The PostgreSQL matrix and the Windows suite run in CI; the plan-shape assertions in `test/storage/traversal.test.ts` are SQLite-specific by construction, and the shared contract covers ordering on both drivers. The export byte order changes described in the traversal record are covered by the portable tests, not re-derived here.


yzxoi:
Reviewed against current dev: the indexed scan/list queries, key-index record traversal, query-plan regression suite, and final verification progress flush are already integrated through #1401 and subsequent storage fixes. The traversal test and benchmark match dev, and dev retains the 1200-record final-progress assertion. This PR still targets the old codex/sqlite-upgrade-scans branch and has no remaining runtime fix to land; closing as superseded rather than merging duplicate history.

PR #1414, opened 2026-09-18
fix(sandbox): make Linux helper mounts follow host reality

EricSanchezok:
## What does this PR do?

Fixes Linux sandboxed execution so the `bwrap`-backed helper can actually start and run commands. Three stacked defects in `LinuxBackend.prepare` made every sandboxed `autonomous` Bash call fail at mount setup; each one masked the next.

## Why?

Sandboxed Bash on Linux was unusable. The failure surfaced as empty stdout, `exit: 1`, and a wall of benign "skipping read-only mount" warnings that hid the single real error line.

1. **Readable roots were passed unfiltered, but bwrap hard-fails on a missing bind source.** `defaultRuntimeReadRoots` is shared across platforms and carries macOS-only entries (`/System/Library`, `~/Library/Caches/…`). `LinuxBackend.prepare` forwarded them as-is, so mount setup died with `Can't find source path /System/Library`. Protected paths and network config roots were *already* existence-filtered — readable roots simply did not follow that same invariant.
2. **No dynamic-linker roots.** `/lib` and `/lib64` are the ELF interpreter entry points on usr-merged distributions, and `defaultRuntimeReadRoots` does not cover them. Even with mounts succeeding, every dynamically linked child died with `execvp … No such file or directory`.
3. **The profile was staged where stage 2 cannot see it.** The helper's plan ends with a controlled-`/tmp` bind, which shadows every host path under `/tmp`. A profile written to the host tmpdir is invisible to stage 2 inside the sandbox, which then re-reads its own profile path and gets `ENOENT`.

- bubblewrap (`bwrap` 0.9.0, system package): a `--ro-bind` source that does not exist is a hard failure at mount setup, not a warning — drives the existence-filter requirement.
- Ubuntu 24.04 default `kernel.apparmor_restrict_unprivileged_userns=1`: without an AppArmor profile granting `userns`, `bwrap` cannot create a user namespace at all. This is a **host prerequisite, not part of this repository change** — see the note below.
- ELF `PT_INTERP` resolution to `/lib64/ld-linux-*.so.2` (or `/lib/ld-musl-*.so.1`) on usr-merged distributions.

These are reflected as inline constraints in `packages/runtime-local/src/sandbox/linux.ts`.

## How was it tested?

```bash
cd packages/runtime-local && bun test test/sandbox/linux-readable-roots.test.ts   # 4 pass / 0 fail
cd packages/runtime-local && bun test test/sandbox/                            # 227 pass / 0 fail
cd packages/runtime-local && bun test test/session/tool-resolver-autonomous-sandbox.test.ts
bun run typecheck
bun run decision:check
bun run quality:quick
```

The new suite includes a real-helper end-to-end execution (prepare → stage 1 → `bwrap` → stage 2 → child) asserting `exitCode: 0`; it skips honestly on hosts with no installed helper or with unprivileged user namespaces blocked, since that is environment rather than a code defect. **On CI the `Test Shards (runtime)` job now enforces this test**: it installs bubblewrap, loads the operator AppArmor userns profile for `/usr/bin/bwrap`, builds and installs the helper to `~/.synergy/sandbox-helper/`, and sets `SYNERGY_TEST_LINUX_SANDBOX_E2E=1` — under that flag every skip condition becomes a hard failure naming the missing prerequisite, so a broken Linux sandbox can no longer hide behind a green build. Local runs without the flag keep the honest skips.

`bun run quality:quick`: 16 of 17 gates pass. `secrets:check` cannot run in this local environment because `gitleaks` is not installed; the change touches no credentials, and the CI secret-scan job covers it.

For a reviewer to verify the end-to-end behavior, a host must first satisfy the AppArmor prerequisite below.

## Important: host prerequisite and scope

This PR fixes the repository-side defects. It does **not** by itself make sandboxed Bash work on a host where `kernel.apparmor_restrict_unprivileged_userns=1` and no AppArmor profile grants `/usr/bin/bwrap` the `userns` permission. On such hosts `bwrap` fails even standalone (`setting up uid map: Permission denied`), and the operator must add the profile and reload it:

```text
abi <abi/4.0>,
include <tunables/global>

profile bwrap /usr/bin/bwrap flags=(unconfined) {
  userns,

  include if exists <local/bwrap>
}
```

Also worth knowing, and recorded in the decision record: the helper's controlled-`/tmp` bind means a **workspace located under `/tmp` cannot be used for sandboxed execution**. Production workspaces are real project directories and are unaffected.

Note that the currently running Synergy instance still executes the old code; this fix takes effect after restarting from merged source. A separate readiness gap is *not* addressed here: `readiness.ts` probes only `bwrap --version` and the sysctls, so an AppArmor `userns` block passes readiness while execution fails.

## Checklist

- [x] `bun run quality:quick` passes (16/17; `secrets:check` needs `gitleaks`, unavailable locally)
- [x] `bun run package:check` passes (included in the gate run above)
- [x] `bun run workflow:check` passes (no workflow changes)
- [ ] Browser capability or App bootstrap changes pass the browser crypto contract and private HTTP browser smoke — not applicable
- [ ] `bun run secrets:check` — could not run locally (`gitleaks` missing); no auth, provider, channel, config, or credential examples changed
- [ ] SDK regenerated if server routes or schemas changed — not applicable, no route or schema changes
- [ ] `bun run localization:check` passes — not applicable, no product copy changed
- [x] Documentation, AGENTS, and `.synergy` help updated if developer workflow changed (architecture doc + decision record)
- [x] Material external sources are marked near the authoritative implementation and listed above


yzxoi:
**Follow-up from my review — CI blind spot: the end-to-end test added here never actually runs on CI.**

The real-helper e2e in `packages/runtime-local/test/sandbox/linux-readable-roots.test.ts` is the only coverage that exercises the full prepare → stage 1 → `bwrap` → stage 2 → child chain, including the two-stage profile re-read (defect #3). On every CI runner it silently skips, twice over:

1. **No helper at the expected path.** The `Test Shards (runtime)` job never installs the helper at `~/.synergy/sandbox-helper/synergy-sandbox-linux`, so the first guard returns early. (`runtime-artifacts` does build the helper, but that job doesn't run bun tests.)
2. **AppArmor blocks unprivileged userns.** `ubuntu-24.04` runner images ship `kernel.apparmor_restrict_unprivileged_userns=1`, so even with a helper present, the `bwrap --dev-bind / / /bin/true` probe fails and the test returns early again.

Both paths are indistinguishable from a pass in the shard summary. That is the same shape of gap that let this bug ship: CI was green on hosts where the skip fires, while sandboxed execution was broken on production-like hosts.

I'm pushing a fix to this branch:

- **CI (runtime shard):** install `bubblewrap`, load the AppArmor userns profile documented above for `/usr/bin/bwrap`, build + install the helper to `~/.synergy/sandbox-helper/`, and export `SYNERGY_TEST_LINUX_SANDBOX_E2E=1` for the test step.
- **Test:** when `SYNERGY_TEST_LINUX_SANDBOX_E2E=1` is set, every skip condition (platform, missing helper, failed bwrap probe) becomes a hard failure that includes the underlying reason — no silent green. Without the variable, local behavior is unchanged (honest skips, as today).

This also gives CI-hosted validation of the operator AppArmor profile documented in the PR description, so CI genuinely exercises the fix instead of skipping past it. (The separate readiness-probe gap noted in the description remains open as future work.)

yzxoi:
**Follow-up: enforcing the e2e immediately paid off — it caught a fourth latent defect.**

Two additional commits are now on this branch:

- `64922bd20` — the CI blind-spot fix announced in my review comment: the runtime shard now installs bubblewrap, loads the AppArmor userns profile for `/usr/bin/bwrap`, builds and installs the helper, and sets `SYNERGY_TEST_LINUX_SANDBOX_E2E=1` (declared in the turbo `test` task `env` for strict-mode pass-through). Under that flag every e2e skip becomes a hard failure naming the missing prerequisite. Also fixes the `packages/harness` → `packages/runtime-local` test-path typo in this description.

- `c787a4fc8` + `f91b5a36f` — what the enforced e2e caught on its very first real run: `backend.test.ts`'s OS-execution tests (never executed before on Linux CI) failed with `exit 1` and empty stdout. Root cause, verified in a privileged Ubuntu 24.04 container (before: both tests fail; after: both pass): a helper resolved under the host tmpdir — which is where `SYNERGY_TEST_HOME` puts the test home, so `tryInstallCargoHelper` stages the CI helper under `/tmp` — sits under the plan's final controlled-`/tmp` bind, which shadows every host path under `/tmp`. Stage 2 re-execs the same absolute path inside the sandbox and dies with `execvp ENOENT` before any command runs. Production never hit this because its helper installs under the real home.

  The fix mirrors the profile staging this PR already introduced: `LinuxBackend.prepare` copies a verified tmpdir-resolved helper into `~/.synergy/cache/synergy-sandbox/` (the same shadow-free directory), execs that copy, and binds the exec directory as a read root — the backend owns the two-stage re-exec contract, so callers never need to know about the helper. Production installs run in place, unchanged. Decision record and `execution-boundaries.md` updated; a unit test pins the staging contract.

The readiness-probe gap noted in the description remains open as future work.

yzxoi (approved):

PR #1418, opened 2026-09-18
fix(tool): unify the timeout contract and stop the countdown from lying

EricSanchezok:
## Problem

Tool-timeout information was produced by the runtime and rendered by the shared UI through two hand-maintained, unlinked definitions (`packages/harness/src/tool/timeout.ts` and a weaker local copy in `packages/ui/src/components/tool/timeout.ts`). Nothing tied them together, and no decision record had ever covered the contract. Four user-visible defects shared that cause.

**The bash metadata read argument names the model cannot send.** `operationForTool` read `args.backgroundAfterSeconds` and `args.timeoutSeconds`, but the model-facing schema is `.strict()` and exposes only `yieldSeconds`; the `backgroundAfterSeconds` mapping happened *after* metadata was computed. A call carrying `yieldSeconds: 300` produced `displayMs: 30_000`. Live session data confirmed it: input `yieldSeconds: 300`, stored `displayMs: 30_000`, actual runtime 253s — the badge stalled at `0s timeout` while the command ran 223 more seconds. The prompts advertised those same absent names, so following the documented instruction was a hard validation error.

**The countdown fabricated its own anchor.** `Countdown` used `Date.now()` captured at component construction whenever `startedAt` was absent. A countdown is a claim about elapsed time; mount time is not that claim's origin. Any remount — session switch, collapse/expand — restarted a 300s window from full. The clearest case was the question card, which never passed an anchor despite `request.createdAt` already being in the payload, resetting an hour-long default on every toggle.

**The label described the wrong event.** Renderers read only `displayMs`; `source` had ten members and zero consumers. An `auto_background` window hands the process to the background and keeps it running, yet it rendered as `Ns timeout`.

**Expiry kept rendering a live-looking number.** The value clamped to zero and kept saying `0s timeout` forever, distinguishable from a settled tool only by color.

## Fix

- **One contract.** `ToolTimeoutSource` and `ToolTimeoutMetadata` are defined once in `packages/util/src/tool-timeout.ts` and imported by both sides. `util` is the only package both `harness` and `ui` already depend on, and the UI may not import runtime-private modules; adding a dependency would have inverted the boundary. Field names keep their `Ms` suffix and millisecond units because the object is persisted on parts — renaming would strand the observation on all existing history.
- **Honest anchors.** `toolCountdown` returns `undefined` without `time.start`, and `Countdown` renders nothing without a finite anchor; the `Date.now()` fallback is deleted rather than guarded. `pending`/`generating` carry no `time`, and those states showed no countdown before, so honest absence costs no real information.
- **Source-aware labels.** `source` maps to a `kind` (`auto_background` → "to background", `tool_timeout` → "to timeout", everything else — including missing/unknown sources from historical parts — → "remaining"), with terminal labels on expiry (`backgrounded` / `timed out` / `past limit`).
- **Question cards** pass `startedAt={request.createdAt}`.
- **`timeoutSeconds` removed end to end** — schema, `BashParams`, the local executor's command timer, the remote strip list, and all three prompts. No model could reach it and the remote path already discarded it, so nothing in use was removed. The 24h hard ceiling and the abort path are unchanged: no enforced limit was weakened.
- **Seconds at the boundary.** Non-generated time-budget tool parameters renamed to `timeoutSeconds` with second-valued bounds and single-point conversions. Dead `countdown`/`countdownStartedAt` props removed, the timer now stops while inactive, and the bash description's default was corrected from 10 to the actual 30 seconds.

## Compatibility

- **No generated contract changed.** `state.metadata` remains an open record, so `packages/sdk/openapi.json` and `packages/sdk/js/src/gen/types.gen.ts` are byte-identical. `settleTimeoutMs` (`BrowserControlRequest`) and agenda `timeout` (`AgendaCreateInput`) keep their millisecond contracts.
- **Historical parts still parse**; missing or unknown `source` degrades to `remaining`.
- Fewer badges render than before — exactly the ones that were fabricated. Old parameter names now fail validation explicitly; no aliases were added.

## Verification

- Failing-first tests for each defect; the DOM regression test fails on **all 6 assertion bodies** against the pre-fix `HEAD` revision (it renders `0s timeout` and a fabricated `29s timeout`), and passes after the fix. New: `packages/ui/test/components/countdown-anchor.dom.test.ts` (8 tests, real Vite/JSDOM render, real resolved English strings).
- `packages/harness` `test/tool/` 55 pass; `test/session/` pass; `packages/runtime-local` tools 200 pass; `packages/connections` bash 46 pass; `packages/product-runtime` processor 53 pass; `packages/ui` suite and `apps/web` suite exit 0.
- `quality:quick` 16/17 gates pass (`secrets:check` fails only because `gitleaks` is not installed locally; CI runs its own secret-scan job), plus `doc:gen:check`, `coverage:check`, `test-layout:check`, `decision:check`, `localization:check`.
- Docs: two decision records (architecture + bug-fix), `docs/architecture/session-and-messages.md`, `.synergy/skill/add-tool` rules, `packages/util/AGENTS.md`, regenerated `docs/reference/tools.md`, and zh-CN translations for the six new labels.

## Not in scope

`browser_action`/`browser_navigation` `settleTimeoutMs` and agenda `timeout` stay in milliseconds: both units are already published in `openapi.json`, and agenda's value is persisted, so changing them would require a contract regeneration and a migration. They are deliberate exceptions, recorded in the decision records.


yzxoi (approved):
Reviewed the shared-timeout contract change end to end (util contract, harness resolver, UI renderers, tool schemas, prompts, decision records) and verified locally on the updated head: packages/ui typecheck is clean, the countdown/tool-timeout suites pass (13 tests across the unit and DOM files), and prettier is clean. The one Oryn finding — the decision record claiming the UI cast was gone while the cast remained — is now fixed on this branch: toolCountdown takes the shared ToolMetadata shape and the cast is truly deleted, with the unit test feeding historical metadata shapes through an explicit bypass helper. The remaining watch items (remote-link yieldSeconds clamping display, stale cached tool descriptions hitting strict schemas) are pre-existing or rollout notes, not merge blockers. Merging once CI is green.

PR #1419, opened 2026-09-18
fix(web): make sidebar session state independent of Scope store eviction

EricSanchezok:
## What does this PR do?

Makes the Web sidebar's session-row state independent of per-Scope store eviction, and replaces the duplicated session-status classifiers with one shared decision.

Two carriers take over the inputs that used to come from an evictable store:

- **Identity** (Blueprint binding and phase, workspace type, workflow kind and activity, parent, category) travels on `SessionNavEntry`, which is paginated, persisted, and refreshed by `session.updated`.
- **Runtime state** (status, pending permissions and questions) moves into the always-present global store, keyed by session id.

`resolveSessionVisualState` becomes a pure function over those two, the per-Scope `session_status` / `permission` / `question` slices are deleted, and `apps/web/src/utils/session-status.ts` owns the single classification every display surface reads.

## Why?

Session rows flickered while switching sessions or projects: a running spinner collapsed into a generic glyph, a returning spinner lost its Blueprint or worktree identity, then the identity came back.

The cause was one data dependency. `resolveSessionVisualState` resolved every rich state inside an `if (store)` gate, where `store` came from `peekScopeState(scopeKey)` — a per-Scope store that is evicted once enough Scopes are inactive. Re-creating it produced an empty shell whose bootstrap waited behind a two-slot request queue, so for a window the row had no status, no Blueprint binding, and no workspace type. Because events recreate an evicted store, the cycle repeated for as long as the session kept working.

The same gate made state unreachable cross-project for a second reason: status, permissions, and questions were stored per Scope even though every consumer reads across Scopes, so surfaces that render many sessions at once (sidebar, mobile drawer, Kanban board, status bar) showed no running or waiting state outside the currently leased Scope.

Three smaller defects shared one root cause — duplicated classification:

- `recovering` had no sidebar branch, so it read as idle there while the status bar and Kanban treated it as work.
- `retry` rendered the running glyph in the sidebar and the retry glyph in the status bar.
- Kanban treated a **missing** status as working, because `undefined?.type !== "idle"` is `true`.

None.

## How was it tested?

Verification follows the same shape as the change: pure classification, then resolution, then the seam between them.

**Unit and integration suites**

| Suite                                        | Result                           |
| -------------------------------------------- | -------------------------------- |
| `apps/web` full suite via the package runner | 2147 pass, 0 fail                |
| `packages/ui`                                | 50 pass, 0 fail (touched suites) |
| `packages/harness` session suite             | 843 pass                         |
| `packages/workflows`                         | 169 pass, 0 fail                 |
| `packages/server` global routes              | 25 pass, 0 fail                  |

**Regression assertions that target the defect directly**

- The resolver keeps identity with **no runtime context at all** — Blueprint binding, worktree, and child rows still resolve their own glyph. This is the assertion the old store-gated implementation could not satisfy.
- Status, permissions, and questions survive evicting the Scope store that carried them.
- `recovering` renders as a distinct intervention state rather than idle.
- The same `(status, waiting)` yields the same activity on the session row, the Kanban header, and the status bar.
- Both nav producers project identical identity fields for the same session, and a migration backfills an index written before the fields existed.
- The cross-Scope status snapshot reports sessions from more than one Scope, and a live status is never overwritten by a recovered one.

**Repository gates**

`bun run quality:quick` passes all gates that run locally. `secrets:check` requires the `gitleaks` binary, which is not installed on this machine; CI's own secret-scan job covers it, and the diff contains no secrets.

**Manual review**

Not performed. No isolated second runtime was exercised, so the end-to-end flicker fix is verified by the assertions above rather than by observing the sidebar. Worth a reviewer's visual pass: switch between two projects that each have a running session.

## Checklist

- [x] `bun run package:check` passes if package exports, build output, release logic, or publishable packages changed
- [x] `bun run workflow:check` passes if `.github/workflows/**`, GitHub Actions config, or CI helper scripts changed
- [x] Browser capability or App bootstrap changes pass the browser crypto contract and private HTTP browser smoke
- [x] `bun run secrets:check` passes locally or CI Gitleaks covers the change if auth, provider, channel, config, or credential examples changed
- [x] SDK regenerated (`./script/generate.ts`)
- [x] Documentation, AGENTS, and `.synergy` help updated
- [x] Material external sources are marked near the authoritative implementation, or the section says `None`

## Notes for the reviewer

- The cross-Scope status operation is `global.session.statuses`, not `global.session.status`. The SDK generator keys a method by the operationId with a leading `global.` stripped, so the latter would collide with the scoped `session.status`; the generated method then silently retargets one route and the other disappears from the SDK — with no build error. Regeneration confirms both methods now exist and target their own paths.
- The global index is seeded at cold load from the bootstrap snapshot and the already cross-Scope permission/question routes. Events alone would render an already-running session as idle until its next transition, and `recovering` would never appear at all, since nothing publishes it on the bus.
- `adoptScopeStatusSnapshot` keeps the per-Scope bucket's deletion semantics: a Scope's bootstrap page is authoritative for the sessions it owns, so a status left stale by a missed `idle` cannot survive a failover resync. The owned set is the raw page rather than the archived-filtered list, so an archived session's stale entry clears too.
- Merging `dev` changed retention semantics (a released lease now demotes into the inactive LRU instead of evicting), so one context test forces eviction by overfilling the LRU and asserts that precondition rather than assuming it.

## CI

All 40 checks pass.

CI caught one regression this work introduced, now fixed: binding a BlueprintLoop reviewer denormalizes the loop's presentation phase onto both bound sessions, so the audit session carries `phase: "auditing"` alongside its loop id and role. The workflows suite asserts that shape, but a `packages/product-runtime` tool assertion still compared against the pre-change value. The earlier targeted run had covered `product-runtime/test/session/` but not `test/tool/`, which is where the stale assertion lived.


yzxoi (dismissed):
Reviewed the full diff and the eviction-independent design. The two-carrier split (identity on SessionNavEntry, runtime state in the always-present global index) removes the store-gate defect mechanism, and the cross-surface consistency matrix pins sidebar/Kanban/status-bar agreement. Both low findings (the optimistic nav patch cannot clear identity; the wholesale runtime-index reset during reconnect) are self-healing and documented as deliberate trade-offs in the decision record. Merging.

yzxoi (approved):
Follow-up review of the two waves added after my first pass. The Cortex move is coherent: the task list now rides the global runtime index with whole-bucket replace semantics plus per-task post-stamp precedence (GlobalRuntimeWriteTracker.cortexWrite/cortexReplace/mergeCortex), the per-Scope slice is gone, and both nav producers project identity from the same deriveSessionIdentity. The Light Loop follow-up fixes the clearable-identity gap correctly — pairing the nav entry's workflow.active with the session record's presence is the right answer given both nav merges treat an absent key as no-update, and the decision record now states that principle for future carriers. New coverage on both waves (subsession pulse, snapshot/event race, light-loop-control matrix) targets the changed behavior directly. Merging on green CI; the reconnect-blank trade-off stays a release-note item.

PR #1420, opened 2026-09-18
perf(agent): derive worker capacity from memory and add priority lanes

EricSanchezok:
## What does this PR do?

Makes the Agent execution layer adaptive to real resource limits, adds priority admission between interactive and background turns, and removes avoidable work from the session-switch, per-round preparation, and telemetry write paths.

Ten independently reviewable changes:

1. **Memory-derived Agent worker capacity.** `execution.agentWorkers` was a CPU-count constant (`Math.max(1, Math.min(4, availableParallelism() - 1))`), so one source constant sized the pool on every machine while the binding constraint is memory. `ServiceMemory.resolveMemoryBudget()` now walks the cgroup v2 ancestor chain, takes the minimum of `memory.max` and `memory.high` at every level, normalizes the `"max"` sentinel and unreadable values, and falls back to `os.totalmem()`. `defaultAgentWorkers()` derives `min(memoryCap, cpuCap, 64)`; `resolveAgentWorkerCapacity()` is the single solver both startup and reload use, returning `{ size, source }`. An explicit value still wins verbatim, including the benchmark-pinned `1`. On a 62.55 GiB host this derives 20 where the constant allowed 4.
2. **Priority lanes.** `AgentTurnInput.lane` is `"interactive" | "background"` and defaults to `"interactive"`. The pool drains interactive strictly first and keeps at least one slot for background work. `lane` is admission-time information and never enters the worker envelope or persisted state.
3. **`minIdle ≤ size` on both capacity paths.** `resolvedAgentWorkerMinIdle()` caps the warm reserve at the resolved capacity, so an explicit `agentWorkers: 1` with a larger reserve produces a constructible pair instead of one the pool constructor rejects at the first turn.
4. **Session-switch instrumentation.** The `phase` label was silently dropped for all 494 stored rows because the host enum did not list the real mark names. `required` also mixed session prerequisites with panel marks, so 48 of 85 samples measured the 5000 ms timeout rather than the switch. Both fixed, with an invariant test binding the mark names to the enum.
5. **Session-switch payload.** `GET /permission` takes an optional `sessionID`; the initial latest page is 100 instead of 200; the message bucket cap is 30 instead of 15.
6. **Per-round preparation reuse.** Instruction files memoize on the configuration revision plus a `dev:ino:size:mtimeMs:ctimeMs` fingerprint; tool availability is collected once per round instead of twice; `Experiment.apply()` memoizes its parse by input identity.
7. **Hot-path write amplification.** `PartWriteBuffer` prices each streamed delta incrementally and clones once at flush instead of re-serializing and deep-copying the accumulated part per delta. `RolloutJournal.write()` commits allocation, evidence, projection, and head in one transaction instead of two.
8. **Telemetry write volume.** Token apply/paint aggregate per flush batch instead of per delta; the permission mirror is behind a default-off switch; aggregated counters collapse across per-turn identity; log writes are batched with a shutdown flush and an INFO local default.
9. **Storage capacity governance.** Incremental auto-vacuum with an owning-domain migration, and budgeted retention with a retention floor and live-session protection.
10. **Reference-catalog generation.** The generator attributed a nested field's `.describe()` to its parent and probed constructors anywhere in an expression, so the published `execution` row carried a child's text and several rows reported the wrong type.

## Why?

Measured on a 62.55 GiB host: with the ceiling at 4, `agent.queue.wait` p50 reached 16.70 s and p90 36.10 s, `agent.queue.depth` p50 4, and turn start to first `llm.request` p50 22.98 s, while demand was unchanged (turns/min p50 16). Peak concurrent workers was exactly 4, so the ceiling was binding. Raising it to 12 through the already-live-applied `execution.agentWorkers` field dropped `agent.queue.wait` p50 to 0.02 s and p90 to 0.16 s, and turn start to first `llm.request` p50 to 0.93 s. A clean baseline before the incident measured `agent.queue.wait` p50 at 19 ms.

Idle write load measured 22.37 MB/s and 7,733 syscalls/s with no user interaction; `agent.sqlite` was ~3.34 GB growing ~17 GB/day with `auto_vacuum=0` and no `VACUUM` anywhere under `packages/harness/src/storage/`.

- AIMD, Vegas, and Gradient2 concurrency limiters (Netflix `concurrency-limits`, Envoy's gradient controller) were evaluated and rejected as the capacity mechanism: they signal on a latency baseline, and LLM turn latency is dominated by prompt length and provider scheduling rather than queueing. The rationale and the rejected alternatives are recorded in `docs/decisions/implemented/architecture/2026-09-18-adaptive-agent-pool-capacity-and-lanes.md`.
- SQLite `auto_vacuum` semantics (`PRAGMA auto_vacuum = INCREMENTAL` requires a full `VACUUM` to convert an existing non-empty database) constrain the migration shape in the storage decision record.
- cgroup v2 `memory.max` / `memory.high` semantics and the `"max"` sentinel follow the kernel's `cgroup-v2` documentation. Runtime APIs are deliberately unused because Bun returns `totalmem` from `process.constrainedMemory()` when unlimited, and Node's libuv reads only the leaf cgroup, missing a stricter nested ancestor.

## How was it tested?

Commands and observed results are listed in the per-change decision records under `docs/decisions/implemented/`. Summary of the gates run locally:

- `bun script/gates.ts ci-static` — all 16 gates.
- `bun run doc:gen:check`, `bun run doc:check`, `bun run decision:check`, `bun run test-layout:check`.
- `packages/harness`: `bun test test/execution/ test/process/ test/session/ test/config/ test/observability/ test/util/ test/permission/ test/storage/`.
- `packages/product-runtime`: `bun test test/runtime/reload.test.ts test/server/runtime-handle.test.ts test/session/`.
- `packages/server`, `packages/workbench`, `apps/web` suites for the touched surfaces.
- `./script/generate.ts` for the `/permission` query parameter; generated SDK and `openapi.json` are included.
- `bun run --cwd packages/harness typecheck`, `bun run --cwd apps/web typecheck`, and the root `bun run typecheck`.

Runtime acceptance (queue wait percentiles by lane, switch-duration p90, idle write rate) is measured by replaying the observability queries in `docs/operations/performance-observability.md` against a running instance.

## Checklist

- [x] Relevant narrow tests pass; commands are listed above
- [ ] `bun run workflow:check` — no workflow or CI helper changes
- [ ] Browser capability or App bootstrap changes — none
- [x] SDK regenerated for the `/permission` query parameter
- [x] Documentation, AGENTS, and `.synergy` help updated for developer workflow changes
- [x] Material external sources are listed above
- [x] No secrets, local auth files, unrelated cleanup, or redundant compatibility wrappers included


yzxoi (dismissed):
Reviewed the capacity solver, priority lanes, session-switch telemetry fixes, hot-path write reduction, and the new retention/vacuum machinery, and verified locally on the updated head: the harness config/storage/observability suites pass, typecheck is clean, and doc:gen:check stays fresh. The one actionable Oryn finding — the documented storage.retentionMs: 0 disable path was rejected by the positive()-only schema, leaving default-on irreversible pruning without a working off switch — is fixed on this branch: retentionMs now accepts 0 (integer, min 0), both canonical config schemas were regenerated through the release pipeline (exclusiveMinimum 0 -> minimum 0, single-line diffs), and new tests pin both schema acceptance and the effective() zero mapping. The remaining watch items (first-upgrade VACUUM duration, aggregated-counter attribution change, like-for-like switch-latency comparisons) are release-note material, not blockers. Merging once CI is green.

yzxoi (approved):
Re-approving after the branch update. The retentionMs fix stands as reviewed (min 0, regenerated canonical schemas, tests pinning schema acceptance and the effective() zero mapping). The dev merge resolves the one conflict with #1419 correctly: the per-session permission fetch keeps the sessionID server-side filter, and its filtered response seeds the global index through a new session-scoped seedSessionPermissions instead of the whole-bucket seedGlobalPermissions — the right semantics, since feeding a filtered response to the whole-bucket seeder would drop other sessions' pending requests. The post-stamp precedence is preserved for the session slice, and the page-limit DOM test still asserts the filtered call shape on the merged tree. Verified locally on the merged head: apps/web typecheck, prettier, and the dom test pass.

PR #1421, opened 2026-09-18
benchmark: strict-endpoint merge conditions, probe budget, and Qwen3.8-27B acceptance results + synergy optimization loop

yzxoi:
## Summary

Two parts:

1. **Evaluator changes** (commit 2dfbfa7c5): named `merge_system_messages` conditions on both the harness axis (Synergy provider option injection) and the model axis (gateway merges leading system messages after protocol bridging), plus an independent 300s probe diagnostic budget. Motivated by vLLM Qwen chat-template endpoints that reject multiple leading system messages ("System message must be at the beginning"). Decision records: `docs/decisions/implemented/architecture/2026-09-15-benchmark-synergy-merge-system-messages.md`, `docs/decisions/implemented/bug-fix/2026-09-15-benchmark-probe-diagnostic-budget.md`.

2. **Qwen3.8-27B acceptance run** (50 cells, native linux/amd64, run `20260918-134939-816369c3` on an alicloud host, endpoint via qworks bridge tunnel) + an ongoing **synergy optimization loop**: idea → trace → diagnose → fix → re-run, results reported per round in comments.

## Results (50/50 terminal)

| harness | pass/10 | timeouts | avg wall | avg turns | output tok/turn |
|---|---|---|---|---|---|
| pi | **4** | 2 | 12.8m | 33.1 | ~1.0K |
| deepseek | 2 | 2 | 18.6m | 49.6 | ~1.0K |
| opencode | 2 | 4 | 27.9m | 25.7 | ~2.1K |
| codex | 1 | 3 | 15.4m | 25.6 | ~1.8K |
| synergy-core | 1 | **6** | **42.7m** | 29.7 | **~2.7K** |

Task × harness (r=1 / time=out / wrong / comp+vm=verifier missing):

| task | synergy | pi | opencode | codex | deepseek |
|---|---|---|---|---|---|
| build-cython-ext | time | time | time | time | time |
| dasel | fail | **1** | wrong | wrong | wrong |
| fd (180m) | time | **1** | time | wrong | **1** |
| large-scale | **1** | time | **1** | **1** | **1** |
| openssl | comp+vm | **1** | **1** | comp+vm | comp+vm |
| polyglot | time+vm | comp+vm | time+vm | time+vm | comp+vm |

## Validity audit (external-factor check before optimization)

- **Verifier suppression: 26/50 cells (52%) ungraded** — the openssl (12/15) and polyglot (14/15) verifiers exceeded their own 15-minute budget on this host, across *all* harnesses (r=None, not 0). Effective graded set is 24 cells; the scoreboard understates everyone, and synergy's 3/3 fast openssl completions (2.4–4.6m, same tier as pi's 1.3–1.6m) went ungraded. Round results must be read on the graded subset.
- **Tunnel outage 17:39–18:20 (~41 min)**: the qworks bridge pod was rescheduled, killing the ssh -R tunnel. Trials 0006 (opencode/fd) and 0021 (synergy/dasel, also crashed with a session-recovery loop, exit 2) lost budget mid-flight; trial 0009 started after recovery and ran its full budget. All 52/1637 wire failures fall inside this window; zero after recovery.
- **Per-request decode asymmetry**: synergy ~30 tok/s vs pi ~50–70 tok/s on identical endpoint/model/params. Same output volume decodes at the same speed, so the gap is non-decode overhead per request (synergy replays ~65K-token contexts; this endpoint's prefix-cache hit is unobservable — ledger shows `cacheWrite: known 0`). This inflates synergy's wall-clock but does not explain output volume (below).
- **Output-volume finding (endpoint-independent)**: same-task pairs show synergy generates 2.2–5.5K tokens/turn vs pi's 0.6–1K. Wire capture attributes this to: 47K-char system prompt (pi: 2K), 3–10K chars/turn of thinking, essay-style narration (dasel: 5,042 chars/turn vs pi's 31), and a compaction storm on fd (history oscillates 11↔156 messages; the compaction template mandates a "### What was accomplished" essay each cycle).
- Controlled conditions otherwise: doctor 30/30 pre-run, identical gateway-frozen sampling params, identical tasks/verifiers/deadlines, ledger reconciled, native amd64 (no Rosetta).

## Optimization loop (target: best pass-rate × token-usage tradeoff)

- [ ] Round 1 (control): synergy baseline at iteration conditions (shorter deadlines for turnaround) — establish the comparison anchor
- [ ] Round 2: compaction template/threshold + per-turn narration discipline (trace-driven)
- [ ] Round 3+: single-turn output caps, reasoning replay policy, prompt slimming; each idea must trace to a observed failure pattern
- [ ] Final: full 50-cell × 5-harness matrix, compared against the baseline above

Each round's raw numbers (passes, timeouts, tokens/turn, wall) get reported as a PR comment with the run ID and the code commit that produced it.

## Test plan

- [x] `pytest benchmark/test` — 258 passed (preset launch contract covers both merge conditions and probe budget)
- [x] `ruff check` + strict `mypy` clean
- [x] `bun run decision:check` green
- [x] Live doctor 30/30 on the vLLM Qwen endpoint before the 50-cell run


yzxoi:
## Optimization loop, fix 0: verifier budget was never wired (explains the 26/50 ungraded cells)

**Round 1 control (commit 306cd1afb, run `20260919-010032-6751501e`) was aborted after 1/10 cells.** The one completed cell (synergy × openssl) reproduced the acceptance run's worst measurement defect: agent finished the task in 3.3 min, verifier timed out at exactly 15.0 min again — despite the iteration suite raising that task's `verifier_seconds` to 2400.

**Root cause (85e626c45):** the evaluator never passed each task's `verifier_seconds` into Pier's `VerifierConfig.override_timeout_sec`. Pier then falls back to `task.config.verifier.timeout_sec` — the upstream task's own 900s budget. The openssl and polyglot verifiers install their own toolchain (apt + uv download) and exceed 900s on this host for *every* harness, which is exactly the acceptance run's 26/50 ungraded cells (12/15 openssl + 14/15 polyglot, r=None across all five harnesses).

**Fix:** `task.get("verifier_seconds")` now flows into the verifier deadline; the preset launch contract pins the propagation (would have failed red before the fix — confirmed 6 failed → 0). Full gates: 258 passed, ruff, strict mypy.

Consequences for the record:
- The acceptance scoreboard in the PR description undercounts everyone; the graded 24-cell subset remains valid for comparison, but per-task conclusions for openssl/polyglot were ungraded, not lost.
- Round 1 control is relaunched on the fixed evaluator (fresh freeze). All round numbers from here on come from graded cells.

yzxoi:
## Optimization loop, forensics correction + Round 1 partials

**Correction to the earlier "compaction storm" claim.** Re-forensics on baseline trial 0009 (synergy × fd) with a reliable mtime timeline, filtering out the one title-generator aux call:

- Main loop = 142 requests. Exactly **3 sharp compactions**: history 38→5, 77→5, 156→5 messages.
- Each post-compaction request is captured on the wire: `[system 47.3K chars] + [original task] + [assistant compaction essay 8.2–11.7K chars, opening "### What was accomplished (completed work)"] + [user "Continue … <recovery-hint> This task is continuing after context compaction"]`. Post-compaction base ≈ 31K prompt tokens — of which the 47K-char system prompt is the bulk.
- Trigger points: ~84–99K API-reported prompt tokens. That is *below* the naive soft threshold for the declared 262,144 context ((262144−65536−margin)×0.85 ≈ 156K) — the runtime budgeter is compacting on an internal estimate that diverges ~1.6–2× from API counts (tokenizer mismatch is the prime suspect; the endpoint's cache accounting is unobservable so calibration data is partial). Each cycle costs one essay-sized generation and discards fine-grained context; 3 cycles over 185 min.

So the earlier mechanism claim stands in corrected form: it is not a storm, it is a **repeat compaction cycle whose trigger fires early and whose summary template mandates an essay** — both are synergy-side, both are fixable.

**Round 1 control partials (run `20260919-015157-144a2612`, fixed verifier wiring 85e626c45):**
- synergy × openssl: agent completed, **verifier graded, reward = 1** — this exact cell was ungraded in the acceptance run.
- pi × openssl: **reward = 1**.
- 4 cells still running (fd/dasel are the long poles at 3600s deadlines).

The verifier fix already converts the previously-dark openssl axis into signal. Flash-persona arm (347061cf7) launches right after the control completes.

yzxoi:
## Round 1 control complete (run `20260919-015157-144a2612`, commits 306cd1afb + evaluator 85e626c45)

**Grading: 10/10 cells graded** (5 per harness). The verifier wiring fix eliminated the acceptance run's 26/50 ungraded-cell problem — every cell now produces a verdict.

| harness | pass | graded | timeout | failed | ctok/req | dur/req | 
|---|---|---|---|---|---|---|
| synergy-core | 1/10 | 10/10 | 3 | 0 | **3,066** | **134.0s** |
| pi | 1/10 | 10/10 | 3 | 1 | 996 | 54.4s |

Per-cell: synergy openssl **r=1** (4.0m, 488 ctok/req), pi openssl **r=1** (2.8m), both polyglot timeouts, both large-scale timeouts, pi fd timeout (60m), synergy fd timeout (15 reqs × 237s/req burned the budget), synergy dasel completed-wrong (60m), pi dasel **failed** (endpoint/gateway anomaly — last wire request `status=failed`, not a harness failure).

**Endpoint degradation between baseline and R1 (wire evidence):** pi/fd 12.6s→33.2s per request; synergy/large-scale 47.5s→170.6s. Both harnesses lost cells they passed in the baseline (pi: fd, large-scale; the endpoint was simply much slower). **R1↔R2 are adjacent and comparable; R1↔baseline absolute numbers are not.**

**The control confirms the flash-arm target under matched conditions:** synergy generates **3.1× the output tokens per request at 2.5× the wall time** versus pi on the same endpoint, same model, same params (3,066 vs 996 ctok/req). The only synergy pass (openssl) is also its cheapest cell (488 ctok/req) — when turns are short, synergy is competitive; long-horizon tasks die by accumulated per-turn cost. Also notable: the baseline's dasel session-recovery crash did **not** reproduce (synergy completed in 60m with a wrong answer instead) — that failure mode is not deterministic.

**Launching Round 2 (commit 347061cf7):** single-variable arm, `agent: synergy-flash` (2.3K-char persona vs the 39.3K-char synergy-max persona), same suite/deadlines/model/params as this control. Hypothesis: instruction density drives per-turn thinking + narration volume; the flash persona should cut ctok/req toward pi's range and convert timeouts into completions.

yzxoi:
## Round 2 complete (flash persona arm, commit 347061cf7, run `20260919-041320-4438398c`) — hypothesis falsified

5/5 terminal, all graded. **Pass face unchanged vs R1**: openssl r=1 (4.9m), fd/polyglot/large-scale/dasel all timeouts (dasel flipped from R1's completed-wrong to timeout).

| task | R1 max-persona | R2 flash-persona | Δ ctok/req |
|---|---|---|---|
| openssl (both r=1) | 13 reqs, 488 ct, 17.4s | 8 reqs, 521 ct, 21.4s | +7% |
| fd (both timeout) | 15 reqs, 3572 ct, 237.4s | 17 reqs, 4167 ct, 211.1s | +17% |
| polyglot (both timeout) | 8 reqs, 3259 ct, 108.0s | 7 reqs, 2286 ct, 124.5s | **−30%** |
| large-scale (both timeout) | 7 reqs, 3993 ct, 170.6s | 5 reqs, 5079 ct, 239.1s | +27% |
| dasel (R1 wrong / R2 timeout) | 26 reqs, 4017 ct, 136.6s | 23 reqs, 2594 ct, 155.9s | −35% |

**Aggregate mean ctok/req: 2,929 vs 3,066 (−4.5%) — noise.** Endpoint drift this window was mixed (dur/req −11% to +40% by task); token counts remain the clean comparison.

**Diagnosis:** the event streams explain the mixed per-task deltas. On large-scale the flash arm used *fewer* turns (19 vs 29 events) yet *more* tokens per request — visible narration is not what shrank; on fd it used more turns (88 vs 78). **Per-turn output volume is dominated by the model's thinking behavior, not by the persona's prose style.** The 39K-char persona does not control how long Qwen3.8-27B deliberates per turn. Persona arm is closed as a dead end; the config remains available but is not the lever.

**Round 3 design (declaring before implementing, per loop discipline):** model-axis single-variable arm on the R1 control config — `chat_template_kwargs: {enable_thinking: true, thinking_budget: 1024}` on the synergy-core (max persona) harness. Endpoint probes (retested twice this window): default thinking burns the full 2,000-token cap; `thinking_budget: 1024` returns ~355 ctok; `enable_thinking: false` ~450 ctok. The knob works and the gateway already forwards arbitrary model parameters (`body.update(model.parameters)`); the only change needed is the evaluator's model-parameter allowlist + shape validation (test first, mirroring the existing `thinking` validation). Hypothesis: capping deliberation per turn cuts ctok/req 3–5× on the timeout tasks and converts at least the borderline timeouts (large-scale, polyglot) into completions at equal or better pass rate.

yzxoi:
## Round 3 complete (thinking-budget arm, commits 64ab95a10 + dfbfaa7e4, run `20260919-055359-62cfdba8`) — falsified at endpoint-enforcement level

5/5 terminal, all graded. **Pass face identical to R1**: openssl r=1 (6.3m), fd/polyglot/large-scale/dasel all timeouts.

| task | R1 (no budget) | R3 (budget 1024) | Δ ctok/req |
|---|---|---|---|
| openssl (both r=1) | 13 reqs, 488 ct, 17.4s | 14 reqs, 506 ct, 25.6s | +4% |
| fd (both timeout) | 15 reqs, 3572 ct, 237.4s | 14 reqs, 4312 ct, 256.5s | +21% |
| polyglot (both timeout) | 8 reqs, 3259 ct, 108.0s | 3 reqs, 5495 ct, 299.1s | +69% |
| large-scale (both timeout) | 7 reqs, 3993 ct, 170.6s | 3 reqs, 159 ct*, 399.1s | n/a (hang) |
| dasel (wrong / timeout) | 26 reqs, 4017 ct, 136.6s | 22 reqs, 2970 ct, 162.3s | −26% |

*large-scale R3: 2 of 3 requests were interrupted at the deadline with 0 ctok — one streamed for 19.5 minutes without completing (endpoint latency degradation this window).

**Decisive finding — the endpoint accepts but does not enforce `thinking_budget`:**
- `chat_template_kwargs` verified end-to-end: **14/14 upstream payloads carried it**, evaluator allowlist + shape validation working (64ab95a10).
- SSE dissection of the polyglot request that billed 16,190 ctok: **6,304 reasoning chunks, 54,378 chars of deliberation (~15K tokens)** — 15× the declared 1024 budget.
- The endpoint emits thinking in a non-standard `delta.reasoning` field (not `reasoning_content`), which is why the earlier small-probe "355 ctok" misled: tiny prompts just deliberate briefly anyway; the budget semantics are not implemented server-side.

**R4 (declaring, then launching immediately):** single-variable arm on the R1 control config with `chat_template_kwargs: {enable_thinking: false}`. Unlike the budget knob, `enable_thinking: false` is *actually enforced* — three non-stream probes returned reasoning=0 — and the risk-elimination probe for the agent workload shape passed: **streaming + tools + thinking-off returns correct `tool_calls` with `finish_reason=tool_calls` in 0.9s**. Zero evaluator changes needed (allowlist already covers the key after 64ab95a10); config-only, then the same freeze→doctor→resume chain. Hypothesis: with deliberation hard-disabled, per-request output collapses toward pi's ~1K range and the borderline timeouts convert to completions; the open question is whether solution quality (pass rate) survives without deliberation.

yzxoi:
## Round 4 complete (thinking-off arm, commit 9e870e6fe, run `20260919-071731-7d82ba6f`) — the speed/quality tradeoff is real

5/5 terminal, all graded. Knob verified enforced: 14/14 upstream payloads carried `enable_thinking: false`; sampled SSE responses contain **zero** reasoning-channel characters (the non-standard `delta.reasoning` field stays empty).

| task | R1 thinking (default) | R4 thinking-off | Δ |
|---|---|---|---|
| openssl | **r=1**, 13 reqs, 488 ct, 17.4s | r=0 completed, 14 reqs, **151 ct**, 8.6s | pass → fail |
| fd | timeout, 15 reqs, 3572 ct, 237.4s | r=0 completed, **119 reqs**, 304 ct, 17.0s | timeout → done (fail) |
| polyglot | timeout, 8 reqs, 3259 ct, 108.0s | r=0 completed, 38 reqs, 352 ct, 14.3s | timeout → done (fail) |
| large-scale | timeout, 7 reqs, 3993 ct, 170.6s | timeout, 115 reqs, 198 ct, 9.5s | still timeout |
| dasel | r=0 completed-wrong, 26 reqs, 4017 ct, 136.6s | r=0 completed, 147 reqs, 325 ct, 15.2s | still fail |

**Aggregate ctok/req: 3,066 → 266 (−91%).** Endpoint drift note: dur/req collapsed to 9–17s partly from tiny outputs; token counts are the clean signal.

**What the A/B isolates:**
- **The cleanest pair is openssl**: same task, same harness, only the knob differs. With deliberation the agent produced a passing self-signed-cert setup (r=1); without it the agent "finished" in 2.7 minutes and failed verification. That is a genuine quality flip, not grading noise (the verifier graded both rounds).
- **Thinking-off converts timeouts into completions but not completions into passes**: fd/polyglot/dasel all now finish inside the deadline, all score 0. fd shows turn-count explosion (119 requests vs 15) — without deliberation the agent thrashes in small steps.
- **Neither form dominates**: thinking-on = 1 pass + 4 timeouts + 3.1K ctok/req; thinking-off = 0 passes + 4 fast completions + 266 ctok/req.

Per the loop's decision discipline, this is where the two viable endpoint-specific configurations should go to the run owner rather than the loop picking unilaterally:

1. **Quality form (thinking on, current default)**: the only configuration that has produced a pass on this endpoint.
2. **Speed form (thinking off)**: −91% output tokens, 4/5 cells reach a terminal verdict inside the deadline; every verdict is currently a fail on these six tasks.

A third path exists but doubles cell count: split model profiles per task difficulty (think for hard tasks, no-think for easy) — feasible with the current evaluator via two variants in one matrix, at 2× the cells.

The optimization loop's finding stands regardless of the choice: **on this endpoint, per-turn output volume is dominated by enforced server-side thinking, and that single flag moves synergy's cost structure by an order of magnitude.** The 50-cell five-harness final validation awaits the form decision.

yzxoi:
## Final validation complete — 54-cell dual-form matrix, six variants (commit 95be7801f, run `20260919-084514-fdba4b2a`)

54/54 terminal, **54/54 graded** (zero verifier suppressions, zero evidence issues — the 85e626c45 wiring fix plus the iteration suite's 2400s verifier budgets held for the entire run). All six variants ran simultaneously on the same endpoint window, so variant-to-variant comparison is drift-free.

| variant | pass/9 | timeout | ctok/req | reqs/cell | total ctok/cell | dur/req |
|---|---|---|---|---|---|---|
| pi | **7** | 2 | 1,237 | 25.3 | ~31K | 33.1s |
| deepseek | **7** | 1 | 1,107 | 29.1 | ~32K | 24.7s |
| codex | 6 | 3 | 2,023 | 20.4 | ~41K | 58.9s |
| **synergy-think** | **4** | 5 | 3,004 | 13.7 | ~41K | 114.8s |
| opencode | 3 | 5 | 2,192 | 25.2 | ~55K | 49.2s |
| **synergy-nothink** | **3** | 1 | **244** | 60.3 | **~15K** | 5.7s |

**Synergy's position changed.** In the baseline 50-cell run (degraded endpoint window), synergy was last at 1/10 while pi scored 7. In this window synergy-think scores 4/9 — fourth place, ahead of opencode — while pi again leads with 7 and deepseek matches it. Part of the gain is the healthier endpoint window (deepseek also dipped relative to its baseline 9/10), so cross-run comparisons carry a window caveat; the same-run variant ordering is the reliable signal.

**The dual-form tradeoff, quantified at 54-cell scale:**
- **synergy-think** buys passes with deliberation: 4 passes, but the highest per-request output in the field (3,004 ctok/req) and the most timeouts (5) — long-horizon tasks still die by accumulated per-turn cost.
- **synergy-nothink** is the token-efficiency champion: ~15K output tokens per cell (pi needs ~31K, think ~41K) and 8/9 cells reach a terminal verdict inside the deadline — but it converts only 3 into passes. Without deliberation the agent thrashes in cheap turns (60 reqs/cell, 2.7× the field median).
- Normalized: passes per 100K output tokens — pi 2.5, deepseek 2.2, **synergy-nothink 2.3**, codex 1.6, synergy-think 1.1, opencode 0.6. The thinking-off form reaches pi-class token efficiency; the thinking form is what reaches passes.

**Recommendation (pass-rate × token-usage tradeoff):** on enforced-thinking endpoints like this one, synergy should ship with `enable_thinking` as an explicit model-profile axis, not a fixed behavior. Quality-critical deployments keep it on (the only form that has produced 4 passes here); cost-sensitive deployments turn it off and accept the pass ceiling, landing at pi-class token cost. The evaluator now supports declaring this per experiment (`chat_template_kwargs`, 64ab95a10), and the product's `parameters` pass-through makes it a config change, not a code change.

The four-round loop record: R1 confirmed the 3× output-volume gap vs pi → R2 falsified the persona hypothesis → R3 falsified thinking_budget (endpoint accepts, doesn't enforce) → R4 proved the thinking switch is the dominant lever (−91% ctok/req). This final matrix confirms both forms at scale with all five harnesses graded fairly.

**Deliverable evaluator changes worth landing on dev:** verifier-budget wiring fix (85e626c45), `chat_template_kwargs` parameter support with shape validation (64ab95a10), strict-endpoint merge conditions (2dfbfa7c5), probe diagnostic budget, and the iteration suite with raised verifier budgets.

yzxoi:
## Round 5 declared: reasoning-replay elimination (clear_thinking) + prefix-cache probe, with same-window paired control

**Gap decomposition from the final matrix (why 4 vs 7):** the deficit is not uniform. openssl: tied (2-3 passes each). large-scale: synergy wins. The entire net gap sits on the two long-horizon tasks — polyglot (1 pass vs 3; one synergy solve was graded r=1 but missed the 15-min deadline) and dasel (0 vs 1; pi completed in 37m). The mechanism is turn economics on a forced-thinking endpoint: mean dur/req 90.3s vs pi's 35.6s means ~39 turns per 3600s budget for synergy vs ~101 for pi. The dasel specimen is stark — synergy emitted 133K output tokens (1.9× pi) in only 27 turns (pi: 85 turns) and never converged. On tasks needing ≥60 turns, a 2.6× turn-speed deficit is structural.

**Levers already falsified:** persona (R2), thinking_budget (R3). **Levers proven real:** enable_thinking (R4, −91% ctok/req but pass ceiling). **Remaining untested levers, now being run:**

1. **`thinking.clear_thinking: true`** — an existing evaluator-supported parameter (64ab95a10 allowlist already validates it) that strips reasoning from replayed history. This attacks both the prefill side (synergy replays 3-10K chars of thinking per turn into context, compounding the ~65K-token prompt) and the reinforcement side (the model keeps seeing its own verbose deliberation in history). Single-variable arm, zero code change.
2. **Prefix-cache TTFT probe on the current endpoint** — identical 60K-token-context requests, cold vs warm, `max_tokens=1` so response time ≈ prefill. Decides whether the 90s/turn includes an unreported prefill tax (the endpoint's cache accounting is unobservable: `cacheWrite: known 0`). If warm ≈ cold, prefill is not the tax and decode is everything; if warm ≪ cold, caching is active and harness context growth is cheaper than assumed.

**R5 design (same-window paired control, 10 cells):** two synergy variants in one matrix on the iter suite — `synergy-plain` (R1 control profile) and `synergy-clear` (identical + `clear_thinking: true`), 5 tasks each, fd/dasel at 3600s. Variant-to-variant comparison is drift-free by construction. Hypothesis: clear arm cuts ctok/req materially below plain's ~3000 and converts at least one long-horizon timeout, without the openssl quality flip seen in R4 (deliberation still happens; only the replay is dropped).

yzxoi:
## Round 5 verdict: `clear_thinking` is ignored by the endpoint — A/B void, fix moves client-side

Forensics on the live R5 run (20260919-142829-48b9e60a): the `thinking: {type: enabled, clear_thinking: true}` parameter is carried on every upstream payload, but **replayed assistant history still contains full reasoning** (openssl: 3 messages / 8,880 chars of `reasoning_content`; polyglot: 13 messages / 37,704 chars). The model sees byte-identical context to the plain arm — same failure shape as R3's budget knob: this vLLM Qwen deployment silently ignores non-template thinking controls. The two completed clear cells (openssl 590, polyglot 904 ctok/req) are plain duplicates; the run's remaining cells now serve only as plain-variance measurements and are being desco​ped in favor of the actionable arm.

**Round 6 (declaring): client-side reasoning-replay stripping.** The replay policy is the harness transform's job, not the endpoint's. New provider option `stripReasoning` (modeled on `mergeSystemMessages`): when true, the transform drops reasoning content from replayed assistant messages (`reasoning_content` key and reasoning parts) before sending. Mechanisms attacked: per-turn input growth (up to ~10K tokens of replayed deliberation by end-of-task) and self-reinforcement of verbose deliberation. openssl pass preservation is the canary for the known risk (some models need prior reasoning for coherence — the paired arm measures exactly that). Evaluator gets the matching `strip_reasoning` named condition (same contract as `merge_system_messages`, 2dfbfa7c5), R6 runs as a same-window paired matrix (plain vs strip, 10 cells) once the code lands.

yzxoi:
## Round 6 verdict: paired comparison voided by tunnel outage — partial signals support the hypothesis, clean re-run (R6b) launching

**What happened:** the qworks bridge pod was rescheduled again mid-run (~17:20), severing the SII tunnel while 5 cells were in flight. The chain was stopped to preserve evidence; `resume` after restoring the tunnel reconciled the interrupted cells, but 6 of 10 cells have no valid terminal verdict (r=None or interruption-reconciled zeros). **No task has valid terminal grades on both arms — the paired pass comparison cannot be made from this run.**

Valid terminal cells: plain/openssl **r=1**, strip/fd r=0 (completed), plain/polyglot timeout, plain/dasel r=0. Everything else: killed mid-flight or outage-window artifacts.

**Partial signals before the outage (same-task, pre-contamination):**
- strip/openssl: 339 ctok/req vs plain/openssl 976 — **−65% output on the identical task**, direction consistent with the hypothesis (both arms measured; strip arm was killed before verdict collection, so no pass comparison).
- strip/fd: reached a completed terminal inside the 3600s deadline (plain/fd timed out in R1) — direction consistent, not a paired cell.

**Infrastructure note:** tunnel restoration this time bypasses qworks (whose device-flow credential expired awaiting browser authorization): a Mac-hosted `ssh -R` loop into the alicloud host now carries the endpoint (verified 200; `caffeinate` guards the tunnel for the run's duration). Same GatewayPorts pinning, same Squid semantics; the only change is which host holds the reverse tunnel.

**R6b (launching now):** identical frozen config (`qwen38-iter-r6.yaml`, evaluator 9162e58 unchanged), fresh run ID, tunnel pre-verified. The question stands: does client-side reasoning replay stripping cut per-request output toward pi's range while preserving openssl's pass — the one mechanism lever R1-R5 never cleanly measured.

yzxoi:
## R6b final: strip_reasoning is real, quality-neutral — and the loop converges (commits 16d9adbb7 / af00d1cce, run `20260919-171151-f6df71cb`)

10/10 terminal. Mechanism verified on the wire: strip-arm upstream replays contain **0 reasoning characters** vs the plain control's 10,116 (the R5 `clear_thinking` arm's 37,704-character replay is the counter-example that motivated this fix).

### Paired results (same window, interleaved)

| task | plain (control) | strip_reasoning |
|---|---|---|
| openssl | **r=1**, 7 reqs, 858 ct/req, 42.3s | **r=1**, 8 reqs, 983 ct/req, 39.7s |
| fd | r=0 timeout, 17 reqs, 4,762 ct/req, 211.2s | r=0 timeout, 13 reqs, 5,532 ct/req, 276.3s |
| polyglot | timeout (verifier_missing, 2 reqs stalled at 449s) | r=0 timeout, 11 reqs, 1,902 ct/req, 78.4s |
| large-scale | r=0 timeout, 7 reqs, 3,389 ct/req, 170.7s | timeout (verifier_missing, 3,456 ct/req) |
| dasel | r=0 timeout, 15 reqs, 4,345 ct/req, 239.3s | r=0 timeout, 23 reqs, 3,782 ct/req, 155.3s |

### Verdict — the hypothesis is falsified, and the loop converges

- **Quality preserved** (the canary): openssl passes on both arms — no R4-style flip. `strip_reasoning` is safe to ship as a transport option.
- **But replay stripping does not change the long-task economics**: per-request output on long tasks is flat-to-higher under strip (large 3,456 vs 3,389; fd 5,532 vs 4,762; dasel 3,782 vs 4,345). Every long-horizon cell still times out on both arms — unlike R4's thinking-off, which converted all of them.
- Combined with R2 (persona falsified), the six-round loop converges on one mechanism: **on this enforced-thinking endpoint, per-turn output volume is driven by the model's current-turn deliberation alone. Neither the persona nor the replayed history controls it.** The only measured output lever is the server-enforced `enable_thinking` switch itself (R4: −91% ctok/req), which carries a pass-rate ceiling on these tasks.

### Loop record

R1 target confirmed (synergy 3,066 vs pi 996 ctok/req) → R2 persona falsified → R3 `thinking_budget` accepted-but-unenforced → R4 `enable_thinking` real (−91%) with a pass ceiling → R5 `clear_thinking` accepted-but-unenforced → R6/R6b client-side `stripReasoning` real (0-byte replay, quality-neutral) but output-neutral on long tasks.

**Shippable outcome:** `strip_reasoning` lands as a declared, quality-neutral transport option for strict enforced-thinking endpoints (openssl-pair-proven), and `enable_thinking` is documented as the explicit cost/quality axis. Everything an operator needs to trade tokens for passes on such endpoints is now a config declaration, verified end-to-end.

yzxoi:
## Round 7 declared: pi-shaped tool surface (the R2 blind spot, now trace-identified)

**Trace evidence from the final 54-cell matrix (pi-pass vs synergy-fail, same tasks):**

| decode budget spent on | pi (dasel PASS, 85 reqs) | synergy (dasel timeout, 27 reqs) |
|---|---|---|
| reasoning | 147K chars | 495.7K chars |
| content | 5.8K | 12.7K |
| **tool arguments (actual work)** | **73.4K** | **5.2K** |
| tool calls | 86 | 58 |

pi's harness: **2,579-char system prompt + 4 tools** (read/bash/edit/write); per-request reasoning **p50 = 0** — most turns are pure tool dispatch. Synergy: 47K-char prompt + 22 tools; single-request reasoning peaks at 100K chars. On a forced-thinking endpoint, "22-way tool selection + giant schemas + orchestration rules re-read every turn" is exactly the decision material the server amplifies into long deliberation. R2's blind spot: it swapped the persona but kept the 22-tool surface.

**R7 design (same-window paired matrix, 10 cells):** both arms run the flash persona (`agent: synergy-flash`, 2.3K-char prompt) — arm A `synergy-flash` keeps the default surface (13 resident tools, R2 config), arm B `synergy-flash-core` adds one overlay declaring `agent["synergy-flash"].deferredTools` folding every non-core tool behind `expand_tools` (resident: **bash, edit, read, write, search_tools, expand_tools** — pi-shaped, nothing removed, only folded). Single variable = tool surface; the product's existing `deferredTools` mechanism with config-replace semantics makes this zero evaluator/product code. Model parameters identical to R6b (default thinking). Hypothesis: arm B's per-request output collapses toward pi's range (p50-style short turns), converting long-horizon timeouts while keeping pass quality.

yzxoi:
## R7 final + seven-round loop closure (commits 35ba74cab / af00d1cce, run `20260920-003024-552f8b33`)

10/10 terminal. Mechanism verified: flash-core's resident tool surface was exactly 6 (bash/edit/expand_tools/read/search_tools/write) vs the control's 13.

### Paired results (same window, interleaved; both arms flash persona)

| task | flash (13 tools) | flash-core (pi-shaped 6 tools) | pi reference (54-cell) |
|---|---|---|---|
| openssl | **r=1**, 9 reqs, 317 ct/req, 13.5s | **r=1**, 10 reqs, 708 ct/req, 21.4s | 3/3 pass |
| large-scale | **r=1**, 17 reqs, 2,050 ct/req, 61.9s | timeout, 12 reqs, 3,103 ct/req | 0 passes |
| dasel | timeout, 25 reqs, 3,435 ct/req, 143.1s | r=0 **completed**, **91 reqs**, 2,099 ct/req, **32.6s** | 1 pass (37m) |
| polyglot | r=0 timeout, 11 reqs, 1,955 ct/req | r=0 completed, **3 reqs**, 43.6s | 3 passes |
| fd | r=0 timeout, 63 reqs, 2,984 ct/req, 56.3s | r=0 timeout, 32 reqs, 3,688 ct/req | 1 pass |

### Verdict — the mechanism moved, the passes did not follow

- **The turn-economics hypothesis was mechanically confirmed on dasel**: the folded surface produced pi-style pacing — 91 turns (3.6× the control), per-request wall time −77% (32.6s vs 143.1s), per-request output −39%. The tool surface IS a lever on turn economics.
- **But it bought no passes and cost one**: dasel and polyglot flipped from timeout to fast-wrong completions; large flipped from pass to timeout. Pass face: flash 2, flash-core 1. Folding the surface trades deliberation depth for speed, and on these tasks the depth was load-bearing.
- **openssl stayed paired r=1/r=1** across all four quality-bearing arms (R1/R2/R6b/R7) — the only task where cost and quality were never in tension.

### Seven-round convergence

R1 confirmed the 3.1× output-volume gap vs pi → R2 falsified persona → R3/R5 falsified `thinking_budget`/`clear_thinking` (accepted, unenforced) → R4 proved `enable_thinking` is the sole output lever (−91%) with a pass ceiling → R6b proved client-side `stripReasoning` is real and quality-neutral but output-neutral → R7 proved the tool surface moves turn economics (pi-style pacing achieved) but passes still don't follow.

**Final conclusion:** pi's advantage on this enforced-thinking endpoint is not any single knob — not the persona (R2), not the reasoning replay (R6b), not the tool count alone (R7). It is the co-design of its minimal prompt, minimal tool set, and an agent loop whose task priors were built for that surface. Synergy's equivalent edge remains its quality-mode configuration: **thinking-on with the full surface is the only form that reached 4/9 in the 54-cell matrix**, and every cost lever measured (thinking-off, surface-fold) buys tokens by spending passes. The evaluator now ships all of these as declared, single-variable conditions (`strip_reasoning`, `chat_template_kwargs`, `deferredTools` overlay), so the cost/quality point on this curve is a config choice per deployment, with paired-measurement methodology to audit it.

**Evaluator changes worth landing on dev:** verifier-budget wiring fix (85e626c45), `chat_template_kwargs` support (64ab95a10), strict-endpoint merge conditions (2dfbfa7c5), probe diagnostic budget, `stripReasoning` transport option (16d9adbb7), and the iteration suite + seven presets.

yzxoi:
## Round 8 declared: core-surface discipline prompt — plus a product bug it surfaced

**R7 follow-up analysis falsified the "pi-style discipline prompt" transplant**: pi's per-request reasoning p50=0 is an emergent property of its tiny decision surface (measured: the R7 core arm already matched pi's decode economics on dasel — 75.5K vs 73.4K tool-arg chars, 91 turns vs 85 — without any prompt change), and pi's highest-frequency rule (multi-edit batching) is unportable (synergy's edit is single-replacement). Transplanting pi's prompt text imports nothing.

**But the R7 core arm's own failure modes are real and fixable**: polyglot "completed" in 3 requests with zero tool calls and no verification (fast-wrong), and fd's compaction amnesia (resumed from a summary misjudging implementation as not-started). Both are discipline gaps on the folded surface, and the flash persona still references folded tools ("web search") — an instruction/surface mismatch.

**Product bug found while implementing (test-first fix incoming):** a config `agent["synergy-flash"].prompt` override is silently discarded — the native prompt builders (`agent.ts`) unconditionally overwrite config-merged prompts for the three built-in primaries after the config merge loop. The merge loop's own `item.prompt = value.prompt ?? item.prompt` states the intended precedence; the builders violate it.

**R8 design (10 cells, same-window paired):** arm A `synergy-flash-core` = R7's exact folded-surface config (6 resident tools, built-in prompt) as in-window control; arm B `synergy-flash-core-v2` = same surface + a ~1.9K-char core prompt that (a) describes the actual six-tool surface without stale references, (b) adds completion discipline ("done" = acceptance executed on this machine, run the verification; few-steps attempts are a warning to re-verify), (c) adds resume discipline (compaction summaries are unverified claims; re-check disk/git/tests before redoing work). Requires the prompt-override fix deployed. Hypothesis: the fast-wrong and amnesia failure modes convert to verified completions on dasel/polyglot without regressing openssl.

yzxoi:
## R8 final + eight-round loop closure (commits 16d9adbb7 → 7f5cb7158, run `20260920-135555-15659766`)

10/10 terminal. Mechanism verified end-to-end: the core-v2 arm's upstream system prompt began "You are synergy-flash-core" (the `7f5cb7158` precedence fix delivering config prompts to the wire for the first time) with the 6-tool folded surface held constant.

### Paired results (same window, interleaved; both arms flash persona + folded surface)

| task | core (built-in prompt) | core-v2 (discipline prompt) |
|---|---|---|
| openssl | **r=1**, 9 reqs, 797 ct/req, 11.0s | **r=1**, 13 reqs, **343 ct/req**, **5.1s** |
| dasel | r=0 completed-wrong, 17 reqs, 4,101 ct/req, 61.5s | r=0 timeout, 30 reqs, 2,279 ct/req, 119.4s |
| polyglot | r=0 timeout, 14 reqs, 5,237 ct/req | r=0 timeout, 15 reqs, 3,243 ct/req |
| fd | r=0 timeout, 18 reqs, 2,396 ct/req | r=0 timeout, 27 reqs, 4,985 ct/req |
| large-scale | timeout (verifier_missing) | completed (verifier budget timeout, r=None) |

### Verdict — discipline prompt improved cost and behavior, not the pass ceiling

- **openssl**: both r=1; the discipline arm spent **2.3× fewer output tokens per request** (343 vs 797) at less than half the wall time — the "completion discipline" verifiably cuts cheap-answer thrash on short tasks.
- **polyglot**: turn pattern normalized (15 methodical turns vs R7's 3-turn zero-verification claim), but still timed out and scored 0 — deliberation happened, the solve didn't.
- **fd**: both arms graded **F2P 0/43** (P2P 109/109) — this window's fd was unsolved by every arm including the control; the resume discipline cannot be credited or blamed on a task the field failed. The control died in its classic exploration loop ("thorough understanding… let me check a couple more details before implementing"); the discipline arm did one compaction with no amnesia signature, but no implementation either.
- **dasel**: the discipline arm slowed (30 turns, 119s/req) instead of holding R7's pi-style pacing and lost the completed-wrong terminal — with n=1 per cell, run-to-run variance on this task is large enough that single-pair readings are indicative, not conclusive.

**Eight-round convergence.** R1 confirmed the 3.1× output-volume gap vs pi → R2 falsified persona → R3/R5 falsified `thinking_budget`/`clear_thinking` (endpoint accepts, doesn't enforce) → R4 proved `enable_thinking` is the sole output lever (−91%) with a pass ceiling → R6b proved `stripReasoning` is real, quality-neutral, output-neutral → R7 proved the tool surface moves turn economics (pi-style pacing) without moving passes → R8 proved discipline prompts cut cost and normalize behavior without moving passes.

**Final conclusion:** on this enforced-thinking endpoint, the pass ceiling for synergy is set by the model's domain reasoning quality, not by any harness lever measured. All harness levers tested move the cost curve (thinking-off −91%, tool-fold −39% per-turn with pacing gains, discipline prompt −57% on short tasks); none moved passes. Shipped as config-declared axes: `enable_thinking`, `strip_reasoning`, `deferredTools` folds, and now config prompt overrides that actually reach the wire — every cost/quality point on this curve is auditable with the paired methodology this loop used.

**Evaluator/product changes worth landing on dev:** verifier-budget wiring fix (85e626c45), `chat_template_kwargs` support (64ab95a10), strict-endpoint merge conditions (2dfbfa7c5), probe diagnostic budget, `stripReasoning` transport option (16d9adbb7), **agent config prompt precedence fix (7f5cb7158)**, and the iteration suite + eight presets.

yzxoi:
## ×4 clock-hypothesis verdict (commit 3d00b6b18, run `20260920-170058-192d6dd1`): partially confirmed — polyglot was pure budget; dasel and fd are not

5/5 terminal, synergy-think only, every deadline ×4 (dasel/fd 7200s, others 3600s), verifier budgets 2400s.

| task | 54-cell baseline (1×) | ×4 experiment | verdict |
|---|---|---|---|
| openssl | r=1 | **r=1**, 11 reqs, 364 ct/req, 5.6s | stable |
| polyglot | 1×(r=1 timed-out) + 2 timeouts | **r=1 completed**, 9 reqs, 5,215 ct/req | **clock-limited: budget fixes it** |
| dasel | timeout (no grade) | r=0 completed, **F2P 139/146 (95.2%)**, P2P 1012/1012, BINARY 0 | **not clock: 4× time and 408K output tokens still leave the same 7 spec tests failing** |
| fd | timeout (F2P 0/43) | r=0 timeout, F2P **0/43**, P2P 109/109 — 136 reqs, 385K ct over 2h | **not clock: zero must-pass tests reached with 2h** |
| large-scale | r=1 (12 reqs) | r=0 completed-wrong, 6 reqs | run variance (n=1; final text shows it still "verifying whitespace structure" in exploration mode — never wrote macros) |

### What the ×4 run settles

- **polyglot's timeout was pure budget** — the first direct confirmation that one of the five timeouts converts to a pass with more clock.
- **dasel is not budget-limited past ~95%**: with 2 hours and 408K output tokens, synergy-think landed at 95.2% F2P — statistically identical to what deepseek (95.2%) and synergy-nothink (91%) reached in a 60-minute window. The residual 7 tests (HTML entity escaping, script raw-text round-trip, complex implicit closing) are a spec-knowledge gap every harness except pi shares; pi cleared 146/146. More time does not buy knowledge.
- **fd is a capability ceiling**: 136 requests over 2 hours and not one must-pass test graded. The whole field failed it at 1×; 4× changes nothing.
- **large-scale n=1 regression** is variance, not the clock: the arm stalled in exploration at request 6.

### Corrected conclusion for the loop

The five timeouts decompose into: **1 budget-fixable** (polyglot), **1 partially budget-responsive but binary-capped** (dasel — more time raises completion 91→95% yet still scores 0), **2 capability-capped** (fd, and dasel's last mile), **1 variance** (large). Synergy-think's effective pass rate with a 4× clock is ~3/5 tasks (vs 2/5 at 1×), at 4× the wall-clock cost — while pi's 7/9 at 1× remains the reference. Turn economics explain the timeouts; only polyglot's was purely economic.

PR #1422, opened 2026-09-18
fix(web): derive model readiness from the provider store

EricSanchezok:
## Summary

The Web shell showed an "AI model not configured" strip whenever the startup health probe reported `modelReady: false`, and it stayed after the user configured a provider in Settings.

Root cause: the signal was cached, not derived. `apps/web/src/context/server.tsx` wrote `modelReady` in exactly one place — the health effect, which re-runs only when the active server URL changes or `refreshToken` increments. Neither `refresh()` caller runs when a provider is configured. Configuring a provider refreshes a different store: `globalSync.refreshProviders()` resolves to `refreshTargeted(["provider"])`, which writes only the `provider` / `provider_auth` buckets. The `config.updated`, `runtime.reloaded`, and `provider.auth.updated` handlers never re-read health. Desktop has no health polling at all, so its only recovery was a full page reload.

A cold-start false negative compounded it: `Server.resolveHealthModelReady` races the provider build against a one-second window and falls back to `lastSettledProviders`, which is `undefined` before the first build, so it can answer `false` while a model is configured. The frontend cached that `false` permanently.

## Approach

Model readiness is now derived from the already-reactive global provider snapshot instead of a cached probe, which removes the class of bug rather than the instance.

- **`apps/web/src/components/provider/model-readiness.ts`** (new) exports `resolveModelReadiness(provider)`, a pure function whose result is a discriminated union. `ready` means some provider in `connected` has `runtimeAvailability[providerID].available === true`. The three blocking reasons are distinguished: `not-configured` (nothing connected), `needs-attention` (auth health is `action_required` or `exhausted`), and `restricted` (blocked by `disabled`, `no_models`, or `not_connected` only). Evaluation order is `ready` → `not-configured` → `needs-attention` → `restricted`, and the blocking states carry sorted `providerIDs` for deterministic focus.
- **`apps/web/src/components/app-shell/model-unavailable-banner.tsx`** (new) renders from that derivation, reading the global snapshot only. It is a full-width strip placed after `DesktopNativeTitlebar` and before `ConnectionBanner` inside `LayoutContent`, which sits inside `GlobalSyncProvider`. That provider gates its children on `ready` and populates `provider` before flipping it, so the strip cannot flash during load. The strip keeps the existing warning tokens and `⚠` glyph, carries `role="status"` / `aria-live="polite"`, and exposes a keyboard-reachable action that opens `SettingsDialog initialTab="providers"`, passing `providerFocusID` only when exactly one provider is affected. The terminal command is demoted to secondary body text.
- **`apps/web/src/app.tsx`** loses `ModelReadyWarning` and `showModelReadyReadyWarning`; **`apps/web/src/context/server.tsx`** loses the dead `modelReady` signal, its reset, its assignment, and its returned field. `healthy`, `refresh()`, `startupView`, and the health effect are otherwise intact.

The server contract is untouched: `GET /global/health` still returns `modelReady`, `Server.resolveHealthModelReady` keeps its bounded wait for the CLI and daemon consumers, and no route or generated SDK artifact changed.

## Verification

Automated:

- `bun test test/components/provider/model-readiness.test.ts` — 11 cases covering `ready`, the default-empty `not-configured` shape, `needs-attention` for both `action_required` and `exhausted`, `restricted` for `disabled`, `no_models`, and a missing availability entry, precedence when connected providers disagree, and deterministic `providerIDs` ordering.
- `bun test test/components/app-shell/model-unavailable-banner.test.tsx` — 12 Chromium cases covering the env-only provider producing no banner, all three reason copies with their action labels, distinct copy per reason, live clearing with no reload, re-appearance on regression, Settings opening at the provider section with and without a focus target, polite announcement plus keyboard focus, and 375 px layout.
- `bun run --cwd apps/web test`, `bun run --cwd apps/web typecheck`, `bun run --cwd apps/web build`.
- `bun run --cwd apps/web i18n:extract` (leaves catalogs unchanged on re-run), `bun run localization:check`, `bun run test-layout:check`, `bun run decision:check`, `bun run doc:check`, `bun run quality:quick` (all 17 gates).
- `bun test test/server/health-timeout.test.ts` from `packages/server` — unchanged and passing.
- `git status` shows no generated contract files touched.

Behavioral, on an isolated runtime:

- Cold start with no provider reproduced `modelReady: false` with `connected: []`, and the strip rendered the `not-configured` copy with its action and the terminal hint as secondary text.
- After a provider became usable, the strip cleared with the page surviving (no reload). The check used a DOM marker on `documentElement` that persisted across the transition, so a reload would have been detectable.
- The action opened Settings at the Providers section in the real app, listing provider connections.
- A live runtime snapshot confirmed the exact derivation input: `connected: ["deepseek"]` with `runtimeAvailability.deepseek = { available: true, modelCount: 4 }` yields no banner.

## Documentation

- New implemented decision record at `docs/decisions/implemented/bug-fix/2026-09-18-model-readiness-from-provider-store.md`, recording the probe-caching defect and the rejected alternatives (explicit invalidation of the cached probe, health polling, fixing only the server-side bounded wait, reusing the per-scope provider store, lifting readiness above the sync provider, and the two declined product alternatives).
- `docs/architecture/frontend-data-sync.md` gains an invariant that model readiness derives from the global provider snapshot and that the health probe is a startup/diagnostic signal rather than a UI readiness source.
- `apps/web/PRODUCT.md` gains the durable product rule: the model-unavailable strip is actionable, opens provider Settings, clears live, and never depends on a one-shot startup probe.

## Notes

- `apps/web/script/test.ts` registers the new banner suite in `playwrightIsolated` because it launches Chromium.
- Provider failure codes deliberately stay out of the strip; they remain in Settings diagnostics.


yzxoi (approved):
Reviewed the provider-store derivation, bootstrap ordering, Settings navigation and live banner updates. The disabled-provider wording is a documented presentation choice and does not change availability. Validation on the updated dev base: 23 targeted readiness/banner tests passed, including real DOM updates, keyboard access and 375 px layout; all 28 workspace typechecks and pre-push checks passed.

PR #1423, opened 2026-09-18
fix(mcp): repair built-in MCP settings state, status copy, and key feedback

EricSanchezok:
Fixes three defects on the Settings → MCP panel for the built-in `anysearch`/`scholight` servers.

## 1. Phantom "unsaved" state after saving a built-in key

`buildMcpPatch` hoisted type-less stubs ahead of typed user servers, then decided whether to emit a patch with `JSON.stringify(newMcp) !== JSON.stringify(cfg.mcp ?? {})`. The stored domain file keeps its own key order, and the write path (`mergeDeep`, then `sortConfigKeys`, which orders only top-level domain keys) never reorders keys inside `mcp`. So once a config had at least one typed server **and** at least one built-in stub, the comparison reported a change on every save while the content was identical.

Effect: the `未保存`/`Unsaved` badge persisted after a successful save, closing Settings always raised the discard confirm, and each save re-wrote the `mcp` domain and triggered a redundant MCP reload.

Fix: compare structurally with `isDeepEqual` (`remeda`, already a dependency). The emitted key order is deliberately unchanged — order is not part of the stored semantics, and reordering the builder would couple it to a writer-side detail that `mergeDeep` may legitimately change.

Regression tests cover both mixed-order no-op cases plus an end-to-end test that drives the real `ensureInit` hydration path with the exact four-entry config shape from the report, and controls asserting real edits (toggle, rotate, clear) still emit their stub.

## 2. Switch on but label reads "Unavailable"

The switch came from `status !== "disabled"` while the label mapped only 2 of 11 `MCP.Status` members (`connected`, `disabled`) and collapsed the other nine onto "Unavailable" — so "on + unavailable" was structural. The panel also never refreshed status after its one-time fetch, and the MCP dialog already had a complete mapping that the panel duplicated differently.

Fix: `apps/web/src/components/mcp/status-presentation.ts` is now the single authority, consumed by both surfaces. Every status maps to a label, description, and tone; failure/auth states additionally show their error text. The built-in switch is seeded from configuration intent (`builtinServerEnabled`) rather than a live connection string, and `SettingsPanel` refetches on `mcp.*` events so status stays live while the panel is open. Custom server cards get the same treatment, since they had the same switched-on-but-not-connected blind spot behind `Enabled`/`Paused`.

## 3. Weak feedback after pasting a key

A stored key was nearly invisible: the client only had `keyConfigured: boolean`, the save path reset the draft so a configured field looked identical to an unconfigured one, and the configured placeholder read "Key set — paste a key to raise rate limits".

Fix: `builtinApiKeyHint` exposes a server-computed `keyHint` (four-dot mask + last four characters; bare mask for keys ≤ 8 characters; absent when none is stored), and the panel shows a three-state chip — `No key set` / `Key saved ••••1234` / `Not saved yet`. The configured placeholder now reads as a replacement prompt.

**Disclosure note:** `keyHint` intentionally reveals four trailing characters of a stored key. Masking is computed server-side because client read-backs only ever contain `REDACTED_SENTINEL`. It travels only through this field and is never logged. `/mcp/builtins` was deliberately *not* gated behind `requireLocalhost`, because that middleware requires an `http(s)://localhost` origin and would risk breaking desktop/webview origins for a four-character addition. The exception is documented in the `synergy-config` skill's MCP reference.

## Verification

- `useConfigPatch` suite: 83 pass / 0 fail, including the order-insensitivity regression and the end-to-end hydration reproduction
- `status-presentation`: 8 pass / 0 fail — a specific descriptor asserted for every `MCP.Status` union member; no member collapses onto a generic label
- `builtin-key-presentation`: 9 pass / 0 fail — all four chip states distinguished
- `agent-integrations` MCP suite: 59 pass / 0 fail, including four `builtinApiKeyHint` cases
- `bun run --cwd apps/web typecheck`, `i18n:check`, `decision:check`, `doc:check`, `test-layout:check`, Prettier: all green
- OpenAPI document and generated SDK types regenerated for `keyHint` (not hand-edited)

Decision record: `docs/decisions/implemented/bug-fix/2026-09-19-builtin-mcp-settings-state-and-feedback.md`. The owning built-in-search feature record was updated to describe `keyHint` as shipped.


yzxoi (dismissed):
Reviewed the MCP settings state, status events, builtin key feedback, and generated API boundary. Config comparison and switch intent preserve edits without accidental toggles. Validation: 102 pure Web tests, 16 browser-condition render tests, 14 builtin catalog tests; all workspace typechecks and pre-push gates pass.

PR #1425, opened 2026-09-18
perf(storage): cut the JSON-to-SQL upgrade wall's hot path

yzxoi:
The first JSON-to-SQL upgrade serializes backup compression and independent session migrations, increasing startup time on large homes. This change pipelines up to eight compression flights, drains them in source order, and runs independent session migrations with bounded concurrency.

Retirement still verifies the full digest of every source and backup file before deleting originals. Equal-size replacement can change content, so size checks or sampled digests cannot establish safe retirement. A regression covers equal-size corruption outside the former sample positions.

Validation: 17 focused storage tests with 2,204 assertions, all 17 quality gates, and the full workspace typecheck/pre-push checks pass. The earlier benchmark figures in the decision record describe the sampling prototype; they are not measurements of this final implementation. The PostgreSQL and installed-runtime matrices remain CI-owned.

The storage-upgrade decision record and migration documentation describe the final integrity contract and the limits of the recorded measurements.


yzxoi (commented):
Reviewed and fixed retirement verification: every source and backup digest is checked before deletion, including equal-size mutations outside former sample positions. Seventeen targeted tests, all 17 quality gates, workspace typechecks, and pre-push checks pass. This PR is authored by the current account, so approval cannot be submitted; merge will use the explicitly authorized self-review bypass after CI validation.

yzxoi (commented):
The Docker Lifecycle job passed on rerun, and all 35 non-fan-in checks on the reviewed head now pass. The original OOM evidence assertion also passed in an isolated local rerun. No test assertion was weakened or skipped to resolve the remote failure.

Full-file retirement integrity remains covered by the equal-size corruption regression and the focused storage suite. The combined branch includes the deferred-import safeguards and passes the integration checks recorded during review.

PR #1426, opened 2026-09-18
feat(desktop): keep the machine awake while Synergy is working

EricSanchezok:
## Why

A Synergy turn can run for many minutes, and Desktop is routinely left unattended while one is in flight. If the machine's idle-sleep timer fires mid-turn, the local runtime and its provider connections are suspended and the task loses its execution surface. Desktop shipped **no** power management at all before this change: `caffeinate`, `powerSaveBlocker`, `powerMonitor`, and `IOPMAssertion` had zero references across `apps/` and `packages/`.

Codex solves this with `features.prevent_idle_sleep` ("Prevent sleep while running"), implemented on macOS by spawning `/usr/bin/caffeinate -dims`. That carries two known defects: the child must be managed by hand (a leak means the machine never sleeps again), and its own issue tracker records the setting still failing once the display turns off.

## What

Determining whether anything is running is not a question the renderer can answer: session status is server-pushed derived data, a renderer only loads the scopes it has opened, and on macOS closing the window leaves the app and its managed server alive with **no renderer at all** — precisely the unattended case this exists for. Each part therefore owns exactly one thing.

**The server owns the predicate, once.** `GET /global/activity` returns `{ active, sessions, backgroundJobs }`: every non-idle session across all scopes (`listStatuses()` with no scope argument is already a cross-scope in-memory aggregate) plus in-flight loop background jobs via a new `LoopJob.activeBackgroundCount()`. Counting background jobs is required, not incidental — detached loop jobs deliberately outlive their turn lease, so session status alone reports a machine running only background work as idle. The route lives under `/global/*` to reuse the existing scope exemption, and joins `/global/health` in the request-logging exemption since it is polled every 5s.

**The main process owns the OS assertion.** `apps/desktop/src/power-save.ts` wraps `powerSaveBlocker` with `prevent-app-suspension` (system stays awake, display may still sleep) and treats the blocker's own `isStarted` as truth, so an assertion the system dropped across suspend/resume is re-acquired. The watcher applies a rising edge immediately, holds a falling edge behind a 5s debounce so the gap between turns does not flap, and keeps the last known state through transport failures until a 15s grace expires. All three quit paths release it.

**The renderer is only a latency hint.** It asks the shell to re-check; it never reports a state of its own — which is what makes a latched assertion structurally impossible.

Opt-in and off by default, stored per user in the shell's own state directory (not server config), exposed only on Desktop.

## Alternatives considered

Recorded in the decision record, including: renderer-reported state (latches on, misses unloaded scopes), renderer-only (fails the closed-window case), subscribing `/global/event/ws` (rejects clients without `Origin`; main-process WebSocket cannot set it), `caffeinate` shell-out, `prevent-display-sleep` (more intrusive and, per Electron #12345, not reliably stronger on Windows), and defaulting to on.

## Verification

- **Real Electron run, 15/15 assertions passed** against genuine OS assertions via `pmset` — `R1` holds `NoIdleSleepAssertion` while leaving the display free, is idempotent, releases cleanly, re-acquires after an external drop (`R7`), clears after the debounce, and leaves nothing behind on stop (`R6`). Note: macOS always reports a baseline `PreventUserIdleSystemSleep` owned by `powerd`, so assertions must be matched by owning PID — system-wide counters cannot attribute anything.
- **Live HTTP** on an isolated server: `200`, `Cache-Control: no-store`, correct idle shape; route present in live OpenAPI as `global.activity`.
- Server test exercises the **real `Server.App()`** (not a bare router), covering cross-scope visibility and detached background work.
- `quality:quick`: all 17 gates pass. Desktop 289 pass / 0 fail. Web suite green. Harness + server suites green. i18n extract/compile/check pass.

## Known limitations

Documented rather than worked around: closing a laptop lid still suspends on all three platforms; Windows modern-standby (S0) is not guaranteed by `SetThreadExecutionState`; Linux depends on a D-Bus inhibit provider a minimal environment may lack; macOS App Nap may coalesce a hidden app's timers (no `Info.plist` exemption is added); a session still queued for recovery does not count as active until it actually executes, since the poll must stay I/O-free.


yzxoi (dismissed):
Reviewed activity aggregation, Electron power assertion lifecycle, preload contract, and UI preference. Fixed stale probes across watcher restarts, assertion reacquisition after resume, and hung-request timeout handling; regression tests failed before each fix. Desktop, server, loop-job, and Web targeted suites plus all 17 quality gates pass. Native machine suspend was not exercised.

yzxoi (approved):
Reviewed the final desktop activity and keep-awake changes, including generation fencing, bounded activity requests and reacquisition after an operating-system resume. The three behavioral regressions and focused desktop/runtime checks pass.

The new runtime CI failure was reproduced by running an independent Session after the delegated DAG task: the test observed a global last-request variable. Prompt observations are now keyed by the owning Session; every target request retains the upstream result assertions, and the independent Session must not receive that context. The full affected CI shard passes locally (328 tests), all 17 quality gates pass, and the synchronized branch passes pre-push checks.

PR #1427, opened 2026-09-18
fix(enforcement): make full_access a complete permission bypass

EricSanchezok:
## Summary

Makes the three control profiles mean exactly what their names say. `full_access` no longer produces a permission denial for any reason originating in Synergy's own permission subsystem, enabling it from the UI is preceded once by a risk confirmation, and the profile used by non-interactive sessions becomes configurable.

Fixes the reported incident: a user who had selected `full_access` was refused an ordinary SSH command because a Policy Worker classification timeout (default 1s) overrode the profile's allow.

## What changed

**`full_access` is now a complete bypass.** Classification still runs so capability data reaches audit and rollout records, but a worker failure no longer changes the decision. The opaque non-bypassable `protected_op` capability is still synthesized, so the failure stays visible in gate audit records and the rollout authorization artifact, and `enforcement.policy.fallback` still records. `guarded` and `autonomous` keep the transient denial unchanged — only the profile whose purpose is to pre-authorize everything opens.

**Transient failures read honestly.** `PolicyDenied` now carries `permanent` and `guidance`, `retryable` derives from `permanent`, and a `modelMessage` getter stops rendering an infrastructure failure as a restriction the agent "must not retry" — which previously contradicted the gate's own retry guidance.

**One-time confirmation before Full Access.** A danger-toned dialog describes what the mode removes. All three human activation paths route through one composable: the prompt-input selector, the kanban pane selector, and the Settings panel. It records `fullAccessAcknowledged` and never re-prompts, including when re-selecting the mode already in force. It is an awareness gate, not a security boundary — it does not gate the HTTP API and never applies to programmatic session creation.

**Configurable non-interactive default.** New `permissions`-domain key `nonInteractiveControlProfile` (`autonomous` | `full_access`, default `autonomous`). `guarded` is excluded because an unattended session under it would leave a permission ask pending with nobody able to answer. `Session.defaultControlProfileForSessionSource` becomes the single owner, covering both Channel and Agenda roots, and the three hard-coded `autonomous` sites are collapsed into it.

**Vocabulary.** Each profile description now names its own behavior: `guarded` asks, `autonomous` never asks and denies instead, `full_access` never asks and allows everything.

## Notable design points

- The new key resolves **ahead of** the generic global `controlProfile` for non-interactive roots. Without that hoist, removing the Channel host's explicit profile would let an operator's interactive default silently turn unattended Channel work into `guarded` — the one profile that can only stall on an unanswerable ask. This regression has a dedicated test.
- Existing non-interactive sessions are deliberately **not** migrated: a session keeps the profile persisted on it, because re-permissioning a running remote task retroactively is worse than a setting that governs new sessions.
- `interaction.mode` stays. It answers a different question ("can this session ask a human?"), changes behavior in only two places, and feeds session grouping and the statistics rollup.

## Verification

- `packages/harness`: control-profile + enforcement 1428 pass; permission + session 1063 pass
- `apps/web`: 2118 pass, 0 fail
- `packages/workflows`: 425 pass; `packages/connections`: 471 pass; product-runtime unattended 9 pass
- `bun run quality:quick`: all 17 gates pass; `decision:check` and `doc:check` pass; i18n `compile --strict` passes
- **Per-profile regression pinned in one place**: `policy-fallback.test.ts` now asserts the three profiles side by side on one denied command along both paths — with classification available, the documented `ask` / `deny` / `allow` split; with the Policy Worker stopped, `guarded` and `autonomous` producing identical envelopes apart from the answering profile, and `full_access` keeping its `allow` with the opaque marker retained for audit and no refusal. A future edit to either profile's failure handling fails here first.
- New coverage: three-profile split in `policy-fallback.test.ts` (including a genuinely stopped Policy Worker), transient-vs-permanent message rendering, config-schema rejection of `guarded` for the new key, and the non-interactive hoist regression
- **Acknowledgement is driven through all four human activation paths**, not just the predicate: 19 behavioral tests mount the real components and assert that each call site prompts when the key is absent, writes the key on confirm, stays silent once recorded, and leaves the config untouched on dismiss. Deleting the composable call from the three edited call sites turns **12 of the 19 red** (prompt-input 3, kanban 3, settings 6) while the 7 already-acknowledged and non-Full-Access cases correctly stay green, so the tests prove the wiring rather than restating it:
- The full `packages/harness` suite reports 32 pre-existing failures in migration/lifecycle files unrelated to this change; verified by running the same suite with these changes stashed (same 32), and each file passes in isolation

## Docs

Implemented decision record under `docs/decisions/implemented/architecture/`, corrected `execution-boundaries.md` (which asserted the old fail-closed behavior) and `runtime-and-scope.md`, updated the product docs that stated a fixed `autonomous` profile, regenerated the reference pages, and updated the `change-execution-boundaries` Skill checklist.

## Known limitations

- `guarded` combined with unattended interaction can still leave a permission ask pending indefinitely (`PermissionNext.ask` has no timeout). The new key keeps that combination out of unattended automation but `session_control.set_mode` can still produce it; documented rather than silently changed.
- While the Policy Worker is down, `full_access` executes operations the classifier would otherwise have labelled, with the audit record and fallback metric as the evidence trail. Accepted and documented: `full_access` already allows every classified operation, so the classifier's labels are never what stops anything under that profile.
- The end-to-end check was performed at the gate level with a genuinely stopped Policy Worker, not on a separate isolated runtime instance.


yzxoi (dismissed):
Reviewed full_access precedence and the policy-worker failure path against the execution contract. Guarded and autonomous restrictions remain intact, while full_access bypasses permission-system denials; runtime failures remain visible. Permission, session-profile, and Web acknowledgement tests and pre-push gates pass.

yzxoi (dismissed):
Reviewed full_access behavior against guarded and autonomous profiles. The existing permission, policy-worker failure and UI acknowledgement regressions pass.

The coverage failure came from a one-second race around a correctness-only parent completion observation. Both adjacent ordering fixtures now await the actual notification under the test framework deadline and drain the owned Cortex task before restoring mocks. Terminal status and exact output assertions remain intact. All 55 manager tests and all 232 workbench tests under coverage instrumentation pass locally; synchronized-base pre-push checks pass.

PR #1428, opened 2026-09-18
fix(retry): retry unmapped certificate failures within a bounded budget

EricSanchezok:
## Problem

Bun's bundled BoringSSL throws `unknown certificate verification error` with **no reason code** when it cannot attribute a TLS verification failure. The shared classifier matched only known reason codes and fixed certificate wording, so the error fell through to `UnknownError`; session retry admission rejects unknown errors.

The consequence was disproportionate to the cause: a Cortex child task that had already completed **74 successful model calls over 8.2 minutes** died on its 75th call and delivered nothing, right after the model said it was doing its final checks before writing up. The failure was a transient local proxy handshake break — an immediate retry would have recovered it.

The root cause is a missing semantic, not just a missing string. The three-kind classifier (`transient`/`permanent`/`aborted`) can express "retry may help", "deterministic failure", and "cancelled", but **not "cause unknown but empirically transient"**. The prior policy assumed certificate failures are deterministic, which holds for a mapped code and fails for an unattributable one — so the evidence-weakest case got the most aggressive verdict.

Environment evidence confirms this was transient, not misconfiguration: the endpoint resolved into `198.18.0.0/15` (RFC 2544, the fake-IP range used by Clash/mihomo-style proxies) with a TUN default route, yet the presented certificate was genuine, publicly issued, and verified against the system trust store. Interception was at the DNS/routing layer, and the same string recurred 34 minutes later in a sibling session with the identical 110-byte `partial` response signature.

## Change

**Classification by evidence strength.** Adds an `indeterminate` kind carrying a `tls-verification` category. A mapped reason code (`CERT_HAS_EXPIRED`, `DEPTH_ZERO_SELF_SIGNED_CERT`, `ERR_TLS_CERT_ALTNAME_INVALID`, …) stays terminal; certificate verification wording *without* a code earns a narrower budget. `undefined` keeps meaning "unrecognized".

**Bounded retry.** `SessionRetry.retryable()` now returns `{ message, maxAttempts }` so the budget is explicit at the call site rather than a global constant. Unattributable failures get **2 extra attempts and a 5-second backoff ceiling**; every other case keeps 10 attempts and 30 seconds. Each caller still owns its own budget and they never multiply; SDK retries stay at 0.

**Honest retry reason.** The retry message reports the TLS verification problem instead of reusing "Provider is temporarily unavailable".

**Endpoint evidence.** Persisted `APIError.metadata` gains `networkKind`, `category`, and `endpointHost`. The host is derived on the control plane from the same connection identity `providerRetryKey` already uses, so the worker protocol is untouched (`VERSION` stays 9) and no credential-bearing URL path reaches persisted metadata — provider keys are commonly embedded in the URL path, so reducing a URL to its pathname is not redaction.

**Diagnostics.** `synergy doctor` reports provider **endpoint routing** (fake-IP ranges `198.18.0.0/15`, `240.0.0.0/4`; half-default route pairs; TUN-bound default route) and **endpoint TLS** (real handshake, issuer reported). Probes are injectable for testing, strictly time-bounded, and degrade to `warn` offline. `100.64.0.0/10` is deliberately excluded because Tailscale serves real endpoints there.

TLS verification is never relaxed: no `rejectUnauthorized: false`, no CA auto-trusting, no verification bypass.

## Verification

Local gates: `quality:quick` **17/17 pass** (`decision:check`, `doc:check`, `skill:check`, `test-layout:check`, typecheck, deps, secrets, deadcode, and the rest).

Focused suites, all green:

| Suite | Result |
| --- | --- |
| `util` full package | 206 pass |
| `harness` session/provider retry | 78 pass |
| `harness` provider retry (incl. coordinator) | 21 pass |
| `harness` processor retry (end-to-end) | 9 pass |
| `harness` agent call | 19 pass |
| `harness` endpoint diagnostics | 24 pass |
| `runtime-local` webfetch | 19 pass |
| `cli` doctor endpoints | 1 pass |

The incident string is covered end-to-end by a processor-level test asserting classification, admission, a budget of exactly 2, and persisted `networkKind`/`category`/`endpointHost`, alongside a companion test proving a classified transport failure still reaches 10 attempts. Regression coverage keeps every mapped certificate code terminal.

Live verification on the machine that exhibited the fault, with no injected fakes — real DNS, real route table, real TLS:

```
WARN | Provider endpoint routing
   api.deepseek.com resolved to 198.18.0.118, inside 198.18.0.0/15 …
PASS | Provider endpoint TLS
   api.deepseek.com completed a TLS handshake with a certificate issued by
   "TrustAsia DV TLS RSA CA 2025", resolved to 198.18.0.118
```

That output is the incident's environment reproduced and correctly diagnosed, with genuine end-to-end TLS distinguished from DNS/routing interception.

## Docs and records

Correction to `docs/architecture/llm-loop.md` (certificate failures split by evidence), a new implemented decision record, a factual correction plus cross-link on the superseding `2026-09-15-network-retry-boundaries.md` record, postmortem `0017`, `cli-guide` diagnostics guidance, and rule updates to the `integrate-llm` and `find-logs` skills. The prior record also claimed Bun ignores `NODE_EXTRA_CA_CERTS`/`SSL_CERT_FILE`; that is no longer true on 1.3.14 and the guidance now reflects verified behavior.

## Not covered

Cortex terminal semantics are unchanged — no task revival, bounded auto-continuation, or partial-output retention. MCP, plugins, and other write paths may surface the same string; only model calls and web fetch adopt this kind, which is recorded as an explicit boundary rather than left implicit.


yzxoi (dismissed):
Reviewed network error normalization, bounded certificate retries, endpoint diagnostics, and fetch retry deadlines. Permanent certificate failures still fail without retry; only unmapped verification failures receive the bounded budget. Validation: 135 targeted utility, harness, webfetch, and CLI tests plus all pre-push gates pass.

PR #1429, opened 2026-09-18
refactor(workflows): make Plan prompt guidance instead of tool enforcement

EricSanchezok:
## Problem

The Plan workflow implemented its "research and design before executing" contract as a tool allowlist inside `SessionModePolicy.visibility()`. A session in Plan resolved a tool only when its name appeared in `PLAN_EXPLICIT_ALLOW` (37 names) or when `ToolTaxonomy.classify()` returned a kind in `PLAN_READ_KINDS` without the `stateful`/`externalIO` traits. Against the shipped built-in catalog that admitted 40 of 112 tools.

The list failed at both of its jobs:

- **As a boundary it was inert.** `bash` sat in the allowlist with no call-time restriction, and the suite asserted that `rm`, `git commit`, and `git push origin main` produced no Plan diagnostic — so the contract's "do not commit, push, or deploy" was never enforced anywhere. The enforcement gate does not read workflow state.
- **As guidance it obstructed its own goal.** The prompt asks the agent to inspect code, docs, sessions, memory, and external sources "until you can enumerate the viable routes", while the allowlist blocked every MCP tool, the read-only browser inspection tools, `attach`/`render`, `process`, `email_read`, `computer_observe`, `todowrite`, and `memory_write`.

Two structural gaps kept it wrong for new tools. `ToolTaxonomy` has no `mcp__` registration and every fallback pattern is anchored at `^`, written for the retired built-in `websearch`/`arxiv_search` that bundled MCP servers replaced — so `mcp__anysearch__search` fell through to `platform.external` and was blocked, along with every third-party MCP server. And being an allowlist, each newly added tool was blocked by default.

Maintaining it there also mis-stated where enforced limits live: under `guarded`, `file_write` is `allow` without approval, so the allowlist was the only thing between a Plan session and `edit` — a policy living in a collaboration mode rather than the profile that owns it. Sixteen blocked entries were already gated correctly elsewhere (`workflowToolAvailability`, `latticeVisibility`, `bossVisibility`, `channelToolVisibility`); those are semantic role/ownership boundaries, not policy.

## Change

- Delete `PLAN_EXPLICIT_ALLOW`, `PLAN_READ_KINDS`, `planBlocked()`, `isPlan()`, and the workflows `evaluateCall()`; Plan now gates no tool.
- Keep every semantic gate: `latticeVisibility`, `bossVisibility`, `workflowToolAvailability`, `channelToolVisibility`, and the forced `note` group.
- `unavailable()` returns `undefined` for `permission`/`user_disabled`/`deferred` so the core implementation owns those messages; the registration site drops its `isPlan()` guard.
- Remove the now-unproduced `plan_mode_blocked` code and the `mode` field. No migration needed: tool-part metadata is unvalidated and `isAISDKToolError` keys off `metadata.source`.
- Rewrite the Plan prompt around the deliverable rather than a restricted toolkit. The old `You MUST NOT` block forbade file writes, which the removed allowlist had made stale, so it goes too. Every sentence carrying the Blueprint quality bar is preserved verbatim and now pinned by strength-regression assertions; `plan-synergy.txt`/`plan-synergy-max.txt` are untouched.
- Update `docs/architecture/workflows.md`, add a `change-execution-boundaries` rule, and record the decision.

## Verification

- `packages/workflows`: `test/session/` (107 pass), `test/lattice/tool-visibility.test.ts`, `test/lattice/controller.test.ts`
- `packages/product-runtime`: `test/tool/exposure.test.ts` (33 pass), incl. new coverage that Plan gates no tool and that MCP tools are eligible rather than diagnosis-blocked
- `packages/harness`: `test/session/domain-contributions.test.ts`; `packages/note`: `note-blueprint-policy`; `packages/connections`: `channel/tool-policy`
- `bun run quality:quick` — all 17 gates pass; typecheck clean for `workflows`, `harness`, `product-runtime`

One local failure, `bundled-workflow.test.ts`, is environmental and pre-existing: it `copyFile`s a Darwin-only SQLite artifact built by `build-helpers.yml`, absent in a fresh worktree, and fails identically on stashed baseline. It is guarded by `process.platform === "darwin"`, so it cannot affect the Ubuntu CI runners.

## Accepted consequences

- A Plan session can now edit repository files, since `guarded` allows `file_write` without approval. The same capability was already reachable through `bash`, which changes files without a reviewable diff, so this moves the action onto an auditable path rather than opening a new one. Plan must therefore not be described as a read-only boundary anywhere.
- Under `full_access`, outbound tools become reachable from Plan; under `guarded` they still go to approval. The compensating control is the prompt contract plus the profile choice.
- Prompt compliance now carries the part of the contract code used to approximate, so the load-bearing sentences are pinned by regression assertions.

## Alternatives considered

- **Keep a narrower deny-list.** Improves on the shipped state but keeps approximating a boundary the `bash` path bypasses.
- **Keep the allowlist and register MCP classification.** Fixes the symptom while keeping a list that must be extended for every future tool and third-party server.
- **Give Plan a read-only control profile.** The correct mechanism for a hard no-write boundary, recorded in the architecture doc as the path to take if one is later required; adding a profile is a separate change to the execution-boundary model.


yzxoi (approved):
Reviewed Plan tool exposure, delegation visibility, and guidance against the configured permission profile. Updated stale read-only claims in product and architecture documentation. Workflow and product-runtime targeted tests, all 17 quality gates, and pre-push checks pass.

PR #1430, opened 2026-09-18
feat(storage): defer legacy session import past activation

yzxoi:
Adds an opt-in deferred import boundary for already-current archived Session cohorts. Bootstrap still seals the complete backup. If any registered owner migration is pending, the central runner stages all unresolved Session records first, preserving the original files until migration succeeds. Failed owner fences survive later runs of unrelated migration domains. Non-archived Sessions import before startup recovery.

Touch and endpoint resolution import before index lookup. Pending listing projections stay in memory; index writers only use SQL entries. Import batches are bounded by rows and bytes, binary evidence is copied without whole-file buffering and pinned against collection, and all recognized source hashes and target artifacts are verified before retirement. Missing or malformed owners are quarantined; storage and I/O failures remain retryable. Unknown auxiliary files are preserved.

The Runtime owns and drains background convergence. Portable export, merge, move and target switches reject pending or quarantined aggregates. Compatibility metadata does not cross portable archives. Historical retired endpoints and nullable Channel metadata remain intact in canonical history.

The previous per-Session replay registry was removed because it omitted Scope and product-owned migrations. This conservative boundary does not promise faster startup for homes needing schema migrations, and no performance figures are claimed for the final implementation.

Validation: 235 harness storage/migration/navigation tests passed with 5,340 assertions. The complete storage directory subsequently passed against PostgreSQL 17 on the final head: 136 tests, 5,114 assertions and no skips. Nine product transfer/registry tests and 61 focused migration ownership tests passed. Historical v1.2.33, v2.4.4 and v3.0.22 writer fixtures passed with deferral both enabled and disabled, and an already-current archived cohort passed activation and touch import. All 17 quality gates and pre-push checks pass. New regressions reproduced the original migration, ownership-fence, endpoint, projection and export failures before correction.


yzxoi (commented):
Completed the storage review and pushed fixes for migration coverage, source retirement, bounded binary import, endpoint discovery, index projection isolation, startup recovery, shutdown, and transfer integrity. Validation: 235 harness tests and nine product integration tests; all 17 quality gates and pre-push checks pass. PostgreSQL-only cases remain CI-owned. As this is authored by the current account, the explicitly authorized self-review bypass will be used only after CI validation.

yzxoi (commented):
Reviewed the final migration ownership fence: a failed owner remains fenced when a later run processes only an unrelated domain. The regression fails before the fix and passes afterward; 61 focused storage and migration tests pass on the final head. The earlier broad storage, historical upgrade and transfer suites also pass, and pre-push checks pass.

The opt-in deferral remains limited to already-current archived cohorts. It makes no performance claim for homes that still require schema migrations.

yzxoi (commented):
Final-head verification also passes the complete storage directory against an isolated PostgreSQL 17 instance: 136 tests and 5,114 assertions, with no skipped cases. The earlier two PostgreSQL-only omissions are now covered locally. Remote CI is still queued; this supplements the migration, historical writer and transfer regressions already reported.

PR #1431, opened 2026-09-18
fix(session): release sessions pinned by driverless workflows and settle interrupted turns

EricSanchezok:
## What

A host suspend froze the backend for ~500s. On wake the storage driver killed a healthy SQLite worker over a wall-clock deadline and kept serving HTTP against a closed store. After the restart, one session stayed pinned in `recovering`, the abort control returned 200 without doing anything, and two in-flight `bash` calls rendered as permanently running. Seven abort presses did nothing; the only interaction that unblocked the session was an undocumented 2-second long-press.

The change fixes the six defects that produced that state.

## The invariant

> A persisted active workflow is a real driver **only** when something durable will resume it.

Startup reconciliation previously only *restored* loop/session/note bindings; it never asked whether anything was driving the loop. Since the client derives `recovering` from that same record, no user-facing control could clear it. Reconciliation now adjudicates a loop with no resumable evidence instead of rebuilding its bindings.

Preserved as genuine drivers (never terminalized): a user-paused (`waiting`) loop, a loop awaiting review (`stopRequest`), a Lattice-owned loop (Lattice reconciles its own loops after session recovery), and a loop whose session has queued inbox work or a continuation-recovery intent. An orphaned `armed` loop is `cancelled` (not `failed`) because `armed → failed` is not a legal transition; a started orphan is `failed` with an `interrupted:` error.

## Defects fixed

| #   | Defect                                                                                                                                                      | Fix                                                                                                                                   |
| --- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
| A   | A persisted active loop was treated as proof the session was still working, pinning it in `recovering` forever                                              | Adjudicate loops with no resumable evidence; clear their bindings through the existing terminal branch                                |
| B   | `POST /session/:id/abort` unconditionally returned `true`, so a no-op stop was indistinguishable from a real one                                            | Return `outcome`/`repaired`/`abandoned`/`settled`; release a driverless workflow so the visible control actually unblocks the session |
| C   | Interrupted-turn repair terminalized only the assistant message, never its tool parts — leaving them `running` forever and contradicting the rollout ledger | Settle non-terminal tool parts to `error` in every repair branch; migration repairs the same contradiction in historical records      |
| D   | `recovering` collapsed three unrelated causes into one opaque status, so the UI claimed an incomplete turn when an active workflow was the cause            | `recovering` carries a `reason` and optional `description`; the status bar reports the real cause                                     |
| E   | The only control that worked was a 2-second long-press, and the stop affordance vanished once the composer had text                                         | Single-click stop control, available while a draft is present                                                                         |
| F   | A wall-clock request deadline fired across the suspend and killed a healthy worker; the process then served HTTP over a dead store                          | Monotonic deadline budget with re-arm; corroborate a kill with a liveness probe; fail fast into the managed restart path              |

## Scope boundary

Esc/Ctrl+G and the plugin `stop()` remain **turn-only**. Cancelling a workflow stays a separate, explicit action: the backend abort route releases only *driverless* workflows, so a healthy loop is never terminated by an ordinary stop.

## Verification

- New behavioral tests: `loop-liveness-adjudication` (9), `abort-escape-hatch` (6), `crashed-runtime-recovery` (4, end-to-end incident reproduction), `orphaned-tool-parts` (4), `orphaned-tool-parts-migration` (6), `storage-worker-suspend` (4), `storage-unavailable-escalation` (2), plus status-bar and stop-control tests.
- Negative control on the storage fix: restoring the pre-fix driver fails 3 of 4 suspend tests.
- Negative controls on adjudication: a `waiting` loop, a review-pending loop, an inbox-driven loop, a Lattice loop, and an already-terminal loop are each asserted to survive.
- Package suites pass through the repo's sharded CI runner (`bun run test`): harness, product-runtime, workflows, server, apps/web — all exit 0.
- `bun run quality:quick`: all 17 local gates pass.
- SDK/OpenAPI regenerated via `./script/generate.ts`; regeneration is content-stable.
- i18n extract/compile/check pass. `decision:check` and `doc:check` pass.

## Known limitations

- No real OS suspend was exercised in a test. The storage suite injects the monotonic clock, which proves the deadline arithmetic and the escalation state machine but not that `performance.now()` excludes suspend on every platform. If that premise fails on some platform, the false kill would persist there.
- Detecting a genuinely dead worker is now slower (request deadline plus up to 3×30s probes). That is the deliberate trade for never misclassifying a slow worker as dead.
- The escalation path (`Storage.onUnavailable` → `gracefulShutdown`) is verified through the store contract, not by observing a real server process exit under a supervisor.
- The stop control was verified through unit tests, typecheck, and i18n contracts; its rendered appearance was not visually confirmed (no browser in the sandbox).

## Docs

Postmortem `docs/postmortem/0018-host-suspend-pinned-session-in-recovering.md`, four decision records under `docs/decisions/implemented/`, and updates to `docs/architecture/{workflows,session-and-messages,llm-loop,agent-storage}.md`.


yzxoi (dismissed):
Reviewed session liveness adjudication, abort repair, orphaned tool-part settlement, SQLite worker deadlines, and terminal storage escalation. Fixed swallowed driver-read failures so unavailable evidence cannot falsely fail a live workflow. The two new regression tests fail before the fix. Validation: 66 product-runtime and 18 harness recovery/storage tests, all 17 quality gates, and all pre-push checks pass.

yzxoi (dismissed):
Revalidated the reviewed recovery implementation after documentation-only integration fixes. The host-suspend postmortem now has a unique number and all incident references follow the rename. The earlier 84 recovery/storage tests and all quality/pre-push gates passed; no runtime behavior changed in these follow-up commits.

PR #1432, opened 2026-09-18
refactor(enforcement): one owner for shell risk classification

EricSanchezok:
## What this changes

Makes the shell risk classifier the single owner of every shell decision, so no capability is decided by scanning command text.

- **Destructive classification has one owner.** `ShellSafety.classifyBashRisk` decides host-level destruction from tokenized operands, and the two substring layers (`analyzeDestructiveCommand` with `DESTRUCTIVE_PATTERNS`, and `isDestructive`) are deleted. The eleven host-level shapes those layers uniquely covered — `rm -rf /`, `rm -rf ~`, `mkfs.ext4 /dev/sda`, `chmod -R 777 /`, `shred /etc/passwd`, `sudo rm -rf x`, and their wildcard/`$HOME` spellings — are covered first, in a separate commit, asserted before the deletion lands.
- **Remote irreversibility follows the destination branch.** A force or delete against a protected branch stays a remote write; the same operation against a named non-protected branch publishes, because the loss is bounded by that branch's own commits. `--force-with-lease` and `--force-if-includes` are no longer destructive in themselves.
- **Reflog- or worktree-recoverable operations are no longer refused**: `rebase` (including `--abort`/`--continue`), `reset --soft`/`--mixed`, `stash pop`, `commit --amend`, `revert`, `pull --rebase`, `git rm`. Irreversible history (`reset --hard`, `clean -fdx`, `stash drop|clear`, `checkout -- <path>`, `filter-branch`) keeps its boundary.
- **Network activity is decided from the resolved operation, not from text.** A closed set of network clients, the git/OpenSSL subcommands that contact a remote, package-manager subcommands that reach a registry, and a remote `rsync` operand. Argument text that names a tool or contains a URL — `echo "see https://example.com"`, `rg -n "ssh " docs/`, `git commit -m "fix npm install docs"` — is no longer a network request. This matters beyond reporting accuracy: an allowed `network_request` also relaxed the sandbox network mode to full.
- **macOS sandbox path canonicalization fixed.** `canonicalize()` fell back to the caller's raw spelling when `realpathSync` threw, which is the normal case for a protected subpath that does not exist yet. On a firmlinked host the writable-root allow was bound to `/private/var/...` while the deny was emitted as `/var/...`, so the deny never intersected and a git hook could be planted in an aliased workspace. Now resolves the nearest existing ancestor and re-appends the remainder.
- **Unreachable code removed.** The `ExecPolicy` prefix-rule engine had no production caller (`GateOptions.execPolicy` was never populated) and is deleted; the refusal-amendment factory is reachable and stays.
- **Containment baseline added.** Real-command probes through the canonical wrapper for external writes, credential reads, ordinary external reads, workspace writes, and protected subpaths, so the layers that may later be removed have a recorded substitute.

## Not included

The sandbox does not yet own shell filesystem reach, so bash still predicts `file_external_*` capabilities. The ownership change is gated on per-platform containment evidence: the Linux restricted-mode read model is an enumerative allow-list (unlike macOS's global read allow), so removing the read prediction without converging the platforms first would regress ordinary external reads on Linux. That convergence is a security-boundary change to the Linux sandbox that cannot be verified on this host.

Consequence worth stating plainly: the false positive this work started from — a `grep -v "/test/"` argument being classified as a write to `/test/` — is **not** fixed here. It lives in the path-prediction machinery that the gated phase removes.

## Verification

- `packages/harness` enforcement suite: 1343 pass, 0 fail. Permission + session: 1063 pass. Sandbox: 295 pass. `packages/util`: 190 pass. `packages/product-runtime` workspace boundary: 20 pass.
- Full `packages/harness` suite: 3935 pass; the failures observed under full parallel load are in migration/terminal-display files this change does not touch and pass in isolation (42/42).
- `bun run quality:quick`: all 17 gates pass. `doc:check`, `decision:check`, `skill-check`, `test-layout:check` pass.
- Real-sandbox baseline on this host: external writes contained, credential reads denied, protected subpaths blocked through both firmlink spellings.
- Regression corpora pin both sides of every boundary: `test/enforcement/destructive-boundary.test.ts`, the network corpus in the same file, and `containment-baseline.test.ts`.

## Decisions

`docs/decisions/implemented/simplification/2026-09-19-single-owner-for-shell-destructive-classification.md`, `.../2026-09-19-remove-unreachable-exec-policy-engine.md`, `docs/decisions/implemented/bug-fix/2026-09-19-macos-sandbox-alias-canonicalization.md`.


yzxoi (dismissed):
Reviewed the shell-risk ownership consolidation and native sandbox exception scope. Fixed network detection for git and go global options, env split-string wrappers, and command lookup modes. Updated the destructive-path and publish contracts in the decision record. Validation: 1,079 harness classifier/gate tests and 45 native sandbox tests, all 17 quality gates, and all pre-push checks pass.

PR #1433, opened 2026-09-19
fix: recover missed title runs and align small-model call retries

### LINKED ISSUE #1417: A stalled or missed title run leaves a session permanently untitled
## Summary

Session title generation has two independent ways to fail permanently.

1. The `ensure-title` background job has a 120 s budget (`timeoutMs: 120_000` in `packages/harness/src/session/title.ts`). When the provider stalls past that budget, the job's abort signal fires, `AgentCall` refuses to retry an aborted call (`packages/harness/src/agent/call.ts`), and nothing re-queues the job — so a transient stall silently costs the session its title forever, even though the configured provider timeouts are far more generous.
2. `ensureTitle` returns early unless the history contains **exactly one** prompt-visible user message (`packages/harness/src/session/title.ts`). A session that already has two or more user messages can therefore never receive an auto title, which makes failure mode 1 unrecoverable and also means every forked session starts permanently untitled.

## User-visible behavior

- The session list shows `New session - <timestamp>` indefinitely; nothing in the UI reports a problem.
- Sending further messages does not fix it, and in fact makes failure mode 2 apply.
- The only trace is in the log: `failed to generate title` plus `ensure-title outcome=timeout`.

## Steps to reproduce

Defect 2 is deterministic:

1. Create a session and send two messages (or fork any session whose history already contains two or more user messages).
2. Send a new message.
3. The session keeps its default title and no title request is made at all.

Defect 1 is intermittent and needs the first title request to outlive the 120 s budget:

1. Start a session and send a first message while the model endpoint stalls for longer than 120 s (in the reported incident, all provider traffic went through a local HTTP proxy that stalled for minutes).
2. The title job fails with `Agent title was cancelled` / `job timed out after 120000ms`.
3. Sending a second message produces no new title attempt, because of the exactly-one guard.

## Expected behavior

A session that is still untitled should get a title generated on a later turn, and a transient provider stall should leave the session retryable rather than permanently disqualifying it. The title budget should be consistent with the configured provider timeouts, so a stall that the provider layer is willing to wait out does not silently destroy the title job.

## Actual behavior

The session stays on its default title with no further attempts for the rest of its life:

```
ERROR service=session.title error=Agent title was cancelled failed to generate title
ERROR service=session.loop-job type=ensure-title outcome=timeout error=Background job timed out after 120000ms. job failed
```

### Observed

- Over roughly one hour on one `bun dev desktop --managed` instance, three title jobs failed with exactly the two log lines above. Each failure was exactly 120 s after the corresponding title request assembly record, confirming the job budget (not the request) is what expires.
- Pairing each session's LLM-stream start and end records in the same log shows the stalls were simultaneous across unrelated sessions and providers: nine concurrent streams that started within a 65 s span all ended within a 46 s span about two minutes later, while streams that started just before the window completed normally. Two more such windows occurred in the same hour. This is what pushed the first title attempts past the 120 s budget.
- In that instance, provider traffic inherited `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` from the process environment, and `timeout.provider.idle_sec` / `ttfb_sec` were set to 900 / 3600 — three orders of magnitude more patient than the title budget.
- Code: `packages/harness/src/session/title.ts` registers `ensure-title` with `timeoutMs: 120_000`, and `ensureTitle` returns early when `promptVisibleUsers.length !== 1`.
- Code: `packages/harness/src/agent/call.ts` skips the retry when `abort.aborted` is true, so once the job budget expires there is no further attempt.
- Reproduced in isolation with an isolated home and a dead proxy: a session with one user message whose first title request could not complete kept its default title, and after a second message the log contained no further title-agent call at all.
- Reproduced defect 2 in isolation: forking a session that had two user messages and then sending a message produced a title-agent call for the per-message summary title (`packages/harness/src/session/summary.ts`) while the session title stayed at its default value.

### Inferred

- The 120 s job budget converts a provider stall that the provider layer would otherwise tolerate into a permanent, silent title loss, because the aborted call is never retried and the job is never re-queued.
- The exactly-one guard exists to avoid titling a long conversation from the wrong point, but as written it also blocks recovery for any session whose first attempt failed, and it blocks forked sessions by construction.

### Unknown

- Whether the 120 s title budget was chosen deliberately against the provider timeout configuration.
- Whether titling from the first user message when the history is longer (i.e. relaxing the guard to at least one) is acceptable, or whether a per-session notion of "the message to title from" is needed.
- Whether the title job should be re-queued on failure or retried within the same budget.

## Environment

- OS: Ubuntu 24.04.5 LTS (Linux x64)
- Synergy version: `dev` branch, commit `c959fa905`, launched with `bun dev desktop --managed`
- Bun version: 1.4.2

## Reproducibility

Intermittent for defect 1 (three of roughly ten title attempts in one hour, each requiring a provider stall longer than 120 s); deterministic for defect 2 (any session that reaches two user messages while untitled, including every fork of a multi-message session).

## Additional context

Proposed direction: retry or re-queue `ensure-title` when its budget expires instead of treating the session as permanently untitled, and derive the title from the first user message when the history contains more than one (or fall back to the fork source title). Related but separate: #1412 covers the settings UI being unable to clear a model role variant.

yzxoi:
Closes #1417

Retry automatic title generation on later turns while the title is still the default, including conversations with multiple user messages and forks. The title prompt remains bounded to the first real user message, and an in-flight call cannot overwrite a user rename.

Align bounded retries for title generation, SmartAllow, agent generation and the plugin agent.call service. Each call keeps its existing total deadline; an exhausted title deadline recovers on a later turn.

Validation: 51 harness tests and four plugin host-service tests pass. The title tests cover later-turn recovery, first-message context, total timeout and rename protection. All 17 quality gates and workspace pre-push checks pass. The decision record now describes the actual deadline and recovery behavior.


yzxoi (commented):
Reviewed title retry collection, first-user prompt boundaries, rename protection, and shared small-model call deadlines. Corrected the decision record: exhausting the total deadline still aborts the attempt; recovery occurs on a later turn, while retryable failures can retry within the remaining budget. Restored the timeout assertion and verified later messages do not enter title context. Validation: 51 harness tests and four plugin-host integration tests plus all pre-push gates pass. Self-review bypass is authorized after CI validation.

PR #1443, opened 2026-09-19
fix(server): make /global/activity in-memory and align its record

EricSanchezok:
## Problem

`GET /global/activity` is polled every five seconds by every Desktop with the keep-awake setting enabled. It computed `sessions` via `SessionManager.listStatuses()` **with no scope argument**, whose no-scope path walks the registered runtimes and then late-imports `SessionNav`/`SessionRecovery`, resolves every scope ID, and calls `SessionRecovery.recoverableStatuses()` per scope — reading session records from disk. Real cross-scope storage IO on a 5s poll.

The owning decision record claimed the opposite: that the call was "already a cross-scope, in-memory, non-awaiting aggregate".

## Change

- `packages/harness/src/session/manager.ts` — new `activeRuntimeCount()`: counts non-idle runtimes by iterating the module-private `runtimes` map. No storage access, no scope scan, no await.
- `packages/server/src/server/global-activity.ts` — `sessions` now comes from `activeRuntimeCount()`; OpenAPI description aligned with what the predicate actually counts.
- `apps/web/src/components/settings/SettingsPanel.tsx` — the desktop-power `createEffect` returned the unsubscribe function, which Solid ignores (that is React `useEffect` semantics), leaking one IPC listener per settings-panel mount. Now passed to `onCleanup`.
- Owning decision record — corrected the two false statements about how activity is computed.
- `packages/sdk/js/src/gen/{sdk.gen.ts,types.gen.ts}` and `packages/sdk/openapi.json` — regenerated with `./script/generate.ts`; not hand-edited.

Unchanged by design: the response shape `{active, sessions, backgroundJobs}` and the `Cache-Control: no-store` header. `LoopJob.activeBackgroundCount()` was already in-memory and is untouched.

### Behavior note

A session still queued for recovery after a server restart no longer counts as active until it registers a non-idle runtime. This was already the documented intent for the poll, and it self-corrects as soon as the session runs. Detached background work still counts, via the unchanged `backgroundJobs`.

## Verification

| Command | Result |
| --- | --- |
| `cd packages/harness && bun test test/session` | 912 pass, 3 skip, **2 fail (pre-existing)** |
| `bun test --cwd packages/server test/server/global-activity-route.test.ts` | 4 pass, 0 fail |
| `bun run typecheck` | 28/28 tasks successful |
| `bun run quality:quick` | All 17 gates passed |
| `bun run decision:check` | Decision validation passed |
| `bun run doc:check` | Document validation passed |
| `bun run --cwd apps/web test` | 54 pass, 0 fail |

The two `packages/harness` session failures (`compat-import.test.ts` "a deferred aggregate imports on touch, and leaves no JSON" and `summary.test.ts` "post job only collects for terminal assistant replies") are cross-file state pollution that reproduces on a pristine `origin/dev` checkout without this change, and both pass when their file is run alone.

### New tests

- `packages/harness/test/session/active-runtime-count.test.ts` — asserts the accessor counts busy/retry/recovering runtimes, ignores idle and unregistered ones, and performs zero `Storage` reads/scans/queries with a positive control proving the spies can observe IO.
- `packages/server/test/server/global-activity-route.test.ts` — new behavioral case: with a session queued for recovery that only the disk scan can see, the route serves the in-memory answer and never calls `SessionNav.getAllScopeIDs` / `SessionRecovery.recoverableStatuses`, while the same spies do observe the replaced `listStatuses()` path.

Both use the positive control so a "not called" assertion cannot pass because the spy was inert.

Review verification against current dev: in-memory runtime count tests (2), global activity route tests (4), the complete Web suite, and all 17 quality gates passed. Current dev is integrated; the fresh CI run must revalidate the previously failing installed-artifact timeout outcome.


yzxoi (approved):
Reviewed the in-memory activity count and event listener disposal. The route no longer triggers cross-scope recovery, and its response contract remains stable. Focused harness/server tests, the Web suite, and local quality gates passed. Fresh CI must pass before merge.

PR #1448, opened 2026-09-19
refactor(enforcement): sandbox owns bash filesystem reach

EricSanchezok:
## What and why

The enforcement gate predicted the filesystem reach of a bash command string by parsing its text, while the OS sandbox decided the same question from the real syscall. Two layers answering one question, and each answer a separate opportunity to be wrong: the gate passed `npm install -g pkg` that the sandbox then refused, and hard-refused `grep -v "/test/"` that the sandbox never cared about.

Each capability class now has exactly one execution-time owner:

- **Precise input belongs to the policy layer.** A structured tool's literal path argument is still resolved by `classifyPathCapability` / `classifyProtectedPathCapability`. Those functions survive; only their bash-branch callers were removed.
- **Imprecise input belongs to the execution layer.** bash now contributes only what the sandbox cannot express. Its branch mints no `file_*` capability at all.

## Changes

**bash loses its path-prediction machine.** `extractAbsolutePaths`, `extractShellPathArguments`, `writeRedirectTargets`, `maskHeredocBodies`, `isNullDeviceSink`, `SAFE_PSEUDO_PATHS`, `NON_PATH_PATTERNS`, the copy-operand role tables, `shellTokenize`, `aggregateWriteCapable`, the per-segment loop, and the `args.workdir` prediction are gone, along with `shell-safety.ts`'s directory-change analyzer and its thirteen helpers. `gate.ts` 1672 → 1159 lines; `shell-safety.ts` 4050 → 3246.

**The read-only catalog is gone, and with it `shell_read`.** The capability is removed from `SYNERGY_CAPABILITY_DETAILS`, `SYNERGY_PROFILE_CAPABILITIES`, the classifier, and the tool resolver. With no precise input to reason about, a read-only tier could only restate a guess about argument text. `find -delete`, `find -exec`, and in-workspace subtree removal become sandbox-owned.

**Authorization follows containment.** The sandbox's containment verdict is decided before the authorization outcome, so a contained call keeps its sandbox instead of switching it off, and a call that is not contained goes through the ordinary capability flow.

**`shell_destructive` copy narrowed** to what the sandbox cannot express.

**An actionable sandbox denial on the built-in bash path.** A denied write now names the exact path and access so `guarded` can approve that path and the retry re-enters the sandbox. The denial surface is gated on the child having failed, because a partially denied command that finished still reports its own exit status and must not be reported as blocked.

**macOS denial audit timing.** The audit stream is live before the child spawns: a fast command's record is emitted microseconds after spawn, so binding the pid afterwards lost the record that names the denied path. Cleanup no longer stops the session, because kernel records trail the child.

## Preserved

Privilege escalation (`hasSudoInvocation` and the sudo/runuser/pkexec/nsenter/container option tables), the fd/heredoc/stdin replay chain, irreversible local-history loss, remote mutation, `full_access` behaviour, the `autonomous` fail-closed fallback, and the structured tools' `file_external_*` gate semantics are all untouched.

Two costs are accepted and recorded in the decision record: a compound command can now produce partial side effects before the execution-time boundary refuses it, and `bash -c '…'` payloads surface as `network_request` for fewer spellings.

## One test probe updated, not its contract

`packages/cli/test/cli/artifact.test.ts` asserts a permission outcome by refusing a bash call in non-interactive mode. Its probe was a plain `touch` outside the workspace — exactly the filesystem reach this change moved to the sandbox. The gate now allows it because containment satisfies `shell`, the kernel refuses the write at execution time, and the fixture model (which always replays its tool call) looped until the run timed out, so the mode recorded an execution-time boundary as a timeout instead of the permission refusal it means to test.

Privilege escalation is not expressible to the sandbox, so it stays a gate decision. Probing with that keeps the assertion about authorization while leaving the containment handover intact.

## Verification

- `cd packages/harness && bun test test/enforcement` — **1317 pass / 0 fail** (includes `ownership-inversion.test.ts`, which pins the eleven sandbox-inexpressible shapes as still refused, the historical false positives as released including the original `grep -v "/test/"`, and `full_access` parity)
- `cd packages/harness && bun test test/sandbox test/session` — green
- `cd packages/runtime-local && bun test test/sandbox test/session` — **297 pass / 0 fail**, including `tool-resolver-autonomous-sandbox.test.ts` **5/5** and the real-Seatbelt end-to-end denial/approve/retry test
- `bun run typecheck` — 28/28
- `bun run quality:quick` — all 17 gates
- `bun run test-layout:check` / `decision:check` / `doc:check` — passed
- Zero-remnant greps return no production-code hits for the removed predictor, the read-only catalog, and `shell_read`.

## Dependency and merge order

This branch is rebased onto the Linux read-model convergence (`readDenyPathsFor`), which lands first. Without it, removing the gate's prediction would leave Linux enumerating read roots, so ordinary external reads like `cat /etc/hosts` would become sandbox failures while the architecture document already promises they are allowed. That ordering is a safety constraint, not a preference.

Review hardening: scope macOS denial log processes to every foreground return and exception, assemble complete records before interpreting denied paths, retain bounded records and filter other process IDs. Integrates the canonical Linux mount-order fix from prerequisite #1449. Validation: all five new regressions failed before repair; 1,407 Harness enforcement/sandbox tests, 303 Runtime Local sandbox/session tests, and all 17 quality gates passed. Linux execution evidence remains owned by CI.


yzxoi (dismissed):
Reviewed containment-based shell authorization and the sandbox denial lifecycle. Fixed audit stream disposal on success and failures, complete-line framing, and bounded candidate retention. Enforcement and runtime sandbox regressions, including real Seatbelt approval/retry, and local quality gates passed. Merge after the Linux deny-model dependency and fresh CI.

PR #1449, opened 2026-09-19
feat(sandbox): converge Linux reads onto the deny-list model

EricSanchezok:
## Summary

Linux adopts the same deny-list read model macOS already uses, so ordinary external reads no longer depend on the enforcement gate predicting the host paths a shell command will touch. This is the stated precondition for removing that gate's shell path predictor (WS3): on macOS the global read allow already covers any path, but on Linux deleting the predictor would have turned `cat /etc/hosts` into a sandbox failure.

Linux now declares `/` as its single readable root — selecting the helper's existing full-read bind — and carries the credential deny list instead.

## What changed

**One deny-set owner.** `readDenyPathsFor` in `packages/harness/src/sandbox/policy.ts` is now the single owner of the deny set, consumed by `buildPermissionProfile`, the macOS compiler path, and the Linux backend, so the platforms cannot drift. It deliberately does **not** accept the writable roots: the defect that `#1440` fixed was exactly a writable root pruning the denies it contained, and omitting the parameter makes that unreintroducible from a caller. `buildPermissionProfile`'s inline derivation is replaced by a call to it; macOS deny-set semantics are unchanged and `macos-policy.ts` is untouched by this PR.

**bwrap deny cover mounts, with a containment ordering rule.** bwrap has no deny rule, so a denied path is covered by an empty tmpfs (directory) or the null device (regular file); a tmpfs destination must be a directory, which is why a file takes `/dev/null`. Cover ordering is the Linux half of the contract `#1440` deferred, and follows the same containment rule as the macOS compiler because bwrap resolves shadowing by mount order, not specificity:

- a deny **containing** a writable root mounts **before** the writable-root binds, so a workspace nested inside a credential directory still works while its credential siblings stay hidden;
- a deny **equal to or inside** a writable root mounts **after** that bind, so the deny wins (equal path fails closed this way).

A covered path is never also read-only bound, and the skip is an **exact** match rather than ancestor containment — the ancestor form is what made `<ws>/.git/hooks` and `<ws>/.git/config` writable (git-hook planting) for a workspace nested under a deny root. Cover targets are canonicalized first, since bwrap refuses to mount over a symlink destination and credential stores such as `~/.ssh` are commonly links.

**Linux deny spellings.** `READ_DENY_PATHS` gains the Linux credential/session stores and Synergy runtime stores that only the removed `--tmpfs /` had been hiding (shell histories, browser/mail profiles, keyrings, `~/.password-store`, rclone/Terraform/MySQL/Postgres credentials, browser profiles, Inspire token, `library.db`, log, state). `~/.synergy/cache` itself stays readable because stage 2 re-reads the staged profile and re-execs the staged helper from it.

**Landlock eligibility.** `can_use_landlock_fallback` requires an empty read-deny set in addition to the full-read root, because a Landlock ruleset carries only allow rules and cannot express a subpath deny. Every Linux profile now carries a deny list, so the full-read check alone would have answered "eligible" for all of them and downgraded restricted reads to full reads. The disabled plan also returns empty rules.

**Unchanged:** write containment, `.git/hooks`/`.git/config` protection, the controlled temp root, the `autonomous` fail-closed fallback, the protected-create monitor, and the two-stage helper re-exec.

## Ordering assertion

The mount-order contract is a security property that fails silently when wrong, so it is asserted rather than commented. `packages/runtime-local/src/sandbox/helper-linux/src/bwrap.rs` unit tests pin: a denied directory covered by tmpfs; a denied file covered by the null device; a missing deny producing no mount; a deny containing a writable root mounting before it; a deny inside a writable root mounting after it; an equal-path deny winning; and a workspace nested inside a deny root keeping its `.git/hooks`/`.git/config` read-only binds after its writable bind.

## Verification

- `cd packages/harness && bun test test/sandbox` — **81 pass, 0 fail** (5 files)
- `cd packages/runtime-local && bun test test/sandbox` — **239 pass, 0 fail** (19 files), darwin
- `cargo test --release` in `packages/runtime-local/src/sandbox/helper-linux` — **121 pass, 0 fail**
- `bun run typecheck` — 28/28 tasks successful
- `bun run quality:quick` — **all 17 gates passed**
- `bun run test-layout:check`, `bun run decision:check`, `bun run doc:check` — passed

## What is NOT verified here

- **The Linux helper was not rebuilt and no hash was refreshed.** This host has only the `aarch64-apple-darwin` Rust target installed, so no Linux binary could be produced. The helper's SHA-256 is embedded at compile time by the release/runtime artifact build (`script/release/shared/build-runtime.ts` assigns `SYNERGY_SANDBOX_HELPER_SHA256`), and `TRUSTED_LINUX_HELPER_HASHES` is untouched by this PR. The helper rebuild and hash refresh are **UNVERIFIED locally and depend on CI's Linux matrix.**
- **No Linux real-sandbox execution was performed.** This is a darwin host: there is no bwrap and no user namespace support. The mount order is asserted at plan level in Rust, but end-to-end mount behavior is settled only by CI's Linux sandbox job.
- **Authoritative Linux evidence:** the `Test Shards (runtime)` job on `ubuntu-latest`, which installs `bubblewrap`, loads the AppArmor userns profile for `/usr/bin/bwrap`, builds and installs the helper to `~/.synergy/sandbox-helper/`, and sets `SYNERGY_TEST_LINUX_SANDBOX_E2E=1` (under which the containment-baseline skip conditions become hard failures). The Windows-focused job's `cargo check`/`cargo test` also compiles the helper crate.

`containment-baseline.test.ts`'s "ordinary external reads still work" now asserts ordinary external reads on **both** platforms (OS-home file, `/etc/hosts`, `~/.gitconfig`), replacing the previous darwin-only branch.

## Test notes

- New tests live under each owning package's `test/` directory; none beside `src/`.
- `containment-baseline.test.ts` keeps its existing availability gates (non-darwin/linux returns unavailable; Linux probes bwrap user namespaces).
- Deny probes use an injected synthetic deny root under the OS home plus a non-credential sibling as a control, so a denial is never asserted vacuously and no real credential content is read into test output.

## Docs

- New record `docs/decisions/implemented/simplification/2026-09-19-linux-sandbox-full-read-model.md`.
- `docs/architecture/execution-boundaries.md`: the macOS/Linux shared read model, the Linux mount plan and cover ordering, and the removal of the gate-forwarded read-root requirement.
- `docs/decisions/implemented/bug-fix/2026-09-18-linux-sandbox-readable-roots.md`: partial-supersession note (the enumerated read-root model no longer ships; staging, `/lib`/`/lib64`, and two-stage re-exec remain current).

## Scope note

This PR does **not** re-do or revert `#1440`'s macOS work: `git diff origin/dev...HEAD -- packages/runtime-local/src/sandbox/macos-policy.ts` is empty, and `#1440`'s existing workspace-shape tests still pass unmodified.

Review verification: added canonical mount comparisons for symlink aliases and root workspaces. Both regressions failed before repair. All 123 Rust tests, 81 Harness sandbox tests, 239 Runtime Local sandbox tests, and 17 quality gates passed on macOS; Linux execution remains covered by the required CI runtime job.


yzxoi (approved):
Reviewed the Linux read-deny model and fixed mount ordering for canonical symlink targets and writable root containment. Regression tests cover those cases. Rust helper tests, harness/runtime-local sandbox suites, and local quality gates passed. Fresh CI must pass before merge.

PR #1450, opened 2026-09-19
docs(process): require BlueprintLoop approval for acceptance

EricSanchezok:
## What and why

Four BlueprintLoops from one batch (2026-09-19) were merged as if complete while their own records showed otherwise. One never reached `completed` at all. Two had been explicitly rejected by the bound supervisor and were merged anyway. One described its work as done while its own `remaining[]` still listed the defect the work had started from as NOT FIXED. A fifth reported `Run Count: 0` and had never run.

Every one of those readings came from the executor's narration — generated inside the same session, from the same context, that produced the work. The loop's persisted state already carries an independent signal, so this rule points at that instead of at prose.

## Changes

- **`AGENTS.md`** — in "Protect Checkouts and Runtimes", next to the existing publication rules: a PR whose intent corresponds to a BlueprintLoop is accepted only when that loop reached `completed` through `blueprint_loop_approve`; an executor's self-assessment, or a loop stalled in `auditing`, failed, or rejected without a fresh audit, is not acceptance; the PR body states the loop's final status and audit conclusion.
- **`.synergy/skill/git-guide/SKILL.md`** — the matching step in the "Push and Open a PR" checklist, plus the companion rule that a red gate due to runner conditions (the local artifact build needs `docker`) must be re-run until green rather than merged past, and a check that cannot be made green is a blocker to report.
- **Decision record** — `docs/decisions/implemented/process/2026-09-19-blueprint-loop-acceptance-gate.md`, with the alternatives considered and why this is prose rather than a CI check.

The rule does not restate the reviewer contract; it points at it. `docs/architecture/workflows.md` already establishes that the persisted BlueprintLoop transition is the authoritative evidence the reviewer settled the stop request, and that a prose response or a terminal Cortex task is not an approval.

## Design note

The acceptance rule is a merge-authorization boundary, so it lives in `AGENTS.md` beside the other authorization rules. The green-gate guidance is procedural, so it lives in the Skill — which also keeps `AGENTS.md` inside its word budget.

Loop state lives in the note store outside the git tree, so a workflow job cannot read it without new pipeline infrastructure. Adding that would turn a process contract into a maintained dependency; the repository's other process conventions outside `decision:check` are prose for the same reason. This is deliberately the smallest reversible route and can be mechanised later if the rule proves hard to follow.

## Verification

- `bun run doc:check` — passed (`AGENTS.md` 1689 / 1700 words)
- `bun run skill-check.ts` — validated 22 repository Skills
- `docs/reference/tools.md` was **not** touched (generated file)


yzxoi (approved):
Reviewed the acceptance rule, Git workflow guidance, and decision record against the existing BlueprintLoop approval contract. Document and decision validation passed; all 22 repository Skills validated. No actionable findings.

PR #1452, opened 2026-09-19
fix(worktree): repair worktree lifecycle and add a bounded managed reaper

EricSanchezok:
## Summary

Allow a session to leave and remove its own clean worktree without rejecting its own use token or Git lock. Report deferred cleanup when removal cannot complete. Add a configurable managed-worktree cap with background sweeps and Git health guidance.

## Reclamation contract

- Explicit removal, automatic sweeps and stale reconciliation share the lifecycle gate. Anonymous and competing users remain blockers; idle bindings return to the main checkout before reclamation.
- Automatic sweeps preserve every on-disk lock. A Synergy marker alone does not prove its owner exited. Explicit owner-turn cleanup releases only a lock acquired by this process whose reason still matches; replacement user locks survive.
- Missing remote refs mean local-only history, and failed commit probes block reclamation. Dirty, running, locked and otherwise unverified worktrees remain with a reported reason.
- Missing managed registrations are reconciled individually; no repository-wide prune touches external worktrees.
- Branch deletion requires landed-content evidence and retains unlanded merge resolutions that git cherry cannot see.
- Scope disposal cancels timers and awaits the active sweep. Creation-triggered requests only affect Scopes with an enabled janitor.

## Verification

Regression tests reproduce active-use deletion, unpublished history without remote refs, external-registration pruning, stale session bindings, cross-process and replaced locks, unlanded merge-only changes, and premature disposal. Runtime Local, Workbench and Product Runtime workspace tests and local quality gates validate the fixes. CI must pass on the final submitted commit before merge.

A cap can remain exceeded when worktrees are protected. Stale on-disk locks require explicit operator recovery; the janitor does not infer cross-process liveness.


yzxoi (approved):
Reviewed the removal and janitor ownership boundaries. Fixed active-use bypasses, unverified local commits and lock ownership, global pruning, idle binding migration, merge-only branch loss, and disposal races. Integrated Runtime Local, Workbench and Product Runtime suites passed (40, 44 and 48 tests respectively), along with all 17 local quality gates. Fresh CI must pass before merge.

PR #1455, opened 2026-09-19
fix(product-runtime): assert the escalation, not the teardown tail

EricSanchezok:
## What and why

`dev` went red after this batch merged, on `Test Shards (runtime)` and `Coverage Shards (product-platform)`, both failing one assertion in the escalation test:

```
(fail) a terminally failed store escalates once into a non-zero runtime exit [240041.00ms]
Expected: 1
Received: -1
```

That test boots a second full Runtime in a child process and asserted the child's **exit status**. `gracefulShutdown` reaches `process.exit` only after `handle.close()`, `Log.flush()` and the rest of the teardown chain return, and on a loaded shard that tail is not a bound this test can honestly assert on: a run whose escalation had already closed admission was still draining when the parent gave up, and the assertion read that as a missing escalation.

The previous version made it worse rather than better — it wrote its marker **unconditionally** after a poll loop, so it passed whether or not escalation happened. That is why it looked green on its own pull request and then failed on the tip.

## Changes

- **The marker is now written only when the escalation is actually observed.** Closing admission is the first act of `gracefulShutdown`, so a `503` from the health route is evidence the storage-unavailable listener reached the runtime shutdown path. An unobserved escalation now fails the assertion instead of being reported by a marker that would have been written regardless.
- **The assertion is about that event, not the exit status that follows it.**
- **The parent owns termination**, because the teardown tail is not this test's to bound.
- The child still has no deadline of its own: `run()` never settles (`await new Promise(() => {})`), so it holds until the escalation or the parent stops it.

The exit code and the supervisor contract it serves remain covered by the deterministic sibling suite `packages/harness/test/storage/storage-unavailable-escalation.test.ts` (listener fires once, store fails terminally, host is notified exactly once) plus `registerShutdown` → `gracefulShutdown(signal, 1)` → `process.exit(1)`. That boundary is stated in the decision record rather than presented as behavioural verification here.

## Evidence

| Condition | Previous version | This version |
| --- | --- | --- |
| CI shape (one file, own process, coverage) | pass, then **failed on the merged tip** | pass |
| 8 concurrent coverage runs | — | **8/8 pass** |

Reproduced before the change at 8-way concurrency under coverage and on the `dev` tip: `Expected: 1, Received: -1`.

`bun run typecheck` 28/28, `bun run quality:quick` 17/17, `test-layout:check` / `decision:check` / `doc:check` pass.

## Note on scope

One test file. No production code changed: the escalation behaviour (`Storage.onUnavailable` → `registerShutdown` → `gracefulShutdown` → non-zero exit) was already correct, and this only changes what the test observes.


EricSanchezok:
Superseded by a branch rebuilt directly off `dev` — this one carried pre-merge history and showed CONFLICTING, so its CI never queued.

PR #1457, opened 2026-09-20
fix(product-runtime): assert the escalated exit status and correct stale capability docs

EricSanchezok:
The storage escalation regression verifies admission closure and a non-zero process exit observed both by the child exit handler and by the parent. It waits for the runtime's `server.start` event before killing the storage worker, since HTTP health alone can precede completed startup and shutdown registration. Marker reads handle the final process-exit race, and fixture cleanup waits for child termination.

The GitHub CLI permission matrix also uses the current `shell` risk floor for read-only commands and documents the additional `network_request` capability.

Validation:
- `bun test --cwd packages/product-runtime test/server/storage-unavailable-escalation.test.ts`: passed, including all four assertions. The initial reviewed version reproduced an exit timeout before startup synchronization was added.
- `bun run quality:quick`: all 17 gates passed.

BlueprintLoop status: the author's associated loop was reported as awaiting re-audit. This independent PR review does not change that loop status or claim a `blueprint_loop_approve` result.


yzxoi (approved):
Reviewed the current permission matrix and the process-level escalation assertions. Fixed the startup synchronization race, terminal marker read race and fixture cleanup. The isolated escalation test and all 17 local quality gates passed; merge remains contingent on fresh CI. This independent review does not assert completion of the author's associated BlueprintLoop.

PR #1462, opened 2026-09-20
fix(session): keep prompt accounting and calibration trustworthy

EricSanchezok:
## What and why

A long-running session grew past its model's input limit and the provider rejected the request outright, because the prompt budget never saw the growth and token accounting could not show that anything was missing. Three independent defects converged on that failure.

**1. A lost transport recording silently blanked token accounting.**
`RolloutAccounting.charge()` read provider usage only from a recorded attempt. Transport recording is deliberately best-effort, so a provider reaching the model through a non-recording transport had every call counted as unknown — even though the SDK reported usage that `RolloutLedger.finishCall` already persists as `call.sdkUsage`. The loss tracked the bypass exactly: that provider's transport-capture rate fell from about 85% to 0.2% while every other provider stayed at 99.8%, so the one signal that would have explained the missing tokens went missing with them. Because the affected call was the only provider-reported baseline available, the prompt budget was left anchored to an anchor that no longer described the request.

`charge()` now falls back to `call.sdkUsage` through a new `RolloutUsage.normalizeSdk()`, which maps the AI SDK's camelCase usage shape (`inputTokens`/`outputTokens`/`reasoningTokens`/`cachedInputTokens`). That shape shares no field names with the provider wire format the existing `normalize()` parses, so it needs its own mapping. Cached input is subtracted from the SDK's inclusive input total, and a count the SDK did not report stays unknown instead of becoming zero. A recorded attempt still takes precedence.

**2. The prompt budget's calibration path could go stale and skip measurement.**
`SessionInvoke.buildCalibration()` took a provider-reported baseline and added a cheap `chars/4` delta of everything appended since, letting `PromptBudgeter.decide()` skip a full tokenizer measurement. It accepted an anchor from any provider and model, so after a model switch the baseline described a different system prompt, tool set, and tokenizer; and its delta had no `reasoning` branch, so reasoning text was never counted. Reasoning accumulates every turn and does reach the provider, so the under-count grew with conversation length — a gateway-reported input count of 997,665 tokens reconciled only when reasoning was included (about 953.8k estimated) and not when it was excluded (about 344.6k).

Calibration now requires the anchor's `providerID` and `modelID` to match the model being invoked, counts `reasoning` parts, and declines itself — returning `undefined`, forcing a full measurement — once the estimated delta exceeds a quarter of the baseline (`CALIBRATION_MAX_DELTA_RATIO`). That guard is what makes a stale anchor non-fatal rather than silently wrong: calibration is an accelerator that may only be trusted while its approximation error stays small relative to the number it corrects. The relevant heuristic error was measured on real content — about -1% on reasoning text, -26% on tool output, -51% on plain text.

**3. A provider's context-overflow wording was not recognized.**
`SessionCompaction.isContextExceeded()` gates emergency compaction and the mechanical-summary fallback on a list of substrings. Alibaba Model Studio returns `InvalidParameter: Range of input length should be [1, <N>]` for an over-long prompt, which matched none of them, so an overflow the harness could have absorbed was treated as a fatal provider error and the turn died. The matcher now also accepts `range of input length` and the pair `input length` + `exceeds`.

## Verification

- `packages/harness/test/session/rollout-usage.test.ts` — SDK field mapping, inclusive cached input, missing counts stay unknown
- `packages/harness/test/session/rollout-accounting.test.ts` — call-level usage survives a lost recording; a recorded attempt still wins
- `packages/harness/test/session/invoke-calibration.test.ts` — reasoning counted, foreign anchor ignored, ratio guard declines, no anchor means no calibration
- `packages/harness/test/session/compaction.test.ts` — the input-length wording and the bare `input length` + `exceeds` pair

```bash
cd packages/harness
bun test test/session/invoke-calibration.test.ts test/session/rollout-usage.test.ts \
  test/session/rollout-accounting.test.ts test/session/compaction.test.ts
bun run quality:quick   # all 17 gates
```

93 focused tests pass. All 17 `quality:quick` gates pass, and the pre-push checks (format, lint, typecheck, monorepo, doc, decision) pass.

The full `packages/harness` session suite reports 6 failures, all in unrelated migration, recovery, and summary tests. They are **pre-existing**: at the base commit the same suite reports 917 passing and the same 6 failing; with this change it reports 928 passing (the +11 new tests) and the same 6. They also pass in isolation and when run as a group, so they look like cross-file interference rather than a regression here. I did not attempt to fix them.

## Alternatives considered

Recorded in the decision record, with the losing options: repair the transport bypass so every attempt is always written; raise the ratio threshold; drop calibration and always measure; gate reasoning counting on `capabilities.interleaved`; and fix only the error matcher. The `interleaved` gate in particular is worth flagging — it is the narrower condition and it is wrong for the actual wire shape: decoding a recorded request body showed `{"type":"reasoning"}` parts serialized to the provider with no interleaved-style provider option, so gating on that capability would leave the under-count in exactly the models that emit large reasoning traces.

## Documentation

`docs/architecture/llm-loop.md` states the calibration contract, and `docs/decisions/implemented/bug-fix/2026-09-20-prompt-budget-accounting-and-calibration-validity.md` records the decision.


yzxoi (dismissed):
Reviewed the diff against `dev` and verified the behavior directly.

**Verification (applied to a clean worktree at base `7a5f1f606`, tree matches head `3a38d3f` exactly; 19/19 behavior checks pass):**

- `normalizeSdk`: camelCase mapping correct, cached input subtracted from the inclusive `inputTokens` total, missing counts stay unknown (`complete=false`), payloads without any count → `null`.
- `charge()` fallback: a snapshot with `call.sdkUsage` and no recorded attempts now yields known tokens (input 1000 / cacheRead 200 / total 1500) while `attempts=0, unobservedCalls=1` keeps the evidence gap visible; a recorded attempt still wins over call-level usage.
- Calibration: reasoning counted in the delta (400 chars → 100 tokens), foreign provider/model anchor rejected, delta > 25% of baseline declines to a full measurement, no anchor → `undefined`. The 25% self-withdrawal guard is the right call given the measured `chars/4` error profile (-26% to -51%): the heuristic must stay a correction, never a replacement.
- `isContextExceeded`: both `range of input length` and the `input length` + `exceeds` pair match; unrelated errors still rejected.
- Confirmed the three fixes are one causal chain: the assistant-message token anchor that calibration reads is derived from the same `RolloutAccounting` summary this PR repairs, so fixing accounting is what re-anchors the budget.

**One known limitation worth recording (low severity, not blocking):** `normalizeSdk` hardcodes `protocol: "openai"` and OpenAI's inclusive-cached-input semantics for every provider. I checked the actual SDK mappings in `ai@5.0.212`: the assumption holds for `@ai-sdk/openai`, `@ai-sdk/openai-compatible`, and `@ai-sdk/google` (`prompt_tokens` / `promptTokenCount` are inclusive of cached tokens). It does **not** hold for `@ai-sdk/anthropic`, whose `inputTokens` maps from `usage.input_tokens` — exclusive of cache reads/writes in Anthropic semantics. An Anthropic-protocol model recovered through the no-recording fallback can therefore compute `uncached = inputTokens − cachedInputTokens` incorrectly (on heavy cache hits the difference clamps to null → `complete=false`, and the derived calibration anchor under-counts the uncached portion). Bounded — this path was all-unknown before this fix, and recorded attempts take precedence — but I'd suggest either a one-line caveat next to the hardcoded `"openai"` or a per-protocol branch as a fast follow.

Good to merge from my side. The cross-file session-suite interference the author flagged is pre-existing at base and separate from this change.

yzxoi (dismissed):
Reviewed the retry accounting, SDK usage normalization, calibration inputs, and compaction callers. Fixed aggregate call usage being charged to failed pre-response retries and preserved unknown Anthropic/Vertex/Bedrock cache-write usage instead of reporting false totals. Added behavioral regressions for both cases. Validation: 97 focused accounting/calibration/compaction tests and all 17 quality gates passed. Approval covers the current reviewed head; merge remains subject to required CI.

yzxoi (approved):
Re-reviewed the additional Google SDK correction at bdb896db4 against the locked SDK implementation: visible candidates and thinking usage are separate, so the normalized output uses total-minus-input or the complete component sum. Missing thinking remains unknown. Re-ran the accounting/usage/calibration/compaction selection: 104 tests passed. The prior retry/cache-accounting fixes remain intact.

PR #1467, opened 2026-09-20
fix(storage): make historical upgrades bounded and resumable

EricSanchezok:
Released JSON installations can be forced through every historical conversation before accepting new work. This change activates eligible v3.0.22 upgrades after required global work, then prepares old conversations on demand or in the background while retaining one SQL business-storage path.

- Classify migration execution and validate cross-domain dependencies against the shipped migration ledger. Keep unknown shared migrations and recovery-active owners as explicit startup barriers.
- Fence unpublished owner records at the storage boundary. Stream bounded import batches, preserve artifact pins, atomically publish records and indexes, and resume cleanup after interruption.
- Prioritize foreground work, bound file and compression concurrency, support pause/resume and cancellation, and back off on disk/WAL pressure or failures.
- Introduce format 4 snapshot segments with original-path protection and independently verifiable recovery copies. Preserve existing upgrade protocols; reject external or symbolic object dependencies rather than claiming a complete backup.
- Add asynchronous preparation status, retry and background controls to the API, generated SDK, Web and CLI. Make full VACUUM conversion opt-in maintenance and reuse the shared driver maintenance lifecycle from #1465 for finite Desktop budgets.

Validation:
- `bun run quality:quick` — all 17 gates passed; pre-push validation passed.
- Harness storage, migration and snapshot suites — 325 passed, 2 PostgreSQL-only skips locally; subsequent targeted fault tests passed.
- Complete Web suite and production build passed, including readiness-gated completion notices and rollback-loop prevention. Desktop suite: 305 passed; real Electron startup-progress tests passed separately.
- Snapshot suite also passed against Git 2.25.1 built from the official release source: 100 tests, 459 assertions.
- Product release-upgrade and startup-reporting tests, server route tests and CLI command tests passed.
- Isolated Web/server testing exercised 100 historical conversations, a 3,001-message conversation, pause/resume, on-demand opening, activated restart, and healthy/quarantined conversation isolation. Shutdown cancellation is tested across independent storage Handles. Managed Desktop completed a deliberately delayed 35-second migration in an isolated test environment. Direct inspection of the native window was blocked by the locked host; the real Electron progress DOM tests passed.
- Local synthetic measurements: 1,000 conversations became usable in about 1.2 seconds; 10× larger transcript bodies did not increase startup time. Three alternating full-import comparisons reduced median duration from 8.39 to 7.65 seconds. These are local measurements, not universal latency guarantees.

The decision record and Chinese research report describe recovery constraints, benchmark reproduction and remaining production stress-test limits. The branch incorporates current dev, including #1465; the combined startup paths and a 35-second isolated managed upgrade were revalidated. No production data or credentials were used. CI must validate the PostgreSQL 16–18 matrix and all platform gates before merge.


yzxoi (approved):
Reviewed historical-session admission, per-owner migration ordering, segmented backup sealing/import/publication, artifact pinning, and foreground/background scheduling. No remaining blocking code finding in the reviewed head. Validation: 225 storage/migration tests, the released 3.0.22 ledger upgrade fixture, 17 storage-route tests, the Web suite including its production CSS build contract, and all 17 quality gates passed. The branch includes the current development base; approval and auto-merge remain subject to required CI.

PR #1468, opened 2026-09-20
fix(storage): stop a busy worker from killing the runtime, and reclaim the store's structure

EricSanchezok:
# fix(storage): stop a busy worker from killing the runtime, and reclaim the store's structure

## Why

On 2026-09-20 at 18:15:56 the runtime shut itself down while its storage was healthy. Retention's owner-enumeration statement — a `GROUP BY` over 15,081,375 rollout rows that also selected `MIN(key_text)` — took 182 s against the 47.7 GiB store. The worker answers probes from the same single event loop it runs statements on, so a healthy-but-occupied worker is indistinguishable from a dead one. The driver treated the probe timeout as terminal and latched, which escalated to `gracefulShutdown` and `exit(1)`.

The index that statement needed has since been added, and the same query now runs in 1.36 s. But nothing about *the mechanism* changed: one long statement could still kill the process. That is what this branch fixes, along with the structural redundancy the same store was carrying.

## What changed

**Phase 1 — a probe timeout is a busy worker, not a dead one.** `SqliteDriver` gets an explicit `healthy | busy | exited | latched` state. A probe timeout now marks the worker *busy*: it surfaces the error that used to be discarded, raises `STORAGE_WORKER_BUSY`, records `storage.worker.probe.timeout`, and fails new work fast with a retryable `StorageBusyError` while the runtime keeps serving. Only an unexpected `worker.exited`, or silence past a hard ceiling, still latches. The latch and the supervisor restart model are unchanged — only *when* they fire. Teardown is bounded by its own small budget rather than the ceiling, because a host on a shutdown deadline was otherwise waiting longer than it had.

**Phase 2 — no statement may outlive a fixed budget.** The byte-scaled maintenance deadline (`600_000 + ceil(bytes/1024²)×1000`, which resolved to 12.16 h at 42.9 GiB) is gone. Every maintenance, DDL, delete and migration path now runs on one fixed chunk budget, and `pruneTree` is chunked. `chunkBudgetMs × 4 ≤ hardCeilingMs` is asserted, so raising a chunk budget cannot silently reopen the terminal window.

**Phase 3 — format 3, a full layout rewrite.** Key columns become `BLOB`, whose lack of affinity lets a byte string and the legacy hex text share one column so the rewrite runs in bounded batches beside the live table. `storage_nodes` drops its cumulative path text; the message index becomes partial on a non-empty key; the artifact pack name becomes a virtual column derived from the JSON beside it. The rewrite is a registered migration that copies in keyset batches, swaps all three tables in one transaction, and then reclaims the pages it freed. It also measures whether the volume can hold it before starting.

**Phase 4 — the retention window follows the budget.** `retentionMs` becomes a promise ceiling and the operative window is the shorter of it and `retentionBytes ÷ measured ingress`, with a one-day floor. Ingress comes from signals a pass already has and is persisted, so a restart resumes the estimate. The permanent `overBudget && pruned=0 && infeasible` spin in the incident's logs is gone: a window the budget cannot hold is now a named degradation, and only a budget that cannot hold the floor stops pruning.

## Reconciliation with `dev`

`dev` advanced with #1465 while this branch was open, and both changed the same driver contract. The two designs encoded the **same** distinction, so the merge keeps one of them:

- `dev`'s `StorageMaintenanceOperation` (`vacuum`, `reclaim`, `integrity-check`, `create-index`, `drop-index`) is the typed name of the work, and it already separates the one splittable operation from the rest — only `reclaim` frees a bounded page count per call. That *is* the classification the chunk budget needs, so it supersedes the separate `unchunkable` flag this branch had added beside it, and `onMaintenanceBudget` disappears because the lifecycle already reports each operation's budget. One field cannot now disagree with the operation it describes.
- `dev`'s byte-scaled maintenance deadline is deliberately **not** kept. It is what let one statement occupy the worker while the probe budget deciding this driver's fate stayed a fraction of it, and removing it is the point of this branch. Everything `dev` added around it — typed lifecycle events, stages, overlap selection in Desktop, the bounded queue — is kept in full.
- `dev` registered its own postmortem 0020 in the same range, so this branch's busy-worker postmortem moves to **0021**.

### Independent measurement, not the implementer's own

Measured by me on a 3.17 GB / 920,000-record fixture built from the production schema:

|                             | before (format 2)         | after (format 3 + reclaim)   |
| --------------------------- | ------------------------- | ---------------------------- |
| file bytes                  | 3,175,882,752             | **1,853,894,656 (−41.63 %)** |
| page bytes                  | 3,175,849,984             | **1,785,966,592 (−43.76 %)** |
| freelist                    | 0                         | **0**                        |
| logical digest              | `4fcb9ee3…`               | **identical**                |
| records / nodes / artifacts | 920000 / 1464004 / 160000 | **identical**                |

A follow-up `reclaim` call released **0 pages**, because the migration itself had already returned all 675,400. `scan`, `list`, `query` and portable export are unchanged.

### The incident's mechanism, reproduced against a real worker process

While the store was `busy`, a request on an independent admission path failed in **0 ms** with a retryable `StorageBusyError`; the worker returned to `healthy` on its own; `onUnavailable` **never fired**. With the statement pushed past the hard ceiling, the store still latched **exactly once** and escalated as before. A slow-but-successful maintenance statement inside the ceiling **committed** rather than rolling back.

### A real runtime upgrading a populated format 2 store, on the final head

An isolated home — its own `SYNERGY_HOME` on a private port, seeded with a genuine format-2 store — was started with `bun ./script/dev.ts server`:

- `health=200`, and the runtime ran the rewrite inline: *"Rewrite records, nodes and artifact locators into the format 3 layout ✓"*
- afterwards: version **3**, `key_id` and `body` type `blob`, `storage_nodes.key_text` gone, `storage_artifacts.pack` **`hidden=2`** (engine-derived, so the supported path was taken — no fallback was reported), `freelist_count` **0**, `storage_format_v3_state.state.phase = "complete"`, and `storage_records_message` carrying `WHERE message_id <> ''`
- **39,469,056 → 21,700,608 bytes, a 45.02% reduction**

Logical equivalence is measured separately from that startup run, because a live runtime also writes its own bookkeeping rows. The same fixture migrated through `StorageFormatV3Migration.run` alone reports **`digestMatch: true`** — digest `3229ee9b33480f51155492035620ee7ba74c17fa6d341c9f39e0d2601fd0b0f3` identical before and after, with body forms moving `plain 9041 → frame 9041` and no `missingInAfter` entries.

The regression that broke CI was reproduced and fixed on the same contract it broke: `bun test --cwd packages/harness test/session/rollout-long.test.ts --test-name-pattern ": completed$"` with `SYNERGY_ROLLOUT_LONG_STREAM=1` now reports **`fixture cleanup completed`** and passes in 199,977 ms, where it previously hung ~21 minutes at that line and timed out.

### Four defects this branch introduced, found and fixed before merge

1. **The body codec emitted a byte frame unconditionally.** A `TEXT` body column cannot hold one — the backend renders those bytes as hex text, and every reader then rejects the value as invalid encoding. Format 2 keeps that text layout (PostgreSQL always; SQLite until its rewrite), so writes to a format 2 namespace stored something nothing could read. The container now follows the namespace's recorded format, exactly as the key encoding already does. Against real PostgreSQL 16 the storage suite went from **208 pass / 8 fail to 216 pass / 0 fail**.
2. **The benchmark's rollout probe could not read a rewritten store.** It is the only reader of the record body column outside this repository, and it decoded the two text forms only. Format 3 stores a compressed body as a byte frame, and Python's `sqlite3` hands a `BLOB` to the caller as `bytes`, so `body.startswith("z:")` raised `TypeError` and the probe reported a recorded provider response as absent. This failed `Benchmark Docker Lifecycle` on a green `dev` tip, which is how it was caught. The probe now decodes all three forms, sizing its decompression buffer from the declared length in the header and rejecting any other result.

3. **Node cleanup was quadratic.** `remove`/`removeTree` left the node rows a removal had emptied, so the cleanup added for it walked the subtree with `FROM storage_nodes child JOIN subtree`, which makes `storage_nodes` the driving table and re-scans it once per recursion step. Measured on the shape a rollout artifact writes — one shared prefix plus one leaf per chunk — that form took **32,270 ms** for 32,000 nodes against **13 ms** once the recursive table drives. Through `removeTree` the teardown went from **101 s to 0.36 s**. This is what turned `Long Rollout Contracts` red: its fixture writes ~30,720 chunk nodes under one session prefix and removes the session, so teardown spent ~21 minutes in the cleanup and timed out. The job was green on the four runs before the cleanup landed and failed on the commit that added it, so the regression was this branch's. Verified by running the same contract the job runs (`bun test --cwd packages/harness test/session/rollout-long.test.ts --test-name-pattern ": completed$"` with `SYNERGY_ROLLOUT_LONG_STREAM=1`): 199,977 ms, exit 0, and teardown now reports *"fixture cleanup completed"* instead of hanging.
4. **Editing public config help text without regenerating the build schemas.** The corrected `chunkBudgetMs` description changed `packages/{cli,product-runtime}/schema/config.schema.json`, and `Test Aux Contracts` compares the generated schema against them. Both profiles were regenerated through the build's own `generateSchema`; the only delta is the description, and `release:test` is 172 pass / 0 fail.

## Bounded close regression

A real-worker review found that closing a busy driver left its liveness monitor repeatedly resolving failed probes until the recovery ceiling, starving the teardown timer and incorrectly reporting deliberate shutdown as terminal unavailability. The monitor now exits when the driver closes or becomes unavailable; requests also recheck driver and pending-request ownership after awaiting it.

The regression failed at approximately 10 seconds with a 2.5-second teardown budget before the fix. It now closes within that budget's test tolerance without an `onUnavailable` notification. The fix includes a permanent regression, architecture documentation, a decision record and the persistence development guide.

Validation for the review fix and integration with current `dev`:

- `bun test --timeout 30000 test/storage` from `packages/harness`: **243 passed, 3 PostgreSQL-only tests skipped, 0 failed**.
- Released 3.0.22 staged-upgrade integration: **1 passed** on the combined historical-upgrade / format-3 branch.
- `bun run quality:quick`: **all 17 gates passed**, also rerun after integrating `dev`.
- The bounded-close regression passed again after integrating `dev`; commit and push hooks, including all 29 package typechecks, passed.

Latest CI must pass on the current PR head before merge. The acceptance results below describe their stated historical revision.

## Gates

All verified on head `f86c626cd`:

- CI: **37 checks pass, 0 fail** (including Test Shards, Coverage Shards, PostgreSQL 16/17/18, Desktop, Windows, Typecheck and Quality)
- storage suite on SQLite: **222 pass / 3 skip / 0 fail**
- `tsgo --noEmit`: clean
- `bun run quality:quick`: **all 17 gates passed**
- benchmark Python contracts: **ruff, format and mypy clean; 241 passed / 21 skipped**
- `decision:check`, `doc:check`, `doc:gen:check`, `test-layout:check`: passed / fresh

## Decisions and their records

- Postmortem: `docs/postmortem/0021-busy-sqlite-worker-declared-dead.md`
- Decision records: probe timeout is a busy worker (not a dead one); maintenance statements bounded to fixed chunk budgets; the retention window derived from budget and ingress; storage format 3.

## Known limitations

- **The hard ceiling rests on a noisy measurement.** Repeated runs against the *same* file returned integrity-check times of 140 s, 173.6 s and 280 s — a two-fold spread governed by page cache. It defaults to one hour, above the pessimistic end, rather than from a point estimate. Three statements cannot be chunked or cancelled at all: `CREATE INDEX`, `PRAGMA integrity_check` and `VACUUM`.
- **The rewrite needs transient disk.** It builds a replacement table before dropping the original, measured at 1.44× the file. On the 47.7 GiB production store that is roughly 21 GB beyond the current file. It now performs a `statfs` preflight and reports the shortfall rather than failing partway with a bare engine error.
- **The production store itself was not migrated here.** Equivalence and reclaim are proven on a 3.17 GB production-schema fixture plus a live-runtime upgrade of a smaller populated store.
- **PostgreSQL coverage is CI-driven.** Real 16, 17 and 18 all pass on this head. A local PostgreSQL 16 run was also used to attribute one timeout (see defect 3's sibling: the packed-import test that timed out on PostgreSQL 17 while 16 and 18 passed the same test on the same commit, so it was load-sensitive rather than deterministic).
- `packages/harness` cannot reach its coverage floor from its own suite alone by design; CI evaluates it in the aggregate job across shards.
- PostgreSQL keeps the format-2 layout and the rewrite is a no-op there.

## Compatibility

`version: 3` is written only by the swap transaction, never by an open of an existing namespace, so a format-2 store cannot claim format 3 while its rows are still hex. Reads accept all three body forms permanently. An older release rejects a version 3 namespace, and downgrade uses the pre-upgrade backup in a separate Home, as the authority decision already requires.

## Blueprint loop status

This change was executed under a BlueprintLoop for the "storage subsystem hardening and v3 full-format rewrite" Blueprint. The loop was submitted for independent audit twice.

The **first audit rejected** it: the headline outcomes were independently reproduced, but four required outcomes were not delivered. Those four were then implemented, tested and verified on head `f86c626cd`:

1. **Bounded node cleanup on `remove`/`removeTree`** (Phase 3 item 6). The predicate `pruneTree` uses counts *any* record row, which is correct there because it deletes rows outright, but never fires after a tombstone-only removal; eligibility here is "no live record at this node's own key and no surviving child", iterated to a fixed point so an interior node drops once its subtree has drained. The candidate set is the removed prefix's subtree plus that prefix's own ancestors, both reached from the prefix, because a digest cannot reconstruct a parent link. `remove`, `removeTree` and the empty-prefix branch are covered, including a format-2 hex namespace, an 80-deep chain and a 16,384-leaf guard on the shape a rollout teardown removes.
2. **Generated-column verification with a physical-column fallback** (Phase 3 item 5). The rewrite stages the replacement table and the index the swap will build on a scratch table, then reads `pragma_table_xinfo` (`hidden >= 2`) and `pragma_index_xinfo` for a key entry on `pack`; a rejected declaration and a rejected index are the same verdict. On failure it stages `pack TEXT NOT NULL` and backfills from `location`, keeping the column name, the `storage_artifacts_pack` index and every reader. The layout is never recorded separately — the staging table is itself the authority, so a resumed run converges instead of re-deciding.
3. **Documentation no longer contradicts the shipped code.** The probe-timeout record says 3600 s (not 1800 s); the maintenance record and `agent-storage.md` state the real model — ordinary work on `requestDeadlineMs`, `reclaim` on `chunkBudgetMs`, and `CREATE INDEX`/`PRAGMA integrity_check`/`VACUUM` on `engineBudgetMs`; the compression floor is 64 B with an honest byte gate, not "512 bytes / 10%". The public `chunkBudgetMs` help text was wrong in the same way and is corrected.
4. **The last two `encode()` bypasses are gone.** `storage_events.payload` and `storage_receipts.result` now use `RecordCodec.encode`/`decode` with the same container rule as the record body, the local helper is deleted, and `restoreEntry` uses `reencode` so a repeated import compares equal.

Two regressions I introduced while remediating were found and fixed rather than shipped: the quadratic node cleanup (defect 3 above) and the stale build schemas (defect 4).

**Re-audit passed and the loop is complete.** The reviewer verified 56/56 required outcomes complete on head `f86c626cd`, reproducing the evidence independently rather than reading the implementer's narration: busy-vs-dead behaviour against a real worker subprocess (bounded occupation → busy with no terminal event and a surviving runtime; past-ceiling → exited with exactly one `StorageUnavailableError`), logical equivalence through the store's own read-path oracle with a byte-identical digest, −43.8% page reduction on the 920k fixture and −45.02% on the isolated runtime upgrade, and the format-3 end state (version 3, `phase=complete`, blob key and body, four-column `storage_nodes`, partial message index, `pack hidden=2`, `freelist_count=0`). The audit recorded the scope as clean — a 42-file delta against `dev`, no OpenAPI route changes and no forbidden additions — and the change is accepted at this head.


yzxoi (dismissed):
Reviewed SQLite busy/dead transitions, the resumable format-3 copy/swap/reclaim, retention, codec compatibility, and tests. Fixed a capacity-preflight omission: records/nodes phases did not count the artifact table still to be copied. Both initial and resumed-node regression tests failed before the fix and now pass; metadata remains at format 2 when preflight refuses the copy. Validation: 222 other storage tests plus all 5 capacity tests passed (3 PostgreSQL-specific cases skipped locally), and all 17 quality gates passed. The PR body records the completed BlueprintLoop and successful 56/56 re-audit; this review additionally covers the capacity correction and integration with dev. Merge remains subject to required CI.

PR #1471, opened 2026-09-21
fix(worktree): preserve session recency during automatic cleanup

yzxoi:
Automatic worktree cleanup updated session activity while moving idle bindings back to the main checkout, causing old sessions to jump ahead of recent conversations. Both cap reclamation and missing-worktree reconciliation now preserve canonical and navigation activity while still persisting and publishing the workspace change. Rebuilding the derived navigation index retains the same ordering. Explicit workspace operations and new conversation activity retain their existing behavior.

Validation: 37 worktree lifecycle tests and 26 session navigation tests passed, including regressions that failed before the fix. All 17 `quality:quick` gates passed. The change includes the workspace contract, persistence guidance, and a decision record. Previously advanced activity timestamps are not guessed or rewritten.


yzxoi (commented):
Reviewed the current diff and automatic cleanup call sites. The preserved timestamp now remains in canonical session.time.updated, so navigation-index reconstruction keeps the original ordering. Both cleanup modes and explicit later activity are covered. Verified on the current dev base together with #1472 and #1473: runtime-local workspace suites passed 42 tests. The existing recency review thread is addressed.

PR #1472, opened 2026-09-21
fix(rollout): settle uploads before ending recorded attempts

yzxoi:
Fetch can fail while its native reader still holds the upload stream lock. Ending the recorded attempt first allowed a pending upload chunk to arrive after the recorder removed that attempt, replacing the original transport failure with `Unable to persist rollout evidence`.

Attempt completion now settles request reads and admitted evidence writes through the stream owner's cancellation path, independently of the consumer lock. It does not wait for cancellation acknowledgement shared with an unowned live sibling of a cloned Request. The same barrier handles early HTTP responses and avoids request-side finalizer recursion. Upload cleanup errors preserve the transport outcome; actual persistence failures remain authoritative.

Validation: all 34 targeted transport/call tests and all 17 `quality:quick` gates passed. Regressions cover a cloned Request interrupted during upload, early empty/streaming responses, live sibling ownership, and rejecting cancellation hooks. All three recorder pressure outcomes (completed, cancelled, failed) passed in separate processes, each validating 30 MiB, 30,720 checkpoints, retained bytes, accounting, archive integrity, and fixture cleanup. Includes the evidence contract, development guidance, and decision record.

Supplemental Docker validation, frozen at the initial patch revision, finished with 3 passing and 3 failing tests. Disconnect, cancellation and container-stop cases passed. The paired matrix passed six core/Library tasks but both full-runtime tasks exceeded their deadline; the long-response case also timed out with a valid partial archive. The short-timeout case ended before any model call, so its complete zero-call recording did not satisfy the fixture's expected partial recording. These failures remain recorded; the matrix was not rerun against the cancellation-ownership follow-up. Owned test containers were cleaned up.


yzxoi (commented):
Reviewed request/response finalization, upload locking, tee cancellation and error precedence. The current head initiates request-side upstream cancellation without waiting for its live sibling, drains admitted evidence writes, and preserves the original transport error when cancellation rejects. All 34 transport/call regressions passed on the current dev base together with #1471 and #1473. Both existing upload review threads are addressed; current CI is green.

PR #1473, opened 2026-09-21
fix(web): reconcile effective history after rewind and redo

yzxoi:
Rewinding a retained conversation could leave messages from earlier rollback branches visible until a browser refresh. The latest rollback summary only describes one branch, and a history transition previously refreshed metadata without necessarily replacing the message window.

The active session now observes rollback identity and redo validity, reloading the authoritative effective message window on rewind, redo, and new roots that invalidate redo. Existing reconciliation removes obsolete message and part buckets and restores newly effective injections; immediate filtering also hides known dropped messages when the cut root is outside the loaded window. Unchanged history/connection keys do not trigger another sync.

Validation: 34 focused sync/filter/store tests, all 17 `quality:quick` gates, the production Web build, and the shared UI suite passed. The full Web run passed every batch except an initialization timeout in the existing decision-surface browser fixture; its isolated rerun passed all 5 tests. Includes a real SyncProvider regression across branch replacement and message/part restoration, the sync contract, development guidance, and a decision record.


yzxoi (commented):
Reviewed the rollback filter, watch key, reload plan and SyncProvider reconciliation. The current head observes both rollback identity and canUnrollback, and forces authoritative message replacement on history transitions. All 34 focused sync/filter/DOM regressions passed on the current dev base together with #1471 and #1472. The existing redo-validity review thread is addressed; current CI is green.

